PR #36, opened 2015-01-01
857 sms model overhaul

linesarefuzzy:


app/mailers/broadcast_mailer.rb
```diff
@@ -4,6 +4,6 @@ class BroadcastMailer < ActionMailer::Base
   def broadcast(recips, subj, msg)
     s = configatron.broadcast_tag + " " + (subj || "")
```
hooverlunch: This line is bit legacy/nasty. Could be moved into 7 and cleaned up.

app/models/broadcast.rb
```diff
@@ -48,50 +46,56 @@ def email_possible?
...
 
-  def no_possible_recipients?
-    recipients.all?{ |u| u.email.blank? && u.phone.blank? && u.phone2.blank? }
+  def recipient_emails
+    @recipient_emails ||= recipients.map { |r| r.email if r.can_get_email? }.compact
   end
 
   private
 
     def has_eligible_recipients
```
hooverlunch: Should end in ?

db/migrate/20141211205915_migrate_sms_message_data.rb
```diff
@@ -0,0 +1,45 @@
...
+    SQL
+
+    remove_column :sms_messages, :direction
+    change_column :sms_messages, :to, :string
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+end
```
hooverlunch: Wow, nice work here. Had no idea how tricky that would be.

spec/models/sms/adapters/frontline_sms_adapter_spec.rb
```diff
@@ -35,11 +35,11 @@
   it 'should correctly parse a frontline-style request' do
     Time.zone = ActiveSupport::TimeZone['Saskatchewan']
 
-    request = {'frontline' => '1', 'text' => 'foo', 'from' => '+2348036801489'}
+    request = {'frontline' => '1', 'text' => 'foo', 'from' => '+2348036801489', 'to' => '+123456789'}
```
hooverlunch: I don't think the to will be included in the request. It should be taken from the mission SMS to number if given. If not, it should be allowed to be null. Both scenarios should probably be tested.

spec/models/sms/adapters/intelli_sms_adapter_spec.rb
```diff
@@ -34,12 +34,12 @@
     Time.zone = ActiveSupport::TimeZone['Saskatchewan']
 
     request = {'text' => 'foo', 'sent' => '2013-07-03T09:53:00+01:00', 'from' => '2348036801489',
-      'msgid' => '1234'}
+      'msgid' => '1234', 'to' => '123456789'}
```
hooverlunch: See above.

hooverlunch:
Great work. Just a few changes. I can handle these myself when merging, so no more work needed on your end. Thanks!

linesarefuzzy:
Cool, thanks!

PR #39, opened 2015-01-15
when admin first logs in, redirect them to user edit page

mapmeld:
What's your thought on redirecting on the first login, to make sure that an admin sets up their password and user account? This could be applied to all users. There could be more descriptive text on the page.

### DIFF
--- a/app/controllers/concerns/application_controller/login_logistics.rb
+++ b/app/controllers/concerns/application_controller/login_logistics.rb
@@ -25,9 +25,15 @@ module Concerns::ApplicationController::LoginLogistics
     pref_lang = @user_session.user.pref_lang.to_sym
     I18n.locale = configatron.full_locales.include?(pref_lang) ? pref_lang : I18n.default_locale
 
-    # Redirect.
-    best_mission = @user_session.user.best_mission
-    redirect_back_or_default best_mission ? mission_root_path(mission_name: best_mission.compact_name) : basic_root_path
+    # Redirect admin's first login to password reset.
+    if @user_session.user.admin && @user_session.user.login_count <= 1
+      flash[:success] = t("user.set_admin_password")
+      redirect_to url_for(@user_session.user) + '/edit'
+    else
+      # Redirect to most relevant mission
+      best_mission = @user_session.user.best_mission
+      redirect_back_or_default best_mission ? mission_root_path(mission_name: best_mission.compact_name) : basic_root_path
+    end
   end
 
   # resets the Rails session but preserves the :return_to key
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1570,6 +1570,7 @@ en:
     reset_password_label_choose: "Password Creation"
     reset_password_label_reset: "Reset Password?"
     send_email_instructions: "send email instructions"
+    set_admin_password: "Set your admin password and profile information"
     show_printable_instructions: "show printable instructions"
 
   welcome:
--- a/config/locales/fr.yml
+++ b/config/locales/fr.yml
@@ -1348,6 +1348,7 @@ fr:
     reset_password_label_choose: "Créer mot de passe"
     reset_password_label_reset: "Réinstalisation de mot de passe?"
     send_email_instructions: "Envoyer les instructions par email"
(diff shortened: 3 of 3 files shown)


app/controllers/concerns/application_controller/login_logistics.rb
```diff
@@ -25,9 +25,14 @@ def post_login_housekeeping
     pref_lang = @user_session.user.pref_lang.to_sym
     I18n.locale = configatron.full_locales.include?(pref_lang) ? pref_lang : I18n.default_locale
 
-    # Redirect.
-    best_mission = @user_session.user.best_mission
-    redirect_back_or_default best_mission ? mission_root_path(mission_name: best_mission.compact_name) : basic_root_path
+    # Redirect admin's first login to password reset.
+    if @user_session.user.admin && @user_session.user.login_count <= 1
+      redirect_to url_for(@user_session.user) + '/edit'
```
hooverlunch: This is a nice idea. Could you include a flash message suggesting that the user change their password? Otherwise I worry they might be confused. (Be sure to put it in en.yml).

app/controllers/concerns/application_controller/login_logistics.rb
```diff
@@ -25,9 +25,14 @@ def post_login_housekeeping
     pref_lang = @user_session.user.pref_lang.to_sym
     I18n.locale = configatron.full_locales.include?(pref_lang) ? pref_lang : I18n.default_locale
 
-    # Redirect.
-    best_mission = @user_session.user.best_mission
-    redirect_back_or_default best_mission ? mission_root_path(mission_name: best_mission.compact_name) : basic_root_path
+    # Redirect admin's first login to password reset.
+    if @user_session.user.admin && @user_session.user.login_count <= 1
```
hooverlunch: Have you confirmed that login_count has already been incremented by this point? Otherwise this will run for the first two logins.

README.md
```diff
@@ -81,9 +84,9 @@ Note to install the software below we recommend the following package managers:
 1. **Bundle, configure, and migrate**
   - Install the required gems by running `bundle install` in the project directory.
   - Copy `config/database.yml.example` to `config/database.yml` and edit `database.yml` to point to your database.
-  - Copy `config/thinking_sphinx.yml.example` to `thinking_sphinx.yml.example` and adjust any settings (usually not necessary).
+  - Copy `config/thinking_sphinx.yml.example` to `thinking_sphinx.yml` and adjust any settings (usually not necessary).
   - Copy `config/initializers/local_config.rb.example` to `config/initializers/local_config.rb` and adjust any settings.
-  - Run database migrations: `rake db:migrate`.
+  - Run database migrations: `rake db:migrate`. If the diagramming step hangs, run `NO_DIAGRAM=true rake db:migrate`.
```
hooverlunch: Thanks for these edits.

README.md
```diff
@@ -45,6 +45,9 @@ Note to install the software below we recommend the following package managers:
   - Create an empty database and accompanying user for use by the app (E.g. development database *elmo_d* with username *elmo*)
   - Set up mysql for timezone support: See [doc here](http://dev.mysql.com/doc/refman/5.5/en/time-zone-support.html)
 
+1. **GraphViz**
+  - Install GraphViz from http://graphviz.org/Download..php
```
hooverlunch: Can you make this one look like the others (minimum version, short description of what it's for, put it near the bottom and put 'development only' since it's not needed for production, remove the download link since folks may want to use their package manager (it's fine to put a link explaining what the software is though).

hooverlunch:
Thanks for this. Can you change it so that the base is the develop branch? Can you also separate out the instruction changes into a different PR against the master branch? Thanks.

app/controllers/concerns/application_controller/login_logistics.rb
mapmeld: Had the same concern - I first tried it with <= 0 and it didn't appear at all. When I cleared the database and re-created the admin user, <= 1 made it appear only on the first login

mapmeld:
recreating this as a PR against the devel branch

PR #42, opened 2015-01-30
855 sms log search

linesarefuzzy:


app/models/sms/incoming.rb
```diff
@@ -8,6 +8,14 @@ class Sms::Incoming < Sms::Message
     return true
   end
 
+  def self.search_qualifiers
+    super + [
+      Search::Qualifier.new(name: "user", col: "users.login", type: :text, assoc: :users, default: true),
+      Search::Qualifier.new(name: "name", col: "users.name", type: :text, assoc: :users, default: true),
+      Search::Qualifier.new(name: "number", col: "sms_messages.from", type: :text, default: true),
+    ]
+  end
```
hooverlunch: I see what you're doing here, but since search_qualifiers is always called on Sms::Message (since that's where do_search is called), I think these qualifiers will never get loaded. I will move them into Sms::Message.

config/locales/en.yml
```diff
@@ -762,6 +762,8 @@ en:
     title: "title"
     type: "type"
     tag: "tag"
+    date: "date"
+    datetime: "datetime"
```
hooverlunch: user, content, and number were missing here.

hooverlunch:
Great work. There was one major bug that I fixed. See notes. I also ended up having to do some surgery on the associations b/c the join was messing things up. See my changes. Boy that was a doozie!

linesarefuzzy:
Yikes! Sorry about that.

hooverlunch:
No worries. All is well now. Going to do release today.

PR #43, opened 2015-02-24
Add components needed for question list and style questions list

emitche:
_screen.html.erb 
line 119, 122 - Is there a more effective/elegant way to get the right links?


emitche:
Submitting new request.

PR #45, opened 2015-02-26
Added crud for qing_group.

mkuklis:
http://redmine.sassafras.coop/issues/2076

### DIFF
--- a/app/controllers/qing_groups_controller.rb
+++ b/app/controllers/qing_groups_controller.rb
@@ -2,20 +2,7 @@ class QingGroupsController < ApplicationController
   # authorization via cancan
   load_and_authorize_resource
 
-  before_filter :prepare_qing_group, :only => [:create]
-
-  def new
-    @qing_group = QingGroup.new(:form_id => params[:id])
-    render(:form)
-  end
-
-  def edit
-    render(:form)
-  end
-
-  def show
-    render(:form)
-  end
+  before_filter :prepare_qing_group, only: [:create]
 
   def create
     create_or_update
@@ -27,15 +14,17 @@ class QingGroupsController < ApplicationController
   end
 
   def destroy
-    destroy_and_handle_errors(@qing_group)
-    redirect_to(edit_form_url(@qing_group.form))
+    begin
+      @qing_group.destroy
+      render nothing: true, status: 204
+    end
   end
 
   private
     # creates/updates the qing_group
     def create_or_update
       if @qing_group.save
-        render(:partial => 'item')
+        render(partial: 'form')
       else
         render(:form)
       end
--- a/app/controllers/qing_groups_controller.rb
+++ b/app/controllers/qing_groups_controller.rb
@@ -1,0 +1,51 @@
+class QingGroupsController < ApplicationController
+  # authorization via cancan
+  load_and_authorize_resource
+
+  before_filter :prepare_qing_group, :only => [:create]
+
+  def new
+    @qing_group = QingGroup.new(:form_id => params[:id])
+    render(:form)
+  end
+
+  def edit
+    render(:form)
+  end
+
+  def show
+    render(:form)
+  end
+
+  def create
+    create_or_update
+  end
+
+  def update
+    @qing_group.assign_attributes(params[:qing_group])
+    create_or_update
+  end
+
+  def destroy
+    destroy_and_handle_errors(@qing_group)
+    redirect_to(edit_form_url(@qing_group.form))
+  end
+
+  private
+    # creates/updates the qing_group
+    def create_or_update
+      if @qing_group.save
+        render(:partial => 'item')
+      else
+        render(:form)
+      end
+    end
+
+    # prepares qing_group
+    def prepare_qing_group
+      attrs = params[:qing_group]
+      attrs[:ancestry] = Form.find(attrs[:form_id]).root_id
+      @qing_group = QingGroup.accessible_by(current_ability).new(attrs)
+      @qing_group.mission = current_mission
+    end
+end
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -77,6 +77,7 @@
         put 'clone'
         put 'publish'
         get 'choose_questions', :path => 'choose-questions'
+        resources :qing_groups, :only => :new
```
hooverlunch: Can we add `:path => 'qing-groups'` for consistency here? Not that the user will ever see this URL in the addy bar...

spec/controllers/qing_groups_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
+require 'spec_helper'
+
+describe QingGroupsController, type: :request do
+  before do
+    @user = create(:user, role_name: 'coordinator')
+    @form = create(:form, question_types: %w(text text))
```
hooverlunch: If you do like `question_types: ['text', ['text', 'text']]`, you will get a qing group for free. You can get it via `@form.children[1]` or `@form.c[1]`.

spec/controllers/qing_groups_controller_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      post(qing_groups_path(mission_name: get_mission.compact_name),
+        'qing_group' => { 'form_id' => @form.id })
+    end
+
+    it 'should be successful' do
+      expect(response).to be_success
+    end
+
+    it 'should render item partial' do
+      expect(response).to render_template(:partial => 'qing_groups/_item')
```
hooverlunch: New hash syntax please.

config/routes.rb
hooverlunch: Don't need this. See above.

config/routes.rb
```diff
@@ -86,6 +87,7 @@
         get 'condition_form', :path => 'condition-form'
       end
     end
+    resources :qing_groups, except: :new
```
hooverlunch: Can we add `:path => 'qing-groups'` for consistency here? Not that the user will ever see this URL in the addy bar...

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+        render(:form)
+      end
+    end
+
+    # prepares qing_group
+    def prepare_qing_group
+      attrs = params[:qing_group]
+      attrs[:ancestry] = Form.find(attrs[:form_id]).root_id
+      @qing_group = QingGroup.accessible_by(current_ability).new(attrs)
+      @qing_group.mission = current_mission
```
hooverlunch: Beautiful.

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
+class QingGroupsController < ApplicationController
+  # authorization via cancan
+  load_and_authorize_resource
+
+  before_filter :prepare_qing_group, :only => [:create]
+
+  def new
+    @qing_group = QingGroup.new(:form_id => params[:id])
+    render(:form)
+  end
```
hooverlunch: We don't need the new or edit actions. Only create, update, destroy.

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def update
+    @qing_group.assign_attributes(params[:qing_group])
+    create_or_update
+  end
+
+  def destroy
+    destroy_and_handle_errors(@qing_group)
+    redirect_to(edit_form_url(@qing_group.form))
```
hooverlunch: Since this will be AJAX only, I think we don't need to redirect or use destroy_or_handle_errors. How about this:

```
begin
  @qing_group.destroy
  render nothing: true, status: 204
end
```

If there is an error, Rails will render some junk with status 500, and the AJAX call can catch the 500 and say server error.

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  private
+    # creates/updates the qing_group
+    def create_or_update
+      if @qing_group.save
+        render(:partial => 'item')
+      else
+        render(:form)
```
hooverlunch: The form will also be a partial.

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  def destroy
+    destroy_and_handle_errors(@qing_group)
+    redirect_to(edit_form_url(@qing_group.form))
+  end
+
+  private
+    # creates/updates the qing_group
+    def create_or_update
+      if @qing_group.save
+        render(:partial => 'item')
```
hooverlunch: New hash syntax please.

app/controllers/qing_groups_controller.rb
hooverlunch: I don't think we'll need a new action since this form can be rendered when the page first loads and thus doesn't need an AJAX call.

app/controllers/qing_groups_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+    render(:form)
+  end
+
+  def edit
+    render(:form)
+  end
+
+  def show
+    render(:form)
+  end
```
hooverlunch: Also don't need this I believe. At least for now.

app/models/ability.rb
```diff
@@ -175,6 +175,9 @@ def initialize(params)
 
           # there is no Questioning index
           cannot :index, Questioning
+
+          # there is no QingGroup index
+          cannot :index, QingGroup
```
hooverlunch: Great job catching this.

app/models/ability.rb
```diff
@@ -175,6 +175,9 @@ def initialize(params)
...
         end
 
         # Users can view/modify only their own API keys
```
hooverlunch: Great job on the ability file.

hooverlunch:
Great work! A few notes. Then you can go ahead and merge.

PR #46, opened 2015-02-27
2069 form items controller

mkuklis:
http://redmine.sassafras.coop/issues/2069

### DIFF
--- a/app/controllers/form_items_controller.rb
+++ b/app/controllers/form_items_controller.rb
@@ -7,7 +7,7 @@ class FormItemsController < ApplicationController
       return render json: { errors: ['ancestry_check_failed'] }, status: 422
     end
 
-    @form_item.ancestry = params[:parent_id]
+    @form_item.parent = FormItem.find(params[:parent_id])
     @form_item.rank = params[:rank]
 
     if @form_item.save
--- a/app/controllers/form_items_controller.rb
+++ b/app/controllers/form_items_controller.rb
@@ -3,12 +3,14 @@ class FormItemsController < ApplicationController
   load_and_authorize_resource
 
   def update
-    if params[:ancestry].blank? or !@form_item.check_ancestry_integrity(params[:ancestry])
+    attrs = params[:form_item]
+
+    if attrs[:ancestry].blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
       return render nothing: true, status: 422
     end
 
-    @form_item.ancestry = params[:ancestry]
-    @form_item.rank = params[:rank]
+    @form_item.ancestry = attrs[:ancestry]
+    @form_item.rank = attrs[:rank]
 
     if @form_item.save
       render nothing: true, status: 204
--- a/app/controllers/form_items_controller.rb
+++ b/app/controllers/form_items_controller.rb
@@ -2,26 +2,18 @@ class FormItemsController < ApplicationController
   # authorization via cancan
   load_and_authorize_resource
 
-  before_filter :prepare_form_item, only: [:update]
-
   def update
+    if params[:ancestry].blank? or !@form_item.check_ancestry_integrity(params[:ancestry])
+      return render nothing: true, status: 422
+    end
+
+    @form_item.ancestry = params[:ancestry]
+    @form_item.rank = params[:rank]
+
     if @form_item.save
       render nothing: true, status: 204
     else
       render nothing: true, status: 422
     end
   end
-
-  private
-    # prepares form item
-    def prepare_form_item
-      attrs = params[:form_item]
-
-      if attrs.blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
-        return render nothing: true, status: 422
-      end
-
-      @form_item.ancestry = attrs[:ancestry]
-      @form_item.rank = attrs[:rank]
-    end
 end
--- a/app/controllers/form_items_controller.rb
+++ b/app/controllers/form_items_controller.rb
@@ -3,19 +3,17 @@ class FormItemsController < ApplicationController
   load_and_authorize_resource
 
   def update
-    attrs = params[:form_item]
-
-    if attrs[:ancestry].blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
-      return render nothing: true, status: 422
+    if params[:parent_id].blank? or !@form_item.check_ancestry_integrity(params[:parent_id])
+      return render json: { errors: ['ancestry_check_failed'] }, status: 422
     end
 
-    @form_item.ancestry = attrs[:ancestry]
-    @form_item.rank = attrs[:rank]
+    @form_item.ancestry = params[:parent_id]
+    @form_item.rank = params[:rank]
 
     if @form_item.save
       render nothing: true, status: 204
     else
-      render nothing: true, status: 422
+      render @form_item.errors, status: 422
     end
   end
 end
--- a/app/controllers/form_items_controller.rb
+++ b/app/controllers/form_items_controller.rb
@@ -1,0 +1,27 @@
+class FormItemsController < ApplicationController
+  # authorization via cancan
+  load_and_authorize_resource
+
+  before_filter :prepare_form_item, only: [:update]
+
+  def update
+    if @form_item.save
+      render nothing: true, status: 204
+    else
+      render nothing: true, status: 422
+    end
+  end
+
+  private
+    # prepares form item
+    def prepare_form_item
+      attrs = params[:form_item]
+
+      if attrs.blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
+        return render nothing: true, status: 422
+      end
+
+      @form_item.ancestry = attrs[:ancestry]
+      @form_item.rank = attrs[:rank]
+    end
+end
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -134,10 +134,6 @@ class FormsController < ApplicationController
         # check special permissions
         authorize!(:rename, @form) if @form.name_changed?
 
-        # update ranks if provided (possibly raising condition ordering error)
-        # We convert IDs and ranks to integer before passing.
-        @form.update_ranks(Hash[*params[:rank].to_a.flatten.map(&:to_i)]) if params[:rank] && can?(:reorder_questions, @form)
-
         # save everything
         @form.save!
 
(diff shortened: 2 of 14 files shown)


app/controllers/form_items_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  def update
+    if @form_item.save
+      render nothing: true, status: 204
+    else
+      render nothing: true, status: 422
+    end
+  end
+
+  private
+    # prepares form item
```
hooverlunch: Probably not necessary.

app/models/form_item.rb
```diff
@@ -27,8 +31,26 @@ def arrange_and_sort(eager_load = nil)
...
+    parent.ancestry != self.id.to_s
+  end
+
   private
 
+    # sets rank if not already set
+    def set_rank
+      self.rank ||= (form.try(:max_rank) || 0) + 1
+      return true
+    end
```
hooverlunch: This method is not aware of the new hierarchical system. It should not be using form.max_rank since that returns the max rank for the topmost level questions only. Rather it should be looking on its own level. In general the old form rank methods are not to be trusted as they are not really aware of the hierarchical stuff.

app/models/form_item.rb
```diff
@@ -27,8 +31,26 @@ def arrange_and_sort(eager_load = nil)
...
+
+    # repair the ranks of the remaining questions on the form
+    def fix_ranks
+      form.fix_ranks
+    end
```
hooverlunch: I also am not convinced this method works.

spec/models/form_item_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      expect(@qing.check_ancestry_integrity(@qing_group.id)).to be_truthy
+    end
+
+    it "should return false" do
+      @qing_group.ancestry = @qing.id
+      @qing_group.save
+      expect(@qing.check_ancestry_integrity(@qing_group.id)).to be_falsy
+    end
+  end
+end
```
hooverlunch: I would like to see some tests here that really challenge the system:
1. If you move an item down, do the ranks adjust properly? Same if you move an item up.
1. If you change an item's parent, what happens to its rank?
1. If you change an item's rank and that invalidates a condition, what happens? (See Condition.verify_ordering method).
1. Each of these things should be tried both on the top level and on a sublevel.

I think if you write these tests you might find that there are bugs since we are using those old methods.

hooverlunch:
This is a great start! I think a bit more work is needed on data integrity. See my comments. Some of the older rank functions may need to be refactored. Let me know if you have questions.

app/controllers/form_items_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+  load_and_authorize_resource
+
+  def update
+    attrs = params[:form_item]
+
+    if attrs[:ancestry].blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
+      return render nothing: true, status: 422
+    end
+
+    @form_item.ancestry = attrs[:ancestry]
```
hooverlunch: It will be easier on the frontend if we just accept parent_id instead of ancestry. You should then be able to just do `@form_item.parent = FormItem.find(params[:parent_id])`. Ancestry will handle rest.

app/controllers/form_items_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @form_item.rank = attrs[:rank]
+
+    if @form_item.save
+      render nothing: true, status: 204
+    else
+      render nothing: true, status: 422
```
hooverlunch: Might as well do `render json: @form_item.errors, status: 422` so that it will be a little easier to debug.

app/controllers/form_items_controller.rb
```diff
@@ -0,0 +1,21 @@
+class FormItemsController < ApplicationController
+  # authorization via cancan
+  load_and_authorize_resource
+
+  def update
+    attrs = params[:form_item]
+
+    if attrs[:ancestry].blank? or !@form_item.check_ancestry_integrity(attrs[:ancestry])
+      return render nothing: true, status: 422
```
hooverlunch: Perhaps `render json: { errors: ['ancestry_check_failed'] }, status: 422`?

app/controllers/forms_controller.rb
```diff
@@ -115,7 +115,7 @@ def odk_itemsets
   def create
     @form.is_standard = true if current_mode == 'admin'
     if @form.save
-      @form.create_root_group!(mission: @form.mission, form: @form, rank: 1)
+      @form.create_root_group!(mission: @form.mission, form: @form)
```
hooverlunch: Yes!

app/controllers/forms_controller.rb
```diff
@@ -134,10 +134,6 @@ def update
         # check special permissions
         authorize!(:rename, @form) if @form.name_changed?
 
-        # update ranks if provided (possibly raising condition ordering error)
-        # We convert IDs and ranks to integer before passing.
-        @form.update_ranks(Hash[*params[:rank].to_a.flatten.map(&:to_i)]) if params[:rank] && can?(:reorder_questions, @form)
-
```
hooverlunch: Yes!

app/models/form.rb
```diff
@@ -79,9 +79,8 @@ def questions(reload = false)
   end
 
   def add_questions_to_top_level(questions)
-    max = max_rank
     questions.each_with_index do |q, i|
-      Questioning.create!(mission: mission, form: self, question: q, parent: root_group, rank: max + i + 1)
+      Questioning.create!(mission: mission, form: self, question: q, parent: root_group)
```
hooverlunch: Yes!

app/models/form.rb
```diff
@@ -186,10 +185,6 @@ def all_first_level_option_nodes
     OptionSet.first_level_option_nodes_for_sets(questions.map(&:option_set_id).compact)
   end
 
-  def max_rank
-    root_group.children.order(:rank).last.try(:rank) || 0
-  end
-
```
hooverlunch: Yes!

app/models/form.rb
```diff
@@ -225,9 +220,6 @@ def destroy_questionings(qings)
         qing.destroy
       end
 
-      # fix the ranks
-      fix_ranks(:reload => true, :save => true)
-
```
hooverlunch: Yes!

app/models/form.rb
```diff
@@ -311,14 +303,6 @@ def qing_answer_count(qing)
...
   end
 
-  # ensures question ranks are sequential]
-  def fix_ranks(options = {})
-    options[:reload] = true if options[:reload].nil?
-    options[:save] = true if options[:save].nil?
-    root_questionings(options[:reload]).sort_by(&:rank).each_with_index{|qing, idx| qing.update_attribute(:rank, idx + 1)}
-    save(:validate => false) if options[:save]
-  end
-
```
hooverlunch: Oh yes!

app/models/form_item.rb
```diff
@@ -1,5 +1,7 @@
 class FormItem < ActiveRecord::Base
-  include MissionBased, FormVersionable, Replication::Replicable
+  include MissionBased, FormVersionable, Replication::Replicable, RankedModel
+
+  ranks :rank, with_same: :ancestry, class_name: 'FormItem'
```
hooverlunch: Does it really need the class_name? It's already inside that class!

mkuklis: Yes I think this is needed to handle STI correctly. But I will double check.

mkuklis:
@hooverlunch I made more changes to this. Should be ready to go. LMK if you want me to merge it.

spec/models/condition_spec.rb
```diff
@@ -166,11 +166,11 @@
       end
 
       it 'should work' do
-        expect(@cond.to_s).to eq "Question #1 is less than 5"
+        expect(@cond.to_s).to eq "Question #0 is less than 5"
```
hooverlunch: Zoiks, this is one place where we need sensible ranks. These strings are shown to the user.

spec/models/form_item_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      expect(@f.c[2].rank).to be < @f.c[3].rank
+    end
+
+    it 'should assign a rank to a newly created group' do
+      expect(@f.c[2].rank).to be < @group.rank
+    end
+
+    it 'should set rank to 1 for existing questioning moved to the empty group' do
+      @qing = @f.c[0]
+      @qing.ancestry = "#{@f.root_group.id}/#{@group.id}"
```
hooverlunch: Curious if this all would work if you did `@qing.parent = @group`, which should be equivalent to what you put.

app/controllers/form_items_controller.rb
```diff
@@ -0,0 +1,19 @@
+class FormItemsController < ApplicationController
+  # authorization via cancan
+  load_and_authorize_resource
+
+  def update
+    if params[:parent_id].blank? or !@form_item.check_ancestry_integrity(params[:parent_id])
+      return render json: { errors: ['ancestry_check_failed'] }, status: 422
+    end
+
+    @form_item.ancestry = params[:parent_id]
```
hooverlunch: This won't quite do it. For one, I think it's bad to set ancestry directly because not sure if stuff like the depth cache would get set properly then. Also if the parent is not root, this won't work. Instead, try `@form_item.parent = FormItem.find(params[:parent_id])`?

spec/models/form_item_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+    @qing_group = @form.c[1]
+  end
+
+  describe "check_ancestry_integrity" do
+    it "should return true" do
+      expect(@qing.check_ancestry_integrity(@qing_group.id)).to be_truthy
+    end
+
+    it "should return false" do
+      @qing_group.ancestry = @qing.id
```
hooverlunch: Use parent= here?

spec/models/form_item_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+      @qing_group.save
+      expect(@qing.check_ancestry_integrity(@qing_group.id)).to be_falsy
+    end
+  end
+
+  describe "sort" do
+    before(:each) do
+      @f = create(:form, questions: ['text', 'text', 'text', 'text'])
+      @group = create(:qing_group, form: @f, ancestry: @f.root_group.id.to_s)
```
hooverlunch: Use parent= here too?

hooverlunch:
Fantastic work here. Please merge!

PR #47, opened 2015-03-03
1917 odk multiple questions

mkuklis:
http://redmine.sassafras.coop/issues/1917


app/models/form_item.rb
```diff
@@ -39,7 +39,7 @@ def _sorted_leaves(nodes)
         if form_item.is_a?(Questioning)
           form_item
         else
-          _sorted_leaves(children)
+          _sorted_leaves(children).flatten
```
hooverlunch: Will this prohibit multiple nesting in general? We don't want to do that I think. Only for now when we're rendering for ODK. The user should still be able to nest multiply when defining a form. There are other reasons this might be needed.

app/views/forms/show.xml.erb
```diff
@@ -101,29 +101,16 @@
...
-                  <value>on<%= node.id %></value>
-                </item>
-              <% end %>
-            <% end  %>
-          <% end %>
+          </group>
+        <% else %>
+          <%= render('odk_questioning', qing: qing, style: @style) unless qing.hidden %>
         <% end %>
       <% end %>
```
hooverlunch: This is nice and succinct. Well done.

hooverlunch:
This is also a really good start! Very impressive!

A key missing piece, though, is the grid-like rendering in ODK for a series of select-one questions with the same option set. You can see an example of this here: https://opendatakit.org/help/form-design/examples/ (scroll down to 'Select one list'). Click 'view snippet' to see how they did it.

It would also be good to have a test for this. We can discuss when we talk on the phone.

app/models/form_item.rb
mkuklis: @hooverlunch I don't think this will prohibit multiple nesting. The previous implementation already does it in the form model: https://github.com/thecartercenter/elmo/blob/develop/app/models/form.rb#L74 and makes the entire array flat. I added it here to get a structure similar to: 
````ruby
[q1, [q2, q3], q4]
````
the q2, q3 are in the array to indicate that they belong to the group(s).

This is used later in the view:

https://github.com/thecartercenter/elmo/blob/1917_ODK_multiple_questions/app/views/forms/show.xml.erb#L106

**Comment by hooverlunch** [comment]
One other thing I just remembered: the QingGroup.group_name needs to be displayed. I think ODK can do this. See the same example (the group name there is 'List Group' I think).

**Comment by hooverlunch** [comment]
Hi! This is much better! I took a look at this on my device. Here is what I got:

image

* I think we should not be rendering the hint in single screen mode. So please strip that out.
* I think the long question names and option names render pretty well! So happy with that. Will look better on a tablet too I think.

File: app/views/forms/show.xml.erb [hunk 3]
```diff
@@ -101,29 +101,25 @@
...
-          <%= odk_input_tag(qing, subq) do %>
-            <label ref="jr:itext('<%= subq.odk_code %>:label')"/>
-
-            <%# Commcare doesn't like hints. %>
-            <% unless @style == 'commcare' %>
-              <hint ref="jr:itext('<%= subq.odk_code %>:hint')"/>
+      <% @form.grouped_questionings.each do |qing| %>
+        <% if qing.is_a?(Array) %>
+          <group appearance="field-list">
+            <label><%= qing.first.parent.group_name_en %></label>
```
hooverlunch: This should just be `group_name`. The language will resolve it self.

app/views/forms/_odk_group.xml.erb
```diff
@@ -0,0 +1,20 @@
...
+  <% end %>
+
+  <% qing.each do |q| %>
+    <%= render('odk_questioning',
+      qing: q,
+      style: @style,
+      grid_mode: grid_mode,
+      options: {appearance: appearance(grid_mode)}) unless q.hidden %>
+  <% end %>
+</group>
```
hooverlunch: Newline.

app/helpers/odk_helper.rb
```diff
@@ -28,6 +29,18 @@ def required_value(form)
...
+    (grid) ? "list-nolabel" : nil
+  end
+
+  # tests if all items in the group are part of the grid
+  def grid_mode?(qings)
+    qings.inject(true) do |memo, q|
+      memo &= q.qtype.name == "select_one"
+      memo
+    end
+  end
```
hooverlunch: * I think this belongs in the QingGroup model.
* Also, you should be checking both that it's select_one and that the option set is the same.
* `inject` is nice, but here you could just do `qings.all?{ |q| q.qtype.name == 'select_one' && q.option_set == qings[0].option_set` }

app/helpers/odk_helper.rb
```diff
@@ -28,6 +29,18 @@ def required_value(form)
     form.allow_incomplete? ? "selected(/data/#{IR_QUESTION}, 'no')" : "true()"
   end
 
+  def appearance(grid)
+    (grid) ? "list-nolabel" : nil
+  end
```
hooverlunch: I think this can go away. See below.

app/views/forms/_odk_group.xml.erb
```diff
@@ -0,0 +1,20 @@
...
+  <label><%= qing.first.parent.group_name %></label>
+  <% grid_mode = grid_mode?(qing) %>
+
+  <% if grid_mode %>
+    <%= render('odk_questioning',
+      qing: qing.first,
+      style: style,
+      grid_mode: grid_mode,
+      options: {appearance: 'label'}) %>
+  <% end %>
```
hooverlunch: I see what you're doing here. And I like it. However, inside odk_questioning a lot of stuff disappears if you're in appearance label mode. See below. I think you can just stuff it right in here and not worrying about calling the partial.

app/views/forms/_odk_questioning.xml.erb
```diff
@@ -0,0 +1,26 @@
...
+<% qing.subquestions.each do |subq| %>
+  <%= odk_input_tag(qing, subq, options) do %>
+    <% if options[:appearance] == 'label' %>
+      <label>Choices</label>
+    <% else %>
+      <label ref="jr:itext('<%= subq.odk_code %>:label')"/>
+    <% end %>
+
+    <%# Commcare doesn't like hints. %>
+    <% unless style == 'commcare' or grid_mode %>
```
hooverlunch: We don't do hints in 'label' case anyway, so this can go altogether for label mode.

hooverlunch: Please use `||` instead of `or`.

app/views/forms/_odk_questioning.xml.erb
```diff
@@ -0,0 +1,26 @@
+<%# Note that subquestions here refers to multiple levels of a cascading select question, not question grouping. %>
+<% qing.subquestions.each do |subq| %>
```
hooverlunch: Please add a comment describing what label mode is since it's kind of cryptic.

app/views/forms/_odk_questioning.xml.erb
```diff
@@ -0,0 +1,26 @@
...
+  <%= odk_input_tag(qing, subq, options) do %>
+    <% if options[:appearance] == 'label' %>
+      <label>Choices</label>
+    <% else %>
+      <label ref="jr:itext('<%= subq.odk_code %>:label')"/>
+    <% end %>
+
+    <%# Commcare doesn't like hints. %>
```
hooverlunch: Please update or remove this comment.

app/views/forms/_odk_questioning.xml.erb
```diff
@@ -0,0 +1,26 @@
+<%# Note that subquestions here refers to multiple levels of a cascading select question, not question grouping. %>
+<% qing.subquestions.each do |subq| %>
+  <%= odk_input_tag(qing, subq, options) do %>
+    <% if options[:appearance] == 'label' %>
+      <label>Choices</label>
```
hooverlunch: This needs to be I18n'd please.

app/views/forms/_odk_group.xml.erb
```diff
@@ -0,0 +1,20 @@
+<group appearance="field-list">
+  <label><%= qing.first.parent.group_name %></label>
+  <% grid_mode = grid_mode?(qing) %>
+
+  <% if grid_mode %>
+    <%= render('odk_questioning',
+      qing: qing.first,
+      style: style,
+      grid_mode: grid_mode,
+      options: {appearance: 'label'}) %>
```
hooverlunch: This looks like an options hash inside an options hash?

Perhaps just `label_row: true` here, and then construct the options hash for the odk_input_tag method inside the partial? Else it's kind of confusing.

app/views/forms/_odk_group.xml.erb
```diff
@@ -0,0 +1,20 @@
...
+  <% end %>
+
+  <% qing.each do |q| %>
+    <%= render('odk_questioning',
+      qing: q,
+      style: @style,
+      grid_mode: grid_mode,
+      options: {appearance: appearance(grid_mode)}) unless q.hidden %>
```
hooverlunch: Same here. We already have grid_mode inside the partial, and `label_row` will be false by default. So then you can figure out whether/what appearance attrib inside the partial. then you can also nuke the `appearance` method.

app/views/forms/show.xml.erb
```diff
@@ -101,29 +101,16 @@
...
-          <% end %>
+      <% @form.grouped_questionings.each do |qing| %>
+        <% if qing.is_a?(Array) %>
+          <%= render('odk_group', qing: qing, style: @style, options: {}) %>
+        <% else %>
+          <%= render('odk_questioning',
+            qing: qing,
+            style: @style,
+            grid_mode: false,
+            options: {}) unless qing.hidden %>
```
hooverlunch: I think given the above, these options hashes on 107 and 113 can go away?

config/environments/development.rb
```diff
@@ -11,7 +11,7 @@
 
   # Show full error reports and enable caching
   config.consider_all_requests_local       = true
-  config.action_controller.perform_caching = true
+  config.action_controller.perform_caching = false
```
hooverlunch: I can see why you might do this temporarily but not sure if it's a good idea to commit it. Perhaps put it in your local_config.rb?

spec/requests/form_odk_rendering_spec.rb
```diff
@@ -5,26 +5,59 @@
...
+      doc = XML::Parser.string(response.body, options: XML::Parser::Options::NOBLANKS).parse
+      expect(doc.to_s).to eq File.read(File.expand_path('../../expectations/group_form_odk.xml', __FILE__))
+    end
   end
 
-  it 'should render proper xml' do
-    expect(response).to be_success
+  context 'grid form' do
+    before do
+      @form = create(:form, question_types: [['select_one', 'select_one']])
```
hooverlunch: This will need to change once you add the same-option-set requirement since this will create two different questions with different option sets. You could add a line beneath that does `@form.c[0].c[1].update_attributes(option_set: @form.c[0].c[0].option_set)`. (`c` is just a shorthand for `children`)

hooverlunch:
Getting close! See comments. Thanks!

mkuklis:
@hooverlunch I think the last commit should cover all of the comments.

hooverlunch:
Well done. I made a few small commits. Not sure why you didn't end up using `defined?` in the partial. Seems to me what you had (`label_row ||= label_row`) won't do anything and will still error if label_row is not explicitly set. Feel free to switch it back if I'm wrong, but do let me know so I can learn!

Great work on this! Please merge!

PR #55, opened 2015-04-01
2192 response form group awareness

mkuklis:
http://redmine.sassafras.coop/issues/2192


app/models/response.rb
```diff
@@ -271,7 +271,7 @@ def populate_from_j2me(data)
   # Groups answers by questioning.
   # Makes sure there are associated answer objects for each questioning in the form.
   def answer_sets
-    @answer_sets ||= visible_questionings.map{ |qing| answer_set_for_questioning(qing) }
+    @answer_sets ||= Hash[visible_questionings.map{ |qing| [qing, answer_set_for_questioning(qing)] }]
```
hooverlunch: I'm pretty sure you need to do it like `Hash[*foo.map{ |x| [x, bar(x)] }.flatten]` or you won't get what you want.

But even better in this situation might be to just map to the answer sets and then `index_by(&:qing)`?

app/views/responses/_answer_sets.html.erb
```diff
@@ -0,0 +1,12 @@
+<% items.each_with_index do |(item, subitems), lindex| %>
+  <% if item.is_a? QingGroup %>
+    <div class="form_field response_form_id">
+      <div class="label_and_control">
+        <label class="main"><%= item.group_name %></label>
+      </div>
+    </div>
+    <%= render 'answer_sets', items: subitems, answer_sets: answer_sets, level: level + 1, index: lindex %>
+  <% else %>
+    <%= render 'answer_set', answer_set: answer_sets[item], index: lindex + index, level: level %>
```
hooverlunch: `index` never gets set here for root-level questions as far as I can tell. I'm getting a nil error.

hooverlunch:
This isn't quite working for me yet. Please see comments. Thanks.

app/views/responses/_answer_sets.html.erb
mkuklis: Oopps missed it.

app/models/response.rb
mkuklis: Hmm this works fine for me. Please give it a try again. I just fixed the index issue.

mkuklis:
I found one more issue related to index. I'm looking looking into it right now.

mkuklis:
Ok @hooverlunch I made couple more small tweaks. It looks like things are working fine now.

PR #56, opened 2015-04-02
2191 form summary report group awareness

mkuklis:
http://redmine.sassafras.coop/issues/2191


hooverlunch:
This looks good but I can't test it until 2192 is done since I can't create any responses to a form with groups. So leaving this here for a bit.

PR #60, opened 2015-04-22
2163: Only show admins in dropdown if they are the submitter of a response

md5:
<s>This PR changes the `User.assigned_to_or_admin` scope to only include admins who have submitted a response to a mission.</s>

Let me know if you want me to look at optimizing the queries in this scope; I'd possibly look at using a correlated subquery with `EXISTS`.

### DIFF
--- a/db/migrate/20150422182449_remove_unassigned_users_from_whitelist.rb
+++ b/db/migrate/20150422182449_remove_unassigned_users_from_whitelist.rb
@@ -1,0 +1,18 @@
+class RemoveUnassignedUsersFromWhitelist < ActiveRecord::Migration
+  def up
+    execute <<-SQL
+      DELETE wl
+      FROM whitelists wl
+        INNER JOIN forms f
+          ON wl.whitelistable_type = 'Form' AND wl.whitelistable_id = f.id
+        INNER JOIN missions m
+          ON f.mission_id = m.id
+      WHERE NOT EXISTS (
+        SELECT 1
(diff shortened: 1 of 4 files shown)


hooverlunch:
Sorry the spec for this one wasn't super clear.

We want to remove admins from both of these dropdowns completely.

The only exception is with the response form, if the admin in question is the submitter of the currently displayed form, their name *should* be included in the submitter dropdown. Since otherwise it would be blank.

Come to think of it, there should also be a migration that removes api permissions for any users that are not members of the form's mission, since otherwise they would not be editable after this change. Should be achievable with one query.

md5:
@hooverlunch Updated. Should be what you're looking for.

I'll work on a migration next.

md5:
Actually, I think I'll change this to avoid checking the `responses` table.

db/migrate/20150422182449_remove_unassigned_users_from_whitelist.rb
```diff
@@ -0,0 +1,18 @@
...
+          ON f.mission_id = m.id
+      WHERE NOT EXISTS (
+        SELECT 1
+        FROM assignments a
+        WHERE m.id = a.mission_id
+        AND wl.user_id = a.user_id
+      )
+    SQL
+  end
+end
```
hooverlunch: Lovely.

hooverlunch:
Looks great. You can go ahead and merge. Thanks.

PR #61, opened 2015-04-24
203: Request throttling for ODK Collect endpoints

md5:
This PR implements the request throttling portion of issue 203 for the ODK Collect endpoints.

I haven't yet made it configurable, but that would involve using a `proc` for `:limit` (to ensure that the config has been fully loaded). Let me know if you actually want this to be configurable.

I also added a spec.

One thing I wasn't sure about was whether to make `direct_auth?` match paths without `.xml`. There are tests in `basic_authentication_spec.rb` that aren't using `.xml`, so I'm not sure what's actually being requested by ODK Collect. I also recall @hooverlunch mentioning that there is a potentially conflicting path for `/forms/:id`.

### DIFF
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -1,13 +1,8 @@
 class Rack::Attack::Request
   def direct_auth?
+    # Match paths starting with "/m/mission_name", but exclude "/m/mission_name/sms" paths
     if path =~ %r{^/m/[a-z][a-z0-9]*/(.*)$}
-      subpath = $1
-
-      return true if subpath == 'formList.xml'
-      return true if subpath =~ %r{^forms/([^/]+)\.xml$}
-      return true if subpath =~ %r{^forms/([^/]+)/manifest\.xml$}
-      return true if subpath =~ %r{^forms/([^/]+)/itemsets\.csv$}
-      return true if subpath == 'submission.xml'
+      return $1 !~ /^sms/
     end
   end
 end
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -1,0 +1,20 @@
+class Rack::Attack::Request
+  def direct_auth?
+    if path =~ %r{^/m/[a-z][a-z0-9]*/(.*)$}
+      subpath = $1
+
+      return true if subpath == 'formList.xml'
+      return true if subpath =~ %r{^forms/([^/]+)\.xml$}
+      return true if subpath =~ %r{^forms/([^/]+)/manifest\.xml$}
+      return true if subpath =~ %r{^forms/([^/]+)/itemsets\.csv$}
+      return true if subpath == 'submission.xml'
+    end
+  end
+end
+
+class Rack::Attack
+  # Limit ODK Collect and other :direct_auth requests by IP address to 30 requests per minute
+  throttle('req/ip', limit: 30, period: 1.minute) do |req|
+    req.ip if req.direct_auth?
+  end
+end
--- a/config/initializers/rack-attack.rb
+++ b/config/initializers/rack-attack.rb
@@ -13,8 +13,8 @@ class Rack::Attack::Request
 end
 
 class Rack::Attack
-  # Limit ODK Collect and other :direct_auth requests by IP address to 30 requests per minute
-  throttle('req/ip', limit: 30, period: 1.minute) do |req|
+  # Limit ODK Collect requests by IP address to N requests per minute
+  throttle('req/ip', limit: proc { configatron.direct_auth_request_limit }, period: 1.minute) do |req|
     req.ip if req.direct_auth?
   end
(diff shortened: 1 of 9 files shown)


hooverlunch:
Good question about paths. 

The only path that will have a .xml ending is /m/mission1/forms/123.xml

All the other direct auth paths will not.

Another way to distinguish them perhaps is by the fact that they have no locale.

If you can think of some better way of organizing this feel free. It's just that we have no control over the /formList or /submission. Anything else we can change.

config/initializers/rack-attack.rb
```diff
@@ -0,0 +1,20 @@
+class Rack::Attack::Request
+  def direct_auth?
+    if path =~ %r{^/m/[a-z][a-z0-9]*/(.*)$}
+      subpath = $1
+
+      return true if subpath == 'formList.xml'
+      return true if subpath =~ %r{^forms/([^/]+)\.xml$}
+      return true if subpath =~ %r{^forms/([^/]+)/manifest\.xml$}
+      return true if subpath =~ %r{^forms/([^/]+)/itemsets\.csv$}
+      return true if subpath == 'submission.xml'
```
hooverlunch: So remove .xml from all but line 7.

spec/requests/request_throttling_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:limit) { 30 }
+
+  before(:each) do
+    Rack::Attack.cache.store = ActiveSupport::Cache::MemoryStore.new
+  end
+
+  context 'with the same ip address' do
+    it 'should not apply to requests below the limit' do
+      limit.times do
+        get "/m/#{get_mission.compact_name}/formList.xml"
```
hooverlunch: Should change the .xmls down here too.

spec/requests/request_throttling_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+
+describe 'throttling for xml requests' do
+
+  let(:limit) { 30 }
```
hooverlunch: I'm concerned about how long this test will take to run. Could we stub the limit to like 3 or 4 just for the test case?

spec/requests/request_throttling_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it 'should not apply' do
+      (limit * 2).times do |i|
+        get "/m/#{get_mission.compact_name}/formList.xml", nil, { REMOTE_ADDR: remote_addrs[i] }
+        assert_response :unauthorized
+      end
+    end
+  end
+
+end
```
hooverlunch: Good job on the test.

config/routes.rb
```diff
@@ -131,6 +131,11 @@
     get '/sms/submit' => 'sms#create'
```
hooverlunch: Can we move the SMS routes into a separate scope block, then your comment below can apply to the whole ODK scope block.

The SMS stuff is going to change soon too, with the token ticket that is upcoming.

Also we could perhaps put direct_auth: true in the defaults for the whole block to simplify things. The condition in authentication.rb would still work even for the the no_auth one. We could consider no_auth a form of direct_auth, if you catch my drift.

hooverlunch:
Nice work. Just a few small things and you can merge.

spec/requests/request_throttling_spec.rb
md5: Did you try running it? It runs in less than a second.

```
$ bundle exec rspec spec/requests/request_throttling_spec.rb

Randomized with seed 35489

throttling for xml requests
  with the same ip address
    should not apply to requests below the limit
    should apply to requests above the limit with response code 429
  with different ip addresses
    should not apply

Finished in 5.18 seconds (files took 4.3 seconds to load)
3 examples, 0 failures

Randomized with seed 35489
```

md5: Also, the limit here has to match the limit in `rack-attack.rb` for the test to work right. If I were to do the `proc` for `:limit`, both values could be pulled from config.

md5:
> The only path that will have a .xml ending is /m/mission1/forms/123.xml
>
> All the other direct auth paths will not.

I'll update the code to reflect that.

> Another way to distinguish them perhaps is by the fact that they have no locale.

Since this middleware runs in front of the Rails stack, I don't think we have access to a locale yet.

md5:
Oh, wait, you mean the locale in the URL. I'll think about whether that can simplify this.

hooverlunch:
You already take this into account via your regexp I believe. So all is well on this front I think.

md5:
I've updated this to include a negative test to make sure SMS requests aren't being throttled by `Rack::Attack`.

This has ended up being a bunch of commits, so let me know if you'd like me to squash them.

hooverlunch:
Looks great!

PR #62, opened 2015-04-27
2167: Incoming sms spec

md5:
This is a quick and dirty conversion that keeps most of the existing `assert` calls instead of moving to `expect` style.


hooverlunch:
Looks great. Just one thing: please convert hashes to new style on any such refactors. On Textmate2 you can just select all and do Ctrl+Shift+=. Hopefully your editor has similar capabilities.

Thanks!

md5:
Updated the hash syntax. I also amended the commits to refer to issue 2167 instead of 203. Branch name will have to stay unless you want a new PR.

PR #63, opened 2015-04-28
203: Require reCAPTCHA on excessive login attempts

md5:
This PR enables support for requiring a reCAPTCHA when an excessive number of login attempts are made.

In the current implementation, both `GET` requests to `/xx/login` and all requests to `/xx/user-session/**` will count toward the threshold. Doing otherwise would require some sort of explicit check against the `rack-attack` tracking data stored in `Rack::Attack.cache`.

This will also need to be localized, particularly the `:verify_login` label on the login page and the reCAPTCHA itself (achievable with the `hl` parameter to `recaptcha_tags`).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -77,6 +77,9 @@ gem 'rails-backbone', github: 'codebrew/backbone-rails'
 # Middleware for handling abusive requests
 gem 'rack-attack'
 
+# reCAPTCHA support
+gem "recaptcha", :require => "recaptcha/rails"
+
 # XLS support
 gem 'roo'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -237,6 +237,7 @@ GEM
     rake (10.4.2)
     random_data (1.6.0)
     rdiscount (2.1.8)
+    recaptcha (0.4.0)
     ref (1.0.5)
     request_store (1.1.0)
     responders (2.1.0)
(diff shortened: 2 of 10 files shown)


hooverlunch:
Great work! Few notes:

* Add keys to local_config.rb.example with a nice comment saying what recaptcha is and how to get a key.
* Can you also make it throw a fatal error on startup if you don't have recaptcha keys. We want to make sure this works.
* Please add verify_login to en.yml. The English translation should be 'Verify'.
* Currently the error on failure is "Error: Verify login: Word verification response is incorrect, please try again." Can it just be "Error: Word verification response is incorrect, please try again.", and can this please be in en.yml.

Then you can go ahead and merge.

This is awesome! Folks will be so impressed with this!

md5:
Updated per your comments. If you don't mind giving it another once-over, I'd appreciate it. I ended up rebasing the some of the changes and it could probably use another pair of eyes.

Also, let me know if you're seeing what seem to be spurious failures. I'm not sure if my reCAPTCHA key is having trouble or what, but I'm seeing failures sometimes when the reCAPTCHA says it's been verified.

hooverlunch:
The changes look good. I did not notice the failures but please be sure it's working. When you're satisified, go ahead and merge. Let me know if you need help rebasing.

PR #64, opened 2015-05-01
2167: Add incoming SMS gateway token

md5:
This PR changes the handling of incoming SMS messages to require an authentication code as part of the incoming URL. This is used to prevent unauthorized submission of SMS messages by malicious third parties.

@hooverlunch This does most of what we discussed for this feature, but I did not yet implement the loading indicator and confirmation functionality we discussed. I'm opening the PR so you can start looking at the changes.

Update: Loading indicator and confirmation have been added.


md5:
@hooverlunch I pushed a commit to use the AJAX treatment for `Setting.override_code`, but it was just a copy-pasta job at this point.

I don't think that the `options[:after]` thing is the right way to deal with combining the success/failure indicators with the existing `loading_indicator`, but I'm not sure whether to actually extend `loading_indicator` or to introduce something like a `status_indicator` with loading, success, and failure states. As I mentioned earlier today, the fact that no details about the failure are shown in the failure case makes it harder to justify in the case of a generalized component as opposed to the specific case of regenerating the `incoming_sms_token` where failures should be extremely rare.

I also noticed that the `error` callback doesn't get called in the case of a 404. I didn't track it down fully, but I think it may be because the 404 doesn't give a JSON response.

app/models/setting.rb
```diff
@@ -70,6 +72,25 @@ def generate_override_code!(size = 6)
     self.save!
   end
 
+  def generate_incoming_sms_token replace=false
```
hooverlunch: We generally put parenths around method args.

app/views/settings/_override_code.html.erb
```diff
@@ -1,5 +1,11 @@
...
-  <input class="submit btn btn-default btn-sm" name="regenerate"
-    data-confirm="<%= t('setting.override_code_warning') %>" type="submit" value="<%= t('common.regenerate') %>">
+  <input class="regenerate btn btn-default btn-xs"
+    data-confirm="<%= t('setting.override_code_warning') %>"
+    data-display-id="display-override_code"
+    data-handler="regenerate_override_code"
+    type="button" value="<%= t('common.regenerate') %>">
+  <%= loading_indicator success_failure: true %>
 <% end %>
+</div>
```
hooverlunch: This whole thing could perhaps be generalized into like a 'RegenerableField' thingie. The JS and the ERB.

md5:
Take a look at the latest commit. I created a `regenerable_field` method on `ElmoFormBuilder` and a corresponding `ELMO.Views.RegenerableFieldView` Backbone model.

Let me know if `app/views/layouts/application.html.erb` was the right place to include the Backbone view, since I wasn't sure whether to somehow try to include it directly from `f.regenerable_field` (e.g. by adding a `<script>` tag in the output of that method).

I still haven't updated `User.api_key` to use `f.regenerable_field`, but I should have some time tomorrow night or in the next few days to make that change as well. I'm just as happy to make a separate PR for that change if you're amenable.

md5:
Updated again to use `regenerable_field` for `User.api_key`. Making the field work with the new helper literally took two minutes.

Fixing the spec took a bit longer, but everything is now passing for me except a Thinking Sphinx-related spec ("response full text searches should work"). That spec fails for me on `develop` too, so I don't think it's related to my changes.

app/models/setting.rb
```diff
@@ -70,6 +72,25 @@ def generate_override_code!(size = 6)
...
+      return
+    end
+
+    # Ensure that the new token is actually different
+    begin
+      new_token = SecureRandom.hex
+    end while new_token == self.incoming_sms_token
+
+    self.incoming_sms_token = new_token
+  end
```
hooverlunch: Our styleguide says to only use `self` when required. So please remove these if you would. Thanks.

app/views/layouts/application.html.erb
```diff
@@ -36,6 +36,7 @@
     )%>);
 
     new ELMO.Views.MissionChangeDropdown({el: '#change_mission select'});
+    new ELMO.Views.RegenerableFieldView();
```
hooverlunch: Can/should we just put this on pages where it's required?

config/routes.rb
```diff
@@ -127,8 +135,7 @@
 
   # Special SMS routes. No locale.
   scope '/m/:mission_name', mission_name: /[a-z][a-z0-9]*/, defaults: { mode: 'm'} do
-    resources :sms, only: [:create]
-    get '/sms/submit' => 'sms#create'
+    match '/sms/submit/:token' => 'sms#create', token: /[0-9a-f]{32}/, via: [:get, :post], as: :mission_sms_submission
```
hooverlunch: Smooth.

hooverlunch:
This is so swanky, I love it. Great work. Client will love it too. Just a few small notes and then please merge.

app/views/layouts/application.html.erb
md5: I actually had a question about that buried in one of my comments:

> Let me know if app/views/layouts/application.html.erb was the right place to include the Backbone view, since I wasn't sure whether to somehow try to include it directly from f.regenerable_field (e.g. by adding a `<script>` tag in the output of that method).

The alternative would be that a developer would add `f.regenerable_field` to a page and have to remember to add `new ELMO.Views.RegenerableFieldView()` themselves, which seems like a recipe for frustration.

hooverlunch: Ah, good point. I think it would work to put it in the method output as you suggest?

md5: But I'd somehow have to avoid instantiating more than one per page, right?

I suppose we could move to passing the element of each individual "Regenerate" field to its own `RegenerableFieldView` instance.

md5: That probably makes more sense anyways, since it's a view for the field, not the whole page.

PR #65, opened 2015-05-16
2168: Add Twilio SMS adapter

md5:


md5:
I should note two things about this PR.

1. The `check_balance` functionality is not implemented because Twilio does not offer an API call to get the balance. I sent a support email to them and they said that they have an internal planning ticket open for the feature and that they'd log our interest there. I told them about the ELMO use case and they seemed eager to support the sort of work ELMO is used for, so perhaps they'll prioritize the feature.
2. I realized that the fact that the SMS submission URLs is going to mean that a different incoming number is required for each mission. That was already the case before the `incoming_sms_token` changes though, but I thought I'd mention it.

app/helpers/sms_helper.rb
```diff
@@ -19,8 +19,9 @@ def format_sms_messages_field(sms, field)
       recips = sms.recipient_hashes(max: MAX_RECIPS_TO_SHOW).map { |r| user_with_phone(r[:user], r[:phone]) }.join('<br/>')
       extra_recipients = sms.recipient_count - MAX_RECIPS_TO_SHOW
       recips << (extra_recipients > 0 ? t('sms.extra_recipients', count: extra_recipients) : '')
+      recips.html_safe
     when "from" then
-      user_with_phone sms.sender, sms.from
+      user_with_phone(sms.sender, sms.from).html_safe
```
hooverlunch: Are we sure this is not an XSS vector? What if someone puts evil stuff in their user name?

app/models/sms/adapters/intelli_sms_adapter.rb
```diff
@@ -3,9 +3,9 @@
 class Sms::Adapters::IntelliSmsAdapter < Sms::Adapters::Adapter
 
   # checks if this adapter recognizes an incoming http receive request
-  def self.recognize_receive_request?(params)
+  def self.recognize_receive_request?(request)
```
hooverlunch: Nice refactor here.

app/models/broadcast.rb
```diff
@@ -3,6 +3,7 @@ class Broadcast < ActiveRecord::Base
 
   has_many(:broadcast_addressings, :inverse_of => :broadcast, :dependent => :destroy)
   has_many(:recipients, :through => :broadcast_addressings, :source => :user)
+  has_many(:sms_messages, :dependent => :destroy, :class_name => 'Sms::Message')
```
hooverlunch: Curious why this got added?

app/views/broadcasts/form.html.erb
```diff
@@ -2,7 +2,7 @@
 
   <%= top_action_links(@broadcast, :only => :index) %>
 
-  <%= f.field(:recipient_ids, :type => :hidden) %>
+  <%= f.field(:recipient_ids, :type => :hidden, :multiple => true) %>
```
hooverlunch: Interesting.

config/initializers/filter_parameter_logging.rb
```diff
@@ -1 +0,0 @@
-Rails.application.config.filter_parameters += [:password, :password_confirmation]
```
hooverlunch: Thanks for cleaning this up.

app/views/settings/index.html.erb
```diff
@@ -26,12 +26,21 @@
...
         <%= f.field(:intellisms_password2, :type => :password) %>
       </div>
     </div>
 
+    <div class="adapter_settings" data-adapter="Twilio">
+      <%= f.field(:twilio_account_sid, :append => link_to(t("setting.change_twilio_auth_token"), "#")) %>
+
+      <div class="credential_fields">
+        <%= f.field(:twilio_auth_token1, :type => :password) %>
+        <%= f.field(:twilio_auth_token2, :type => :password) %>
```
hooverlunch: Not sure if we need re-entry for auth tokens. Typically I think there is just one box since everyone is copy/pasting.

config/locales/en.yml
```diff
@@ -322,7 +323,9 @@ en:
         override_code: "Override Code"
         external_sql: "External SQL"
         timezone: "Time Zone"
-        incoming_sms_token: "Incoming SMS Token"
+        twilio_account_sid: "Account SID"
+        twilio_auth_token1: "Auth Token"
+        twilio_auth_token2: "Auth Token Confirmation"
```
hooverlunch: So this and the hint below could probably go.

spec/models/sms/adapters/twilio_sms_adapter_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+  it 'should be created by factory' do
+    expect(@adapter).to_not be_nil
+  end
+
+  it 'should have correct service name' do
+    expect(@adapter.service_name).to eq 'Twilio'
+  end
+
+  it 'should return true on deliver' do
+    msg = Sms::Reply.new(:to => '+123', :body => 'foo')
```
hooverlunch: Please use new hash style in new files.

spec/models/sms/adapters/twilio_sms_adapter_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+      expect(msg.to).to eq '+123456789'
+      expect(msg.from).to eq '+2348036801489'
+      expect(msg.body).to eq 'foo'
+      expect(msg.adapter_name).to eq 'Twilio'
+      expect(msg.sent_at.utc).to eq sent_at
+      expect(msg.mission).to be_nil # This gets set in controller.
+    end
+  end
+
+
```
hooverlunch: Extra newline.

hooverlunch:
Looks amazing. Just a couple notes in code. Plus:

* Can you look into whether the incoming sms number needs to be normalized? Currently it will take different formats e.g. (555) 555-1212. Want to make sure it doesn't make some adapters choke.
* Some missions may not have an incoming number. Twilio doesn't seem to like outgoing messages without from numbers. What to do? Can we put e.g. ELMO as the from number if one isn't given? I think you can sometimes use textual ones.

Other than that, all good.

app/models/broadcast.rb
md5: I was trying to do `Broadcast.where(...).destroy_all` in the console and it was failing. I can drop it if you like.

spec/models/sms/adapters/twilio_sms_adapter_spec.rb
md5: Fixed in e47b8da70ddd8cc56e2b64f313ec17c345cc94f9

spec/models/sms/adapters/twilio_sms_adapter_spec.rb
md5: Fixed in e47b8da70ddd8cc56e2b64f313ec17c345cc94f9

app/views/broadcasts/form.html.erb
md5: Yeah. This basically just adds `[]` to the end of the `<input>` name. I was expecting it to also create a single `<input>` tag per value, but instead it uses a single comma-separated `value`. Hence the changes to `recipient_ids=` in `app/model/broadcast.rb`.

app/views/settings/index.html.erb
md5: I actually originally implemented it as a simple field before deciding that it needed the confirmation/secret treatment. I guess I should have checked first... I'll revert that.

app/helpers/sms_helper.rb
md5: I'll fix better :+1:

md5: I don't think it was actually vulnerable to an XSS attack previously because `user.name` was being escaped in `link_to` and `user.phone` is being cleaned up in the model layer.

Still, I've gone through and made this code safer against future XSS vectors.

See 12a767f1850e95bbdca1b6c42410d2b94b91b82a

app/models/broadcast.rb
hooverlunch: Perhaps best. These days I try to restrict associations to those needed since each one adds about 9 billion methods to the class' public interface.

md5:
> Can you look into whether the incoming sms number needs to be normalized? Currently it will take different formats e.g. (555) 555-1212. Want to make sure it doesn't make some adapters choke.

I'll look at this next. Twilio definitely expects the numbers to be in E.164 format. They suggest the `phony` gem for normalizing phone numbers. Do you have a preference on how and where to normalize?

> Some missions may not have an incoming number. Twilio doesn't seem to like outgoing messages without from numbers. What to do? Can we put e.g. ELMO as the from number if one isn't given? I think you can sometimes use textual ones.

This is correct that Twilio requires a `From` number for outbound messages. I'm not sure what you mean about a textual identifier, but here's what the docs say:

> A Twilio phone number enabled for the type of message you wish to send. Only phone numbers or short codes purchased from Twilio work here; you cannot (for example) spoof messages from your own cell phone number.

md5:
Updated: https://github.com/thecartercenter/elmo/compare/f399a17a041a6117b9db72af79032b1560709387...db86d244737d16dbf37fa41e8b8003f937f68e20

Please take a look.

PR #66, opened 2015-05-22
204: Password reprompt for sensitive areas

md5:


app/controllers/concerns/application_controller/authentication.rb
```diff
@@ -23,9 +23,7 @@ def get_user
     # If user already logged in via Authlogic, we are done.
     if (user_session = UserSession.find) && user_session.user
 
-      # Look up the current user from the user session
-      # We use a find call to the User class so that we can do eager loading
-      @current_user = User.includes(:assignments).find(user_session.user.id)
+      @current_user = user_session.user
```
hooverlunch: Rationale for reverting this? Just curious.

app/controllers/user_sessions_controller.rb
```diff
@@ -22,7 +22,7 @@ def create
       post_login_housekeeping
     else
       flash[:error] = @user_session.errors.full_messages.join(",")
-      redirect_to(:action => :new)
+      redirect_to(:action => new)
```
hooverlunch: Typo?

app/controllers/concerns/application_controller/authorization.rb
```diff
@@ -43,4 +43,15 @@ def handle_access_denied(exception)
...
     end
   end
-end
\ No newline at end of file
+
+  def require_recent_login(options={})
+    unless current_user && current_user.current_login_recent?(options[:max_age])
+      raise RecentLoginRequiredError
+    end
+  end
```
hooverlunch: Also curious why you did it this way (raising an error) instead of just calling handle_recent_login_required directly. I'm sure there is a reason, just curious what it is...

app/controllers/settings_controller.rb
```diff
@@ -55,6 +55,11 @@ def using_incoming_sms_token_message
   end
 
   private
+    def authorize!(*args)
+      super
+      require_recent_login
+    end
```
hooverlunch: Hmm, just noticed this. I am definitely puzzled by this so I think we need a big fat comment here.

app/models/user_session.rb
```diff
@@ -5,6 +5,13 @@ class UserSession < Authlogic::Session::Base
   allow_http_basic_auth(false) # We handle our own basic auth
   httponly(true)
 
+  # override find() to eager load User.assignments
+  def self.find(*args)
+    with_scope(find_options: User.includes(:assignments)) do
+      super
+    end
```
hooverlunch: Oh I see what you've done here now.

hooverlunch:
Impeccable work. I love it. Just a few notes and questions. I pushed one typo fix (please confirm I was right about it being a typo).

Thanks!

app/controllers/user_sessions_controller.rb
md5: Thanks for catching that. I thought I had fixed that...

app/controllers/concerns/application_controller/authorization.rb
md5: I think I mainly did that because it aligned with how Authlogic and Cancan are doing things. Changing it to call `handle_recent_login_required` directly would be fine too.

hooverlunch: It's fine either way. Just please put a comment.

app/controllers/settings_controller.rb
md5: I did this one the way I did because `load_and_authorize_resource` is not being called. I guess there wouldn't be any danger of checking for a recent login before `authorize!` is called (either by `load_and_authorize_resource` for `UsersController` and `UserBatchController` or directly as in `SettingsController`), but I've been thinking of this as a three-part check: 1) Authentication, 2) Authorization, 3) Stale session check.

This probably gets back to the fact that I originally tried to implement this by adding extra checks at the end of `UserAbility.initialize`. The problem with that approach was that it made `can?` start returning `false` for stale logins, so the UI elements to perform sensitive actions weren't even being rendered.

I then started down the path of overrideing `authorize!` in `ApplicationController` similar to what I've done here in `SettingsController`, but I abandoned that because I thought that having `load_and_authorize` do more magical stuff might be confusing. An explicit `before_action` for checking stale logins seemed easier to maintain.

As for returning a 4XX status code in the `xhr?` case, that sounds reasonable. I'd probably return a 401 instead of a 403 since this is a type of authorization failure, but either one is fine probably.

PR #67, opened 2015-05-22
2215: Implemented delayed delta indexing

md5:
This PR adds the [`ts-delayed-delta`](https://github.com/pat/ts-delayed-delta/) gem to allow Thinking Sphinx deltas to be processed via `delayed_job`.

I've also updated the Capistrano scripts to fix a couple bugs in `staging.rb` as well as to start DJ along with the main Nginx process.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,6 +61,7 @@ gem 'ejs'
 gem 'thinking-sphinx', '~> 3.1.3'
 
 # search: delayed deltas
+gem 'daemons'
 gem 'delayed_job_active_record'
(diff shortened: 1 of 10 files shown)


md5:
One thing I noticed but didn't fix was that `thinking_sphinx:rebuild` is tied to `deploy` instead of `deploy:restart` or `deploy:start`. This makes so that a TS rebuild is not kicked off when `deploy:migrations` is used, since it doesn't actually call `deploy`.

hooverlunch:
I'm assuming this would change the production setup procedure if one were not using Capistrano? Could you update `production-setup.md` to reflect?

PR #70, opened 2015-05-27
2372: Add Travis CI

md5:
Got this working after some trial and error. It has two failing specs, but the same ones fail for me locally.

cf. https://travis-ci.org/thecartercenter/elmo/builds/64334800

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,10 @@
+---
 bundler_args: --without development
 language: ruby
+cache: bundler
 rvm:
   - 2.2.0
-# TODO: Change this to use addons.apt once sphinxsearch is whitelisted
-before_install:
-  - sudo apt-get update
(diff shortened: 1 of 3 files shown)


md5:
I got the response tally test working in 71a28cd8d7e19a57f1215e702dcca69682e8a3d0. The issue there was that the MySQL database didn't have `mysql.time_zone` loaded. The same issue was the source of my problems with the test failing locally (and I fixed it the same way).

As for `send_broadcast_spec`, I was able to get that working locally by reverting the changes to `BroadcastsController` in b3380d4d0e69302a985a1a5800f2bb9ec59bbacc#diff-4561899c68a75e12c885f914be4812f0. It seems like there's been some flap in that file due to different things failing for different people, so it would be nice to sort that out.

PR #72, opened 2015-05-29
2342: Fix mission unlocking

md5:
Also fixes a failure when viewing the forms view for a locked mission in mission mode.


md5:
The Travis failure on the [merged build](https://travis-ci.org/thecartercenter/elmo/builds/64608806) looks spurious. It doesn't fail in the [push build](https://travis-ci.org/thecartercenter/elmo/builds/64608564) or when I run it locally.

PR #74, opened 2015-06-02
892: Scalable user selection

md5:
Uses the [Select2](https://select2.github.io/) library with remote data loading for the `Response.user_id` dropdown.

Also adds a `select2` helper for Capybara that should be able to replace `select` as needed.


md5:
Closing this while I do some more work to check that authorization is working correctly in the non-admin case.

md5:
After looking through the code, I realized that the "User/Team" dropdown is only shown if the user `can?(:manage, Response)`, so the permissions are already working correctly.

PR #75, opened 2015-06-09
854: Batch delete users

md5:
This could still be improved a bit, but I believe it's working as expected. One think in particular that could be improved would be to move the extensive model-related code in `UsersController#bulk_destroy` into the `User` model. The reason I didn't do this is that it seemed weird for the `User` model to have both the Rails-provided `destroy_all` and `delete_all` as well as a `bulk_destroy` or similarly named method.


md5:
FYI, fixed the spec failure with commit 421041527ddf443b98b5757592cfe625802152b9

app/assets/javascripts/backbone/views/index_table_view.js.coffee
```diff
@@ -18,7 +18,7 @@ class ELMO.Views.IndexTableView extends Backbone.View
     @select_all_field = this.$el.find('input[name=select_all]')
     @alert = this.$el.find('div.alert')
     @pages = this.$el.data('pages')
-    @count = this.$el.data('entries')
+    @entries = this.$el.data('entries')
```
hooverlunch: Zomg I love that this is a BB view now!

app/models/ability.rb
```diff
@@ -167,7 +167,7 @@ def initialize(params)
             can :manage, UserBatch
 
             # can destroy users only if they have only one mission and it's the current mission
-            can :destroy, User do |other_user|
+            can [:bulk_destroy, :destroy], User do |other_user|
```
hooverlunch: Doesn't this mean that coordinators won't be able to destroy individual users anymore?

hooverlunch: Not sure if we need a separate ability for bulk_destroy. What is your reasoning here?

hooverlunch:
Looks fantastic. Please merge.

PR #77, opened 2015-06-12
891: Scalable mapping

md5:
This PR updates the dashboard map to work better for missions with large numbers of responses.

1. Uses [`MarkerClusterer`](https://github.com/googlemaps/js-marker-clusterer) to group together responses
2. Uses a smaller icon for individual responses
3. Only shows the most recent 1,000 responses


app/assets/javascripts/backbone/views/dashboard_map_view.js.coffee
```diff
@@ -0,0 +1,117 @@
...
+    @map = new google.maps.Map($('div.response_locations')[0], {
+      mapTypeId: google.maps.MapTypeId.ROADMAP,
+      zoom: 1,
+      streetViewControl: false,
+      draggableCursor: 'pointer'
+    })
+
+    # create the marker clusterer
+    mc = new MarkerClusterer(@map, [], {
+      imagePath: 'https://google-maps-utility-library-v3.googlecode.com/svn/trunk/markerclusterer/images/m'
```
md5: @hooverlunch Let me know if you'd like me to try to vendor these marker images.

app/assets/javascripts/backbone/views/dashboard_map_view.js.coffee
```diff
@@ -0,0 +1,117 @@
...
+      split = loc.split(' ')
+      lat = parseFloat(split[0])
+      lng = parseFloat(split[1])
+
+      # create marker
+      p = new google.maps.LatLng(lat, lng)
+      m = new google.maps.Marker({
+        position: p,
+        title: I18n.t('activerecord.models.response.one') + ' #' + response_id,
+        icon: 'https://storage.googleapis.com/support-kms-prod/SNP_2752125_en_v0',
```
md5: Same here. We may want to vendor this icon.

app/views/welcome/index.html.erb
```diff
@@ -62,6 +62,11 @@
     <%# no need to cache this bit either because there's nothing here (done by JS) %>
     <h2 class="interior"><%= t('dashboard.response_locations') %></h2>
     <div class="response_locations"></div>
+    <% if @location_answers_count > @location_answers.size %>
```
md5: Hopefully the styling for this is OK. I've got a 13" screen on my MBP, so it's difficult for me to test the styling on the dashboard due to its fancy attempts to recalculate element sizes.

md5:
BTW, @hooverlunch I did figure out how to control the text used for the `MarkerClusterer` markers. While looking through the code yesterday, I noticed you could call `setCalculator` to give it a custom "calculator" function. This allows you to control the index of the marker cluster image (they come in sizes 1-5) and the text that appears. I don't think that `setCalculator` is documented, but there are a few StackOverflow and other mentions of people using it.

app/assets/javascripts/backbone/views/dashboard_map_view.js.coffee
hooverlunch: Yes I think yes to both.

app/assets/javascripts/backbone/views/dashboard_map_view.js.coffee
```diff
@@ -0,0 +1,117 @@
+# ELMO.Views.DashboardMapView
+#
+# View model for the dashboard map
+class ELMO.Views.DashboardMapView extends Backbone.View
```
hooverlunch: Yay Backbone!

md5:
This needs some more work. Will reopen this PR or another when it's ready.

app/models/answer.rb
```diff
@@ -34,17 +34,22 @@ class Answer < ActiveRecord::Base
...
+    end
+
+    # return an AR relation
+    where.not(:value => nil)
+      .joins(:questioning => :question)
+      .where(:questions => { :qtype_name => 'location' })
+      .joins(:response)
+      .where(:responses => response_conditions)
+      .order('responses.updated_at DESC')
+      .paginate(:page => 1, :per_page => 1000)
```
hooverlunch: Very nice.

hooverlunch:
This looks fantastic. So clean and nice, a real addition to the codebase.

I do wonder about MarkerClusterer though. I think we'd be better off without it. GMaps should be able to handle rendering the 1000 small dots directly. I think it's more striking visually that way. I just tried it on my own machine and I like it better without. So lets axe it, unless you have a reason for keeping it.

Then please go ahead and merge.

Thanks!

md5:
I agree about `MarkerClusterer`. Besides the fact that 1,000 small dots actually works pretty well, the `MarkerClusterer` itself had some wonkiness. I was seeing cases where it wouldn't zoom correctly on click, as well as cases where it zoomed out instead of in on click. There's also a weird interaction between the bounds setting code and the clusterer that causes the centers of some clusters to end up right on the edge of the view, effectively clipping the cluster marker in half.

I'll remove the `MarkerClusterer` code and merge.

PR #80, opened 2015-06-18
2469: Make backtraces show up for feature spec failures

md5:
By setting `config.action_dispatch.show_exceptions` to `false`, Capybara is able to recover the exception that caused a 500 error and show it in the console on test failure.

Turning off the `show_exceptions` flag turns off the default translation of a few exceptions like `ActiveRecord::RecordNotFound` into their corresponding HTTP status codes (e.g. 404 Not Found in the case of `ActiveRecord::RecordNotFound`). I've updated one spec to catch `ActiveRecord::RecordNotFound` instead of expecting a 404, but another approach later could be add an explicit `rescue_from` for `ActiveRecord::RecordNotFound`.

### DIFF
--- a/app/controllers/sms_controller.rb
+++ b/app/controllers/sms_controller.rb
@@ -1,5 +1,9 @@
 # handles incoming sms messages from various providers
 class SmsController < ApplicationController
+  rescue_from Sms::UnverifiedTokenError do |exception|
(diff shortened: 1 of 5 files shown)


hooverlunch:
Looks good to me. Thanks.

PR #89, opened 2015-07-09
2343, 2472: Implement Excel spreadsheet import for users

md5:


app/helpers/elmo_form_helper.rb
```diff
@@ -4,8 +4,21 @@ module ElmoFormHelper
   # renders a form using the ElmoFormBuilder
   def elmo_form_for(obj, *args, &block)
     options = args.extract_options!
-    form_for(obj, *(args << options.merge(:builder => ElmoFormBuilder,
-      :html => {:class => "#{obj.class.model_name.singular}_form elmo_form #{options[:style]}"})), &block)
+
+    args << options.deep_merge(
```
hooverlunch: This is a bit cryptic. Perhaps a comment here. Also perhaps construct the param to deep_merge on the line before so that it is not so heinous. :)

app/helpers/elmo_form_helper.rb
```diff
@@ -4,8 +4,21 @@ module ElmoFormHelper
...
+      builder: ElmoFormBuilder,
+      html: {
+        class: "#{obj.class.model_name.singular}_form elmo_form"
+      }
+    ) do |key,oldval,newval|
+      if key == :class
+        "#{oldval} #{newval}"
+      else
+        newval
+      end
```
hooverlunch: Ternary?

app/models/operation_status.rb
```diff
@@ -0,0 +1,19 @@
...
+  attr_reader :total, :started, :failed, :completed, :in_progress
+
+  def initialize(operations)
+    sql = operations.unscope(:order).select(
+      'count(1) as total',
+      'count(operations.job_started_at) as started',
+      'count(operations.job_failed_at) as failed',
+      'count(operations.job_completed_at) as completed').to_sql
+
+    @total, @started, @failed, @completed = operations.connection.execute(sql).first
```
hooverlunch: Very slick!

app/models/user_batch.rb
```diff
@@ -23,59 +28,88 @@ def create_users(mission)
...
+        users << user
+
+      end # iteration over rows
 
       # now if there was a validation error with any user, rollback the transaction
       raise ActiveRecord::Rollback if @validation_error
 
     end # transaction
 
+    # TODO: remove the uploaded file
```
hooverlunch: Did this get done?

app/views/operations/index.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="screen">
+  <%= index_table table_css_classes: 'index_table-narrow' %>
```
hooverlunch: Can we do `index-table-narrow` instead? Trying to keep these dash style.

hooverlunch:
Stunning. A work of pure beauty. Thank you.

app/models/user_batch.rb
md5: Yep. It was done in `UserImportOperationJob`. I've removed the `TODO` comment.

app/views/operations/index.html.erb
md5: Fixed.

app/helpers/elmo_form_helper.rb
md5: I've updated this as you suggested. Hopefully it's more maintainable now.

PR #91, opened 2015-07-14
2337: Allow options to have geographic coordinates

md5:
Adds the ability to mark a geographic `OptionSet` with an `allow_coordinates` flag. Setting this flag on an option set exposes the ability to edit the `latitude` and `longitude` fields of an associated `Option`.

For answers that have coordinates, a small map is now shown in the `Response` "show" view.

This PR contains a first pass at implementing coordinate validation in the option editing modal dialog. Since the existing implementation did not show any error messages, I have not added that ability here either. All I do is disable the "Save" button as was done previously.

### DIFF
--- a/app/assets/javascripts/backbone/views/answer_map_view.js.coffee
+++ b/app/assets/javascripts/backbone/views/answer_map_view.js.coffee
@@ -1,0 +1,28 @@
+class ELMO.Views.AnswerMapView extends Backbone.View
+
+  initialize: (params) ->
+    # create the map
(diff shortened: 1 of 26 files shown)


md5:
I'll add that one thing I'm not totally happy with is the changes to add coordinate support to `ELMO.Views.DraggableList`. I feel that this should be done in a subclass somehow, but getting it to work that way seemed like it would be a few hours of extra work for largely aesthetic reasons.

PR #104, opened 2015-08-03
2768: Fix broken migration due to MyISAM.

lmatiolis:
It was trying to add a foreing key related to a MyISAM table (it doesn't support it). Just reverting it back to
InnoDB fixes the problem.


lmatiolis:
@hooverlunch The problem with the `questionable_id` appears if we run the migrations from the start with a clean db. The `schema.rb` generated from that is different from the one we have right now... The quick solution for now was just to use the `schema:load` and run the new migrations after that.

PR #105, opened 2015-08-03
2653: Add NOT NULL constraint on option_id and answer_id in choices.

lmatiolis:


app/models/choice.rb
```diff
@@ -2,6 +2,8 @@ class Choice < ActiveRecord::Base
   belongs_to(:answer, :inverse_of => :choices, :touch => true)
   belongs_to(:option, :inverse_of => :choices)
```
lmatiolis: Any reason to use the old rocket syntax? Should i change the new line i added to use it also?

app/models/choice.rb
```diff
@@ -2,6 +2,8 @@ class Choice < ActiveRecord::Base
...
 
+  validates :answer, :option, presence: true
```
lmatiolis: I noticed we aren't using shoulda-matchers to check this kind of thing on specs. And i didn't see a model spec for `Choice`... Should i just leave it like this?

app/models/choice.rb
hooverlunch: We are slowly moving toward newer syntax. As a general rule, new and refactored code should use new syntax, old stuff should be left alone.

app/models/choice.rb
hooverlunch: I am not a believer in testing the builtin functionality of Rails. Thanks for asking though.

I also don't think we need an app-level validation here if we have null false in the DB. Do you?

lmatiolis: Well, i like having both validations, because i like to easily refer to what is needed on the code and also to have clean messages on the model.errors (using validate and that kind of stuff). But as you said, it's not really needed! ;)
And the specs for that is just to make sure someone doesn't accidentally remove something it was meant to be there... again, this is totally questionable.

But this is just me. Please let me know what do you prefer and i'll rebase the changes here. 

Thanks!

app/models/choice.rb
lmatiolis: Got it! Thanks!

app/models/choice.rb
hooverlunch: I see. Well just for consistency let's go without the app-level validation. Thanks.

lmatiolis: Sure @hooverlunch! No problem.

PR #106, opened 2015-08-04
2684: Fix option suggest breaking with parenthesis.

lmatiolis:
The problem was related to not escaping the query string.

### DIFF
--- a/spec/models/option_suggester_spec.rb
+++ b/spec/models/option_suggester_spec.rb
@@ -4,7 +4,7 @@ describe OptionSuggester do
 
   context 'general case' do
     before do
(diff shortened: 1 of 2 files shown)


spec/models/option_suggester_spec.rb
```diff
@@ -22,6 +22,11 @@
       result = OptionSuggester.new.suggest(get_mission, 'oo')
       expect(result.map(&:name)).to eq ['oo']
     end
+
+    it 'works with a parenthesis on the query string' do
+      result = OptionSuggester.new.suggest(get_mission, '(with')
+      expect(result.map(&:name)).to eq ['(with parenthesis)', '(with']
+    end
```
hooverlunch: Lovely.

PR #108, opened 2015-08-04
2714: Fix problem with empty QingGroup on _odk_group.

lmatiolis:
The bug was happening because the partial `_odk_group` was trying to iterate over an empty `QingGroup`.


app/views/forms/_odk_group.xml.erb
```diff
@@ -1,21 +1,24 @@
-<% children = subtree.keys %>
-<% grid_mode = grid_mode?(children) %>
+<% unless empty_qing_group?(subtree) %>
```
hooverlunch: Why not just do `unless children.empty?`

PR #113, opened 2015-08-05
2275 fix whitelist users on protected form

lmatiolis:
The problem reported was on form creation, where it wasn't saving the users on protected forms. But there was also a bug on edit. It wasn't removing whitelist_users because of a call on destroy instead
of destroy_all.

Also added dependent: :destroy on whitelist_users to avoid database orphans when we deleted a form.


hooverlunch:
Go for it.

PR #114, opened 2015-08-06
2280: Fix html showing on search results.

lmatiolis:
It was just missing a configuration on the thinking_sphinx.yml.
For a while, Thinking Sphinx auto-escaped excerpts.However, Sphinx
itself can remove HTML entities for indexing and excerpts, which
is a better way to approach this.


hooverlunch:
Good!

PR #116, opened 2015-08-06
2371: Fix styles on option sets modals.

lmatiolis:
Just removed some custom css that was making it look strange and added styles to show an inline label -> input on the levels and option edit with latitude and longitude.

### DIFF
--- a/app/assets/stylesheets/all/elmo_forms.css.scss
+++ b/app/assets/stylesheets/all/elmo_forms.css.scss
@@ -233,20 +233,10 @@ div.cascading-selects {
       border: none;
     }
 
-    div.label_and_control {
-      width: 100%;
-    }
-
-    label.main {
-      width: 30%;
-    }
-
-    div.control {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/all/elmo_forms.css.scss
```diff
@@ -233,20 +233,10 @@ div.cascading-selects {
       border: none;
     }
 
-    div.label_and_control {
```
hooverlunch: Eeep! Changing elmo_forms styles globally makes me nervous. But if you've tested extensively and you're sure it works, then I'm ok with it. And also tested all the various modals?

lmatiolis: I did check the modals when i implemented, but just to make sure i decided to check them again. I searched for modal classes on views of the project and checked them on the app and they are all being displayed fine. Will merge. Thanks!

PR #118, opened 2015-08-06
2363 and 2784: Fix admin edit with empty roles

lmatiolis:


lmatiolis:
Do not merge this one yet... the validation bypass generated some other bugs on the system because it is expecting a user on a mission to have a role.

app/models/user.rb
```diff
@@ -344,6 +345,11 @@ def must_have_assignments_if_not_admin
       end
     end
 
+    def clear_assignments_without_roles
+      empty_assignments = self.assignments.select{ |a| a if a.role.blank? }
```
hooverlunch: The `.select` block should return a boolean. So you can remove `a if `.

app/models/user.rb
```diff
@@ -344,6 +345,11 @@ def must_have_assignments_if_not_admin
...
+      self.assignments.delete empty_assignments
```
hooverlunch: You could refactor this as `assignments.delete(assignments.select(&:no_role?))`.

hooverlunch:
This looks good. I think we should add a spec or two to requests/authorization_spec.rb. I know it's an odd filename but it's where a lot of other similar specs currently live, e.g. "observer cant update own role". One day they will be refactored out, but fine for now I think.

PR #120, opened 2015-08-07
2783: Fix <br> tag appearing on sms index.

lmatiolis:
The problem is that the string wasn't being tagged as `html_safe`. To avoid these kind of mistakes of just forgetting to call it, we can change the I18n key to end with "_html" and that automatically sets it as `html_safe`.

### DIFF
--- a/app/helpers/sms_helper.rb
+++ b/app/helpers/sms_helper.rb
@@ -11,14 +11,14 @@ module SmsHelper
     when "time" then
       if sms.sent_at <= sms.created_at - 1.minute
         time_diff = time_diff(sms.sent_at, sms.created_at)
-        t('sms.timestamp_with_diff', time: l(sms.created_at), time_diff: time_diff)
+        t('sms.timestamp_with_diff_html', time: l(sms.created_at), time_diff: time_diff)
       else
         l(sms.created_at)
       end
     when "to" then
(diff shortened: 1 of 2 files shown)


app/helpers/sms_helper.rb
```diff
@@ -11,14 +11,14 @@ def format_sms_messages_field(sms, field)
...
-        t('sms.timestamp_with_diff', time: l(sms.created_at), time_diff: time_diff)
+        t('sms.timestamp_with_diff_html', time: l(sms.created_at), time_diff: time_diff)
       else
         l(sms.created_at)
       end
     when "to" then
       recips = safe_join(sms.recipient_hashes(max: MAX_RECIPS_TO_SHOW).map { |r| user_with_phone(r[:user], r[:phone]) }, '<br/>'.html_safe)
       extra_recipients = sms.recipient_count - MAX_RECIPS_TO_SHOW
-      recips << (extra_recipients > 0 ? t('sms.extra_recipients', count: extra_recipients).html_safe : '')
+      recips << (extra_recipients > 0 ? t('sms.extra_recipients_html', count: extra_recipients) : '')
```
hooverlunch: Nicely done. I was not aware of this trick!

hooverlunch:
Please merge!

PR #121, opened 2015-08-07
2503: Add Search::Qualifier support for multi columns.

lmatiolis:
Fixed bug of number:123 not searching sender of SMS. It turns out it was searching only on the .to column and not on the .from table also. 

Had to add support to generate a sql from multiple columns.

### DIFF
--- a/app/models/search/token.rb
+++ b/app/models/search/token.rb
@@ -84,8 +84,12 @@ class Search::Token
             end
           end
 
-          # insert the comparison itself
-          sql += comparison_fragment(qual, op, lex_tok)
+          if qual.has_more_than_one_column?
+            sql += comparisons_for_all_columns(qual, op, lex_tok)
+          else
+            sql += comparison_fragment(qual, qual.col, op, lex_tok)
+          end
+
           expr.values += lex_tok.is?(:string) ? "\"#{lex_tok.content}\"" : lex_tok.content
 
         # else, if this is an 'OR', insert that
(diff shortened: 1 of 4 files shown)


app/models/search/token.rb
```diff
@@ -106,36 +110,48 @@ def comparison(qual, op, rhs_or_values)
       end
     end
 
-    # generates an sql where clause fragment for a comparison with the given qualifier, operator, and value token
-    def comparison_fragment(qual, op, value_token)
+    def qualifier_has_more_than_one_column(qualifier)
```
hooverlunch: This should be a method on the Qualifier class. I have noticed this in your code several times. Think twice when writing a method on a class that does a small operation on a property of another class. Can it be made into a computed property of that second class?

app/models/search/token.rb
```diff
@@ -106,36 +110,48 @@ def comparison(qual, op, rhs_or_values)
...
+      qualifier.col.is_a? Array
+    end
+
+    def comparisons_for_all_columns(qualifier, op, lex_tok)
+      sql_conditions = qualifier.col.map do |column|
+                         comparison_fragment(qualifier, column, op, lex_tok)
+                       end
+
+      sql_conditions.join(' OR ')
```
hooverlunch: If you use a one letter loop var I think this could be a one-liner without loss of readability: 
`qualifier.col.map{ |c| comparison_fragment(qualifier, c, op, lex_tok) }.join(' OR ')`.

Please prefer code like this when available.

spec/models/search/search_spec.rb
```diff
@@ -285,6 +289,11 @@
     assert_search(str: 'test', sql: "((t1.f2 = 'test') OR (t1.f3 = 'test'))", qualifiers: MULTIPLE_DEFAULTS)
   end
 
+  it "supports multiple columns to generate a sql" do
+    assert_search(str: 'number:987', sql: "((msg.to LIKE '%987%') OR (msg.from LIKE '%987%'))", qualifiers: INDEXED)
+    assert_search(str: 'date:date', sql: "((msg.created_at = 'date') OR (msg.updated_at = 'date'))", qualifiers: INDEXED)
+  end
+
   private
```
hooverlunch: Nice specs!

hooverlunch:
Looks great. A few changes and then please merge.

app/models/search/token.rb
lmatiolis: I did write it at first as a one-liner, but then decided to change because of readability. :D But i guess you are right, using a one letter loop makes this fine. Thanks!

app/models/search/token.rb
lmatiolis: Ooops, you are absolutely right! Thanks!

PR #122, opened 2015-08-10
2676: Fix required asterisk display on SMS guide.

lmatiolis:
It was just missing a `display: inline-block` on it's css. Also refactored class name to use dashes instead of underscore.


hooverlunch:
Good to merge!

PR #123, opened 2015-08-10
2716: Fix questions numbering on sms guide.

lmatiolis:
For questions that were nested on a group, the ranks were being shown without the group index. Just created a helper to calculate and format that.

### DIFF
--- a/app/views/forms/sms_guide.html.erb
+++ b/app/views/forms/sms_guide.html.erb
@@ -42,7 +42,7 @@
     <%# loop over each question %>
     <% @form.smsable_questionings.each do |qing| %>
       <tr>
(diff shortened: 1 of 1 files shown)


app/helpers/forms_helper.rb
```diff
@@ -122,6 +122,17 @@ def sms_example_for_question(qing)
     end
   end
 
+  def rank_for_sms_question(qing)
```
hooverlunch: Would the `full_rank` method not work here?

This approach seems kind of unorthodox...

lmatiolis: Wow! Didn't know about `full_rank`. Searching here, it even has a `full_dotted_rank` method, so it can be used and no need for all this code. Thanks!

PR #124, opened 2015-08-12
2283 fix database schema

lmatiolis:
Had to fix previous migration (remove FK before adding it) because the staging database already have it. 
Also added gem to always keep schema ordered.


hooverlunch:
Good to go. Let's keep an eye on this.

PR #125, opened 2015-08-12
2680 fix option set import/export i18n

lmatiolis:
Fix option set import preferred_locale and also export.
Import wasn't using the ones from configatron and export was just using the name for the option, not one
of it's translations. Now we are getting the translation for the first language preferred.

### DIFF
--- a/app/models/option_node.rb
+++ b/app/models/option_node.rb
@@ -222,7 +222,7 @@ class OptionNode < ActiveRecord::Base
 
   # arranges option descendant nodes into rows for export.
   # rows are created for leaf nodes only and contain the node id and the
-  # localized option names in the current locale.
+  # localized option names in the mission preferred locale.
   def arrange_as_rows(hash = nil, parent_path = [])
     hash = arrange_with_options(eager_load_option_assocs: false) if hash.nil?
 
@@ -232,7 +232,7 @@ class OptionNode < ActiveRecord::Base
       # output a row if we've hit a leaf node or a parent node with coordinates
       if children.empty? || (option_set.allow_coordinates? && node.option.has_coordinates?)
         # use the option path to construct the list of cell values
-        row = [node.id, *path.map(&:name)]
+        row = [node.id, *preferred_name_translations(path)]
 
         # add the coordinates if the option set allows coordinates
         row << node.option.coordinates if option_set.allow_coordinates?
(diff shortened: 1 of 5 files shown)


app/models/option_node.rb
```diff
@@ -232,7 +232,7 @@ def arrange_as_rows(hash = nil, parent_path = [])
       # output a row if we've hit a leaf node or a parent node with coordinates
       if children.empty? || (option_set.allow_coordinates? && node.option.has_coordinates?)
         # use the option path to construct the list of cell values
-        row = [node.id, *path.map(&:name)]
+        row = [node.id, *preferred_name_translations(path)]
```
hooverlunch: Need to also fix the comment on line 225 to say 'mission preferred locale'.

app/models/option_node.rb
```diff
@@ -245,6 +245,12 @@ def arrange_as_rows(hash = nil, parent_path = [])
     end
   end
 
+  def preferred_name_translations(path)
+    path.map do |path|
+      path.name_translations[configatron.preferred_locales.first.to_s]
```
hooverlunch: We should really add a `mission_preferred_locale` entry to configatron. Would be much cleaner and reusable.

app/models/option_node.rb
```diff
@@ -245,6 +245,12 @@ def arrange_as_rows(hash = nil, parent_path = [])
     end
   end
 
+  def preferred_name_translations(path)
+    path.map do |path|
```
hooverlunch: Using `path` as loop var is confusing. For short loops I usually use the initials of the obj class. So in this case you could just use `o`.

app/models/option_set_import.rb
```diff
@@ -199,4 +202,9 @@ def blank_interior_cell?(row)
       !row[row.index(nil)..-1].all?(&:nil?)
     end
 
+    def name_locale_key
+      preferred_locale = configatron.preferred_locales.first
+      name_locale_key = "name_#{preferred_locale}".to_sym
```
hooverlunch: If we use the new configatron key you can combine these two lines and remove the assignment.

hooverlunch:
Just a few comments and looks good to merge.

lmatiolis:
@hooverlunch Please check if i did set the preferred_locale in the right way on the configatron.

app/models/option_set_import.rb
```diff
@@ -199,4 +202,7 @@ def blank_interior_cell?(row)
       !row[row.index(nil)..-1].all?(&:nil?)
     end
 
+    def name_locale_key
+      "name_#{configatron.preferred_locale}".to_sym
```
hooverlunch: Yes!

lmatiolis: Bonus points earned! :)

PR #126, opened 2015-08-12
2819: Fix tags not being rendered properly on form list.

lmatiolis:
It wasn't being marked as html safe. We need to guarantee that the tag.name is being properly escaped
before doing this...

### DIFF
--- a/app/helpers/questions_helper.rb
+++ b/app/helpers/questions_helper.rb
@@ -44,7 +44,7 @@ module QuestionsHelper
     when "actions" then table_action_links(q)
     when "name"
       if params[:controller] == 'forms'
-        q.name + render_tags(q.sorted_tags)
+        html_escape(q.name) << render_tags(q.sorted_tags)
       else
         link_to(q.name, q) + render_tags(q.sorted_tags, clickable: true)
       end


hooverlunch:
Thanks but I would prefer you debug the `render_tags` method to see why it is not producing html_safe output. Calling `html_safe` on the result is not... safe..

app/helpers/questions_helper.rb
```diff
@@ -44,7 +44,7 @@ def format_questions_field(q, field)
     when "actions" then table_action_links(q)
     when "name"
       if params[:controller] == 'forms'
-        q.name + render_tags(q.sorted_tags)
+        (q.name + render_tags(q.sorted_tags)).html_safe
```
hooverlunch: I would rather not do this as it's not super safe.

Can't you just do `h(q.name) << render_tags(q.sorted_tags)`?

`render_tags` should be returning html_safe output. q.name is not html_safe, so it is causing the problem when concatenated. But if you do as above, both operands should be safe, and we should be good.

hooverlunch:
Thanks for updating. Good to merge.

PR #128, opened 2015-08-13
2821: Call to extract was missing param.

lmatiolis:
It wasn't setting the form_ids and, as a result, always showing all questions (Using ALL for the form_ids).

### DIFF
--- a/app/assets/javascripts/views/report/edit_pane.js
+++ b/app/assets/javascripts/views/report/edit_pane.js
@@ -20,7 +20,7 @@
   }
 
   klass.prototype.broadcast_change = function(attrib) {
-    this.extract();
+    this.extract(true);
     this.parent_view.broadcast_change(attrib);
   }
 


app/assets/javascripts/views/report/edit_pane.js
```diff
@@ -20,7 +20,7 @@
   }
 
   klass.prototype.broadcast_change = function(attrib) {
-    this.extract();
+    this.extract(attrib);
```
hooverlunch: As you have probably realized, the extract method is defined on subclasses of EditPane. In some cases, it expects a boolean parameter named `enabled`, which indicates whether the pane is enabled. Admittedly this is gross. But it looks like you're instead passing an attribute name. I think it's working because  `attrib` is evaluating to true. I think instead you can just pass `true`, since we can assume that if the pane is broadcasting a change, it is enabled.

PR #129, opened 2015-08-13
2844: Remove filename for security reasons.

lmatiolis:


hooverlunch:
Looks good. Was manually tested? (To test, just do an option set import or two..)

PR #131, opened 2015-08-15
CSS for right-to-left support

mapmeld:
most elements in the UI look OK in a right-to-left language, but a few appear in the wrong order. Here's a SCSS file which flips the remaining lists and padding inside \<html dir="rtl"\>


hooverlunch:
Thanks for this!

PR #132, opened 2015-08-17
2839 fix inconsistent qing count

lmatiolis:
The bug is related to duplicating a form more than once. After the first copy, each time you duplicate the original form again, it adds questionings copies to the first copy (on the form_items table) and not on the new copy.

Also added a nullify on original_id fk constraint, because user couldn't delete a form that was copied.

### DIFF
--- a/spec/models/replication/form_replication_spec.rb
+++ b/spec/models/replication/form_replication_spec.rb
@@ -46,6 +46,17 @@ describe Form do
         expect(@copy1.c[0].question).to eq @copy2.c[0].question
         expect(@copy1.c[0].question.option_set).to eq @copy2.c[0].question.option_set
       end
+
+      context 'when using eager loaded values from form items query' do
+        it 'keeps the questioning count consistent' do
+          std_qing_count = form_items_qing_count(@std)
+          copy1_qing_count = form_items_qing_count(@copy1)
+          copy2_qing_count = form_items_qing_count(@copy2)
+
+          expect(std_qing_count).to eq copy1_qing_count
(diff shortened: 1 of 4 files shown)


db/migrate/20150817204307_add_nullify_to_form_original_id_fk.rb
```diff
@@ -0,0 +1,6 @@
+class AddNullifyToFormOriginalIdFk < ActiveRecord::Migration
+  def change
+    remove_foreign_key "forms", column: "original_id", name: "forms_standard_id_fk"
+    add_foreign_key "forms", "forms", column: "original_id", name: "forms_standard_id_fk", on_delete: :nullify
```
hooverlunch: Please add the nullify for all original_id FKs. I believe there are three in all.

spec/models/replication/form_replication_spec.rb
```diff
@@ -46,6 +46,17 @@
...
       end
+
+      context 'when using eager loaded values from form items query' do
+        it 'keeps the questioning count consistent' do
+          std_qing_count = form_items_qing_count(@std)
+          copy1_qing_count = form_items_qing_count(@copy1)
+          copy2_qing_count = form_items_qing_count(@copy2)
+
+          expect(std_qing_count).to eq copy1_qing_count
+          expect(std_qing_count).to eq copy2_qing_count
```
hooverlunch: Nice spec! I'm surprised we were alrady calling clone twice in this spec but didn't catch this!

hooverlunch:
Good to merge.

PR #133, opened 2015-08-17
2846: Restrict option set multilevel creation on modal.

lmatiolis:
Now we are disabling the option to be multilevel if it's creating an option set for a select multiple question.

### DIFF
--- a/app/assets/javascripts/views/question_form.js
+++ b/app/assets/javascripts/views/question_form.js
@@ -49,8 +49,8 @@
     var selected_type = self.field_value('qtype_name');
 
     // Show/hide option set field and hint
-    options = options || {}
-    options.multilevel = selected_type == 'select_one'
+    options = options || {};
+    options.multilevel = selected_type == 'select_one';
     self.show_option_set_select(selected_type == 'select_one' || selected_type == 'select_multiple', options);
 
     // show/hide max/min
@@ -89,8 +89,11 @@
     // show the loading indicator
     $('div.question_fields #option_set_id .loading_indicator').show();
 
+    var question_type_param = '?adding_to_question_type=' + self.field_value('qtype_name'),
+        loadUrl = self.params.new_option_set_path + question_type_param;
+
     // populate and show the modal
-    $("#create-option-set .modal-body.option-set").load(self.params.new_option_set_path, function(){
+    $("#create-option-set .modal-body.option-set").load(loadUrl, function(){
       $("#create-option-set").modal('show');
     });
   }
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/views/question_form.js
```diff
@@ -91,6 +91,9 @@
 
     // populate and show the modal
     $("#create-option-set .modal-body.option-set").load(self.params.new_option_set_path, function(){
+      var multilevel_checkbox = $("#option_set_multi_level");
+      self._can_be_multi_level() ? multilevel_checkbox.removeAttr('disabled') : multilevel_checkbox.attr('disabled', 'disabled')
```
hooverlunch: I see what you're doing here but I think that enabling/disabling the checkbox should be the responsibility of the OptionSetForm JS class. The question form really shouldn't know anything about that box, or whether an option set can be multilevel. Principle of least responsibility.

Instead I would pass a param to the OptionSetForm class indicating the question type for which the option set is getting created. Then let it decide how to handle itself.

You can do this by passing a GET param along with the .load call, which you can then pass to the OptionSetForm object on instantiation. (in option_sets/_form.html.erb).

Make sense?

lmatiolis:
@hooverlunch Please check if this is what you meant. I did search for something to be set on the @option_set instance, but didn't find anything suitable to restrict the multi_level... is there something we can use? Right now, i'm just using the param directly on the partial... Thanks!

app/views/option_sets/_form.html.erb
```diff
@@ -17,7 +17,7 @@
     <%= f.field(:geographic, :type => :check_box, :read_only => cannot?(:change_geographic, @option_set)) %>
     <%= f.field(:allow_coordinates, :type => :check_box, :read_only => cannot?(:change_geographic, @option_set)) %>
     <%= f.field(:multi_level, :type => :check_box,
-      :read_only => cannot?(:update_core, @option_set) || @option_set.has_select_multiple_questions? || @option_set.huge?) %>
+      :read_only => cannot?(:update_core, @option_set) || @option_set.has_select_multiple_questions? || @option_set.huge? || params[:question_type] == "select_multiple") %>
```
hooverlunch: Yikes this is getting a bit hairy. How about this:
1. Add an attr_accessor called `adding_to_question_type` to OptionSet with a comment of 'Indicates that this OptionSet is being created to be added to a question of the given type'.
1. Change the GET param to match this.
1. In OptSetController, do `OptionSet.new(adding_to_question_type: params[:adding_to_question_type])` 
1. Add a new method to OptionSet called `can_be_multilevel?` that checks has_select_multiple_questions, huge, and adding_to_question_type.
1. Call that in the template.

lmatiolis: Great, Tom! 
The same concern crossed my mind when i saw all those checks on the template... Doing what you said will improve this a lot!

Will do these changes here... Thanks!

PR #134, opened 2015-08-18
2866: Fix sms broadcast not working with both phones.

lmatiolis:
Twilio doesn't support multiple recipients on the same call, so you need to iterate over each recipient to send a broadcast for more than one phone.


hooverlunch:
Looks good. I worry about the scalability of this, but let's cross that bridge later.

PR #135, opened 2015-08-21
Fix flapping specs

lmatiolis:


lmatiolis:
@hooverlunch Please, review this. I've did some rebuilds on Travis and everything seems to be working! :) I think we can merge this one and keep an eye for the flapping specs and if some appear again (or new ones), we deal with them.

PR #138, opened 2015-08-28
Add file extension on file upload.

lmatiolis:
This was causing errors on user import, for example.


app/controllers/upload_processable.rb
```diff
@@ -4,7 +4,8 @@ module UploadProcessable
   protected
 
     def store_uploaded_file(uploaded)
-      file_name = "#{controller_name}-#{SecureRandom.uuid}"
+      file_extension = uploaded.original_filename.split('.').last
```
hooverlunch: Perhaps use `File.extname`?

PR #140, opened 2015-09-03
3005 optimize user creation

lmatiolis:
Doing checks in batches on users and on the spreadsheet. We are doing this to avoid a lot of calls to the database for each user. Also doing one query to insert in batches.
This required some bypassing on users validation (to not trigger some queries)


app/models/user.rb
```diff
@@ -30,19 +27,26 @@ class User < ActiveRecord::Base
     c.disable_perishable_token_maintenance = true
     c.perishable_token_valid_for = 1.week
     c.logged_in_timeout(SESSION_TIMEOUT)
+
     c.validates_format_of_login_field_options = {:with => /[\a-zA-Z0-9\.]+/}
-    c.merge_validates_length_of_password_field_options minimum: 8
+    c.merge_validates_uniqueness_of_login_field_options(:unless => Proc.new{|u| u.skip_validations_with_queries?})
+
+    c.merge_validates_length_of_password_field_options(minimum: 8, :unless => Proc.new{|u| u.skip_validations_with_queries?})
```
hooverlunch: These lines are too long I think. We usually split at 100 +/- 10. Thanks.

app/models/user.rb
```diff
@@ -30,19 +27,26 @@ class User < ActiveRecord::Base
...
 
     # email is not mandatory, but must be valid if given
     c.merge_validates_format_of_email_field_options(:allow_blank => true)
-    c.merge_validates_uniqueness_of_email_field_options(:unless => Proc.new{|u| u.email.blank?})
+    c.merge_validates_uniqueness_of_email_field_options(:unless => Proc.new{|u| u.skip_validations_with_queries? || u.email.blank?})
```
hooverlunch: These look great, I think this is a nice solution.

app/models/user.rb
```diff
@@ -51,8 +55,11 @@ class User < ActiveRecord::Base
   validate(:password_reset_cant_be_email_if_no_email)
   validate(:no_duplicate_assignments)
   validate(:must_have_assignments_if_not_admin)
-  validate(:phone_should_be_unique)
-  validates :password, format: { with: /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])/, if: :require_password?, message: :invalid_password }
+  validate(:phone_should_be_unique, unless: :skip_validations_with_queries?)
+  validates :password, format: { with: /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])/,
+                                 if: :require_password?,
+                                 unless: :skip_validations_with_queries?,
```
hooverlunch: `skip_validations_with_queries` doesn't seem quite accurate. password validation doesn't involve queries, it's just that we're leaving it out for other reasons. Perhaps we should just call this flag `batch_creation?`?

app/models/user.rb
```diff
@@ -299,6 +306,15 @@ def remember_last_mission(mission)
     save validate: false
   end
 
+  # OVERRIDE AUTHLOGIC METHOD.
+  # This is to avoid unnecessary queries to the database if we want to skip
+  # certain validations.
```
hooverlunch: Nice comment.

app/models/user.rb
```diff
@@ -299,6 +306,15 @@ def remember_last_mission(mission)
...
+  #
+  # Resets the persistence_token field to a random hex value.
+  def reset_persistence_token
+    super unless skip_validations_with_queries?
+  end
```
hooverlunch: Very nice.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
     end
 
     # parse the input file as a spreadsheet
-    data = Roo::Spreadsheet.open(file)
+    @data = Roo::Spreadsheet.open(file)
 
+    validate_headers
+
+    if !@validation_error
```
hooverlunch: unless?

app/models/user.rb
lmatiolis: You are right! It started being a flag for skipping validations with queries, but extended to be a global flag for all unnecessary validations on batch creation. Will change here. Thanks!

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+
+        row_start = @data.first_row
+        @import_num = last_import_num_on_users
+
+        user_batch_attributes = parse_rows(row_start)
+
+        (0..number_of_iterations).each do |i|
+          current_attributes_batch = user_batch_attributes[row_start-1, BATCH_SIZE]
+
+          users_batch = create_users_instances(current_attributes_batch, mission)
```
hooverlunch: I'm a little concerned about creating 1000 ruby objects at once. I had envisioned this as populating and validating a single user object over and over. But if this performs well then ok.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+          create_hash_table_with_fields_and_indexes(users_batch)
+
+          validate_users_batch(users_batch)
+
+          check_uniqueness_on_objects(users_batch, 'email')
+          check_uniqueness_on_objects(users_batch, 'login')
+          check_uniqueness_on_objects(users_batch, 'phone')
+          check_uniqueness_on_objects(users_batch, 'phone2')
```
hooverlunch: Should we not be checking phone and phone2 at once?

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+
+          check_uniqueness_on_db(users_batch, 'users', 'email', row_start)
+          check_uniqueness_on_db(users_batch, 'users', 'login', row_start)
+          check_uniqueness_on_db(users_batch, 'users', 'phone', row_start, ['phone','phone2'])
```
hooverlunch: Seems like this should be doable with one call to the method passing phone and phone2?

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+          check_validation_errors(users_batch, row_start)
+
+          break if errors_reached_limit
+
+          DirectDBConn.insert_on_db(users_batch, 'users')
+          insert_assignments_on_db(users_batch)
+
+          users.concat users_batch
+
+          p "Batch: #{i}"
```
hooverlunch: Do we need this?

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+
+          # Set row to start on the next batch
+          row_start = ((i+1) * BATCH_SIZE)+1
```
hooverlunch: Spaces around operators please.

app/models/user_batch.rb
lmatiolis: Yes, it's performing OK. The problem with setting on the same object is that some callbacks don't get executed (the login name suggestion, for example). I think the GC doesn't make this a problem...

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
-        if errors.count >= IMPORT_ERROR_CUTOFF
-          # we pass the previous row number to the error message formatting
-          # since the error count cutoff was surpassed by the previous rows,
-          # not this row
-          errors.add(:users, :too_many_errors, row: row_number - 1)
-          break
-        end
+  def number_of_iterations
+    users_rows_count = @data.count - 1
+    (users_rows_count/BATCH_SIZE.to_f).ceil - 1
```
hooverlunch: Spaces around ops here please. Also this could be one line without loss of readability IMO.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+  end
+
+  def parse_rows(offset)
+    user_batch_attributes = []
+
+    @data.each_row_streaming(offset: offset).with_index do |row,row_index|
```
hooverlunch: Space after comma please.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+      row_number = 1 + @data.first_row + row_index
+
+      # skip blank rows
+      next if row.all?(&:blank?)
+
+      # turn the row into an attribute hash
+      attributes = Hash[*row.map do |cell|
+        field = @fields[cell.coordinate.column - 1]
+        [field, cell.value.presence]
+      end.flatten]
```
hooverlunch: This should probably be its own helper method.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+
+    columns = ['email', 'login', 'phone', 'phone2']
+
+    columns.each{ |field| populate_hash_with_field_and_occurrences(field, objects) }
+  end
+
+  def populate_hash_with_field_and_occurrences(field, objects)
+    key = correct_field_key(field)
+
+    objects.map{|o| o.send(field)}.each_with_index do |value, index|
```
hooverlunch: Space inside braces please.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+      (@fields_hash_table[key][value] ||= []) << index unless value.nil?
+    end
+  end
+
+  def validate_users_batch(users_batch)
+    users_batch.each{|u| u.valid? }
```
hooverlunch: Balance spaces here.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+      if errors_reached_limit
+        add_too_many_errors(row_number)
+        break
+      end
+    end
+  end
+
+  def create_new_user(attributes, mission)
+    User.new(attributes.merge(
+             reset_password_method: "print",
```
hooverlunch: Just one extra indent please.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+    unless user.errors.empty?
+      user.errors.keys.each do |attribute|
+        if attribute != :persistence_token
+          user.errors.full_messages_for(attribute).each do |error|
+            add_error(error, attribute, row_number)
           end
-          @validation_error = true
         end
+      end
+      @validation_error = true unless error_is_on_persistence_token? user
```
hooverlunch: If you move this inside the `if` statement above you shouldn't need the extra method, right?

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
-    return succeeded?
+  def actual_row_number(row_start, index)
+    index + row_start + 1
+  end
+
+  def check_uniqueness_on_objects(objects, field)
+    key = correct_field_key(field)
+    # Add errors on fields that aren't unique
+    @fields_hash_table[key].each do |k,v|
+      v.each { |i| objects.at(i).errors.add(field, :taken) } if v.length != 1
```
hooverlunch: No space after `.each` please.

Also, why not just add the objects themselves to the hash table instead of their array index? Then you save the kind of awkward `.at(i)` thing, which was a bit hard to decipher.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+    end
+  end
+
+  def check_uniqueness_on_db(objects, table, field, row_start, columns=[])
+    results = DirectDBConn.check_uniqueness_on_db(objects, table, field, row_start, columns)
+
+    results.each do |result|
+      @fields_hash_table[field][result.first].each do |index|
+        p "UNIQ FAIL DB: #{field} : #{result.first}"
+        object = objects.at(index)
```
hooverlunch: Same thing here as above.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+        object.errors.add(field, :taken)
+      end
+    end
+  end
+
+  def correct_field_key(field)
```
hooverlunch: May not need this method if you refactor as I suggested above.

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+    field.include?('phone') ? 'phone' : field
+  end
+
+  def insert_assignments_on_db(objects)
```
hooverlunch: I think you can remove `_on_db`. `insert` is explanatory enough.

app/models/user_batch.rb
lmatiolis: Yes, but then it would be setting true for each iteration and that is a little strange, no? What do you prefer?

app/models/user_batch.rb
```diff
@@ -35,79 +36,247 @@ def create_users(mission)
...
+      #Add Select query on user_id position
+      user_id_index = column_names.split(', ').index('user_id')
+      values_as_array[user_id_index] = users_id[index]
+
+      "(#{values_as_array.join(',')})"
+    end
+
+    sql_query_array = ["INSERT INTO assignments (#{column_names}) VALUES #{object_values.join(', ')}"]
+
+    DirectDBConn.execute_sql_query_array(sql_query_array)
```
hooverlunch: I'm curious why the `INSERT ... SELECT` thing never worked here. Specifically `["INSERT INTO assignments(user_id, mission_id, role) SELECT id, ?, 'observer' FROM users WHERE import_num = ?", mission_id, import_num]`.

app/models/concerns/direct_db_conn.rb
```diff
@@ -0,0 +1,65 @@
+module DirectDBConn
```
hooverlunch: This is not really being used as a concern. You are calling class methods on it directly. You are also passing `table` to many of the methods when it really should be an instance variable I think.

I recommend you move this into the models root directory and call it `BatchInserter`. Make all the methods instance methods and pass `table` to `initialize`. Then you only need to pass `objects` to the instance methods. 

Also, many of the methods should be private.

app/models/concerns/direct_db_conn.rb
```diff
@@ -0,0 +1,65 @@
...
+      object_values = objects.map{|o| "(#{convert_values_to_insert_syntax(o)})"}
+      # sql_query_array = ["INSERT INTO #{table} (#{column_names}) VALUES #{string_params}", object_values].flatten
+      # TODO sanitize
+      sql_query_array = ["INSERT INTO #{table} (#{column_names}) VALUES #{object_values.join(', ')}"]
+
+      # Rescue because this uniqueness validation is already inserted on the model
+      execute_sql_query_array(sql_query_array) rescue ActiveRecord::RecordNotUnique
+    end
+
+    def check_uniqueness_on_db(objects, table, field, row_start, columns=[])
```
hooverlunch: See my later comment on how this could be refactored. Seems like you should only need to call this method once for phone and phone2.

app/models/concerns/direct_db_conn.rb
```diff
@@ -0,0 +1,65 @@
+module DirectDBConn
+  class << self
+    def insert_on_db(objects, table)
+      string_params = generate_string_params_template(objects.length)
+      column_names = objects.first.attributes.keys.join(', ')
+      object_values = objects.map{|o| "(#{convert_values_to_insert_syntax(o)})"}
+      # sql_query_array = ["INSERT INTO #{table} (#{column_names}) VALUES #{string_params}", object_values].flatten
+      # TODO sanitize
```
hooverlunch: Please take care of this before merging.

app/models/concerns/direct_db_conn.rb
```diff
@@ -0,0 +1,65 @@
+module DirectDBConn
+  class << self
+    def insert_on_db(objects, table)
```
hooverlunch: `insert` should be sufficient.

hooverlunch:
Great work here! I made some comments, generally suggestions for refactors to tighten up the code a little. But the fundamentals seem sound. As you said, the specs seem a little thin. Let's talk about this. But great job!

app/models/user_batch.rb
lmatiolis: I remember i had an issue doing a batch insert like that. For one query, the `INSERT ... SELECT` was working, but for more, no. I didn't want to loose more time on that, so i just switched to the other syntax that should be the same thing.

What I'm doing is the same thing, no? It's doing the `"(SELECT id FROM users WHERE import_num=#{o.import_num})"` all in one query, as it would be with `INSERT ... SELECT`. I think it's just a syntactic sugar for the same thing, no?

app/models/concerns/direct_db_conn.rb
lmatiolis: True, it's not really a concern. I just didn't want to put it into the models directory since i didn't want a module there. I can change here!

About the table on the instance, i was just thinking that this module could be used to more things (even on the same model, for different tables). But i guess then the person could just create more instances to use it... 

Good call on this! Will change here.

app/models/user_batch.rb
lmatiolis: I think the solution to just put the objects on the hash table instead of the indexes great! Will change here!

app/models/user_batch.rb
hooverlunch: Come to think of it the whole thing seems a little hacky. Is there no way to disable (or later remove) the error on the persistence token?

Failing that, you could do `error_attrs = user.errors.keys - IGNORE_ERROR_KEYS` at the top and then do `@validation_error = error_attrs.any?`.

app/models/user_batch.rb
hooverlunch: I don't think so. This is one query and one line of code that replaces the whole method above. I think it is also likely faster than running a huge chunk of inserts because it all happens internal to MySQL. I can help you debug if need be.

hooverlunch:
Looks great. Merge please.

PR #141, opened 2015-09-14
3074: Improve non observers show on StandardFormReport.

lmatiolis:
The system was fetching all users to show them on the form. This was
making the system extremely slow when it had a lot of users on the
database. I created one query to fetch the users limiting how many
we are getting and used SQL_CALC_FOUND_ROWS to get the total count
without hitting the db again.

### DIFF
--- a/app/models/report/standard_form_report.rb
+++ b/app/models/report/standard_form_report.rb
@@ -48,6 +48,9 @@ class Report::StandardFormReport < Report::Report
 
   TEXT_RESPONSE_OPTIONS = %w(all short_only none)
 
+  # How many non reporting observers it will show on the report before summarizing the rest
+  MISSING_OBSERVERS_SIZE_LIMIT = 100
+
   def as_json(options = {})
     # add the required methods to the methods option
     h = super(options)
@@ -88,17 +91,34 @@ class Report::StandardFormReport < Report::Report
     @summary_collection
   end
 
-  # returns all non-admin users in the form's mission with the given role that have not submitted any responses to the form
-  # options[:role] - (symbol) the role to check for
+  # Returns all non-admin users in the form's mission with the given role that have
+  # not submitted any responses to the form
+  #
+  # options[:role] the role to check for
+  # options[:limit] how many users we want to fetch from the db
   def users_without_responses(options)
-    # eager load responses with users
-    all_observers = form.mission.assignments.includes(:user).find_all{|a| a.role.to_sym == options[:role] && !a.user.admin?}.map(&:user)
-    submitters = form.responses.includes(:user).map(&:user).uniq
-    @users_without_responses = all_observers - submitters
+    User.without_responses_for_form(form, options)
   end
 
   def observers_without_responses
-    users_without_responses(:role => :observer)
+    users_response = users_without_responses({role: :observer, limit: MISSING_OBSERVERS_SIZE_LIMIT})
+    users = users_response[:users]
+
+    if users.empty?
+      I18n.t('report/report.zero_missing_observers')
+    else
+      format_observers_names(users, users_response[:count])
+    end
+  end
+
+  def format_observers_names(users, users_total_count)
+    users_list = users.map(&:name).join(', ')
+
+    total_count_label = I18n.t('report/report.missing_observers_total_count',
+      user_count: users_total_count)
+    users_list = "#{users_list}, ... (#{total_count_label})" if users_total_count > MISSING_OBSERVERS_SIZE_LIMIT
+
+    users_list
   end
 
   # returns the list of questionings to include in this report
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -155,6 +155,32 @@ class User < ActiveRecord::Base
       LIMIT ?", mission.id, mission.id, limit]).reverse
   end
 
+  # Returns all non-admin users in the form's mission with the given role that have
+  # not submitted any responses to the form
+  #
+  # options[:role] the role to check for
+  # options[:limit] how many users we want to fetch from the db
+  #
+  # Returns a hash with users instances and the count from the select without the LIMIT
+  def self.without_responses_for_form(form, options)
+    users = find_by_sql(["SELECT SQL_CALC_FOUND_ROWS * FROM users
+      INNER JOIN assignments ON assignments.user_id = users.id WHERE
+      assignments.mission_id = ? AND
+      assignments.role = ? AND
+      users.admin = FALSE AND
+      NOT EXISTS (
+        SELECT 1 FROM responses WHERE
+        responses.user_id=users.id AND
+        responses.form_id = ?
+      ) LIMIT ?;", form.mission.id, options[:role].to_s, form.id, options[:limit]])
+
+    # This returns the count of the previous users select without the limit clause.
+    # Format: [[count]]
+    users_count = ActiveRecord::Base.connection.execute('SELECT FOUND_ROWS();').entries[0].first
+
+    { users: users, count: users_count }
+  end
+
   # generates a cache key for the set of all users for the given mission.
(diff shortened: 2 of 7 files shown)


app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
...
-  def observers_without_responses
-    users_without_responses(:role => :observer)
+  # This returns the count of the previous users select without the limit clause.
+  # Format: [[count]]
+  def observers_without_responses_count
+    ActiveRecord::Base.connection.execute('SELECT FOUND_ROWS();').entries[0].first
+  end
+
+
+  def observers_without_responses(limit=100)
```
hooverlunch: I'd rather have this 100 as a constant on this model please.

spec/models/report/standard_form_report_spec.rb
```diff
@@ -77,7 +77,9 @@
 
     it "should return non-submitting observers" do
       # make observers
-      observers = %w(bob jojo cass sal).map{|n| create(:user, :login => n, :role_name => :observer)}
+      observers = %w(bob jojo cass sal).map{|n| create(:user, login: n,
+                                                              role_name: :observer,
+                                                              name: n.capitalize)}
```
hooverlunch: Please avoid this style of indenting. It gives me nightmares :P  Just please break the line and indent one extra level.

app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
   end
 
   # returns all non-admin users in the form's mission with the given role that have not submitted any responses to the form
-  # options[:role] - (symbol) the role to check for
+  # options[:role]) the role to check for
```
hooverlunch: Extra parenth on this line and the one below.

app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
...
-    @users_without_responses = all_observers - submitters
+    User.without_responses_for_form(form, options).sort
   end
 
-  def observers_without_responses
-    users_without_responses(:role => :observer)
+  # This returns the count of the previous users select without the limit clause.
+  # Format: [[count]]
+  def observers_without_responses_count
+    ActiveRecord::Base.connection.execute('SELECT FOUND_ROWS();').entries[0].first
```
hooverlunch: The found_rows thing is a neat trick but as currently is a bad idea I think b/c it's not even in the same class as where the query is run. I would suggest refactoring it so that the without_responses_for_form method calls both and returns both the count and the list of users as a 2 element array.

app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
...
+
+    if users.empty?
+      I18n.t('report/report.zero_missing_observers')
+    else
+      format_observers_names(users, limit)
+    end
+  end
+
+  def format_observers_names(users, limit)
+    users_list = users.map{|u| u.name }.join(', ')
```
hooverlunch: Please balance spaces here.

app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
...
+
+    users_total_count = observers_without_responses_count
+    total_count_label = I18n.t('report/report.missing_observers_total_count',
+                             {user_count: users_total_count})
```
hooverlunch: Just one level of indent please. Also you don't need the enclosing curly braces here.

app/models/report/standard_form_report.rb
```diff
@@ -89,16 +89,38 @@ def run(current_ability)
...
+    users_list = "#{users_list}, ... (#{total_count_label})" if users_total_count > limit
+
+    users_list
```
hooverlunch: This format method really belongs in a helper since it's strictly view logic, right? Please refactor.

app/models/user.rb
```diff
@@ -146,6 +146,29 @@ def self.sorted_observer_response_counts(mission, limit)
...
+                  users.admin = FALSE AND
+                  users.id NOT IN (
+                    SELECT distinct users.id FROM users
+                    INNER JOIN responses WHERE
+                    responses.user_id = users.id AND
+                    responses.form_id = ?
+                  ) LIMIT ?;", form.mission.id,
+                               options[:role].to_s,
+                               form.id,
+                               options[:limit]])
```
hooverlunch: Indent style here too please.

app/models/user.rb
```diff
@@ -146,6 +146,29 @@ def self.sorted_observer_response_counts(mission, limit)
...
   end
 
+  # Returns all non-admin users in the form's mission with the given role that have
+  # not submitted any responses to the form
+  #
+  # options[:role]) the role to check for
+  # options[:limit]) how many users we want to fetch from the db
+  def self.without_responses_for_form(form, options)
+    find_by_sql(["SELECT SQL_CALC_FOUND_ROWS * FROM users
+                  INNER JOIN assignments WHERE
```
hooverlunch: Should you use `ON` for the join attributes (`ON assignments.user_id = users.id`) and WHERE for the other conditions?

app/models/user.rb
```diff
@@ -146,6 +146,29 @@ def self.sorted_observer_response_counts(mission, limit)
...
+                  assignments.user_id = users.id AND
+                  assignments.mission_id = ? AND
+                  assignments.role = ? AND
+                  users.admin = FALSE AND
+                  users.id NOT IN (
+                    SELECT distinct users.id FROM users
+                    INNER JOIN responses WHERE
+                    responses.user_id = users.id AND
```
hooverlunch: Same here. See this answer: http://stackoverflow.com/questions/4160637/mysql-join-clause-vs-where-clause

hooverlunch:
Looks great! Thanks for changing it into a query. Nice work. Just a few refactor suggestions in my comments.

app/models/report/standard_form_report.rb
lmatiolis: Hum, i see your concern here... i separated them just to not have a method doing more stuff than it's supposed to (and to return user instances as it's expected).

What do you think of returning a hash instead of an array? Something like:
`{users: users_found, count: users_count }`. I think it's easier to use this when called, because you don't have to check on which position of the array the information you want is.

hooverlunch: Great idea.

On Tue, Sep 15, 2015 at 1:17 PM, Leandro Matioli <notifications@github.com>
wrote:

> In app/models/report/standard_form_report.rb
> <https://github.com/thecartercenter/elmo/pull/141#discussion_r39537856>:
>
> >    end
> >
> > -  def observers_without_responses
> > -    users_without_responses(:role => :observer)
> > +  # This returns the count of the previous users select without the limit clause.
> > +  # Format: [[count]]
> > +  def observers_without_responses_count
> > +    ActiveRecord::Base.connection.execute('SELECT FOUND_ROWS();').entries[0].first
>
> Hum, i see your concern here... i separated them just to not have a method
> doing more stuff than it's supposed to (and to return user instances as
> it's expected).
>
> What do you think of returning a hash instead of an array? Something like:
> {users: users_found, count: users_count }. I think it's easier to use
> this when called, because you don't have to check on which position of the
> array the information you want is.
>
> —
> <https://github.com/thecartercenter/elmo/pull/141/files#r39537856>.
>



-- 
Tom Smyth

touchstonecohousing.org

spec/models/report/standard_form_report_spec.rb
lmatiolis: Haha, i have the same feelings towards the other way around! :D
I'll change here to adhere to the project's style, no problem.
Thanks!

hooverlunch:
Looks great, please merge.

PR #143, opened 2015-10-01
3227 denormalize geographic answer data

lmatiolis:


md5:
:+1:

app/models/answer.rb
```diff
@@ -186,6 +181,10 @@ def should_validate?(field)
     end
   end
 
+  def simple_location_answer?
+    qtype.name == 'location' && !value.blank?
+  end
```
hooverlunch: Nice touch.

app/models/choice.rb
```diff
@@ -15,4 +17,13 @@ def checked
   def checked=(value)
     @checked = (value == true || value == '1')
   end
+
+  private
+
+  def replicate_location_values
+    if has_coordinates?
+      self.latitude = self.option.latitude
+      self.longitude = self.option.longitude
```
hooverlunch: Don't need `self.` on RHS.

db/migrate/20150930214725_denormalize_geographic_answer_data.rb
```diff
@@ -0,0 +1,66 @@
...
+      INNER JOIN `questions` ON `questions`.`id` = `form_items`.`question_id`
+      LEFT JOIN `option_sets` ON `questions`.`option_set_id` = `option_sets`.`id`
+      LEFT JOIN `options` ON `option_sets`.`allow_coordinates` = 1 AND `questions`.`qtype_name` = 'select_one' AND `answers`.`option_id` = `options`.`id`
+      LEFT JOIN `choices` ON `option_sets`.`allow_coordinates` = 1 AND `questions`.`qtype_name` = 'select_multiple' AND `answers`.`id` = `choices`.`answer_id`
+      LEFT JOIN `options` AS `choice_options` ON `choices`.`option_id` = `choice_options`.`id` WHERE (
+        (`questions`.`qtype_name` = 'location' AND `answers`.`value` IS NOT NULL)
+        OR (`options`.`latitude` IS NOT NULL AND `options`.`longitude` IS NOT NULL)
+        OR (`choice_options`.`latitude` IS NOT NULL AND `choice_options`.`longitude` IS NOT NULL)
+      )"])
+  end
```
hooverlunch: Nice! I like that you are restricting the update to only answers with location. Else this would take forever...

app/models/answer.rb
```diff
@@ -43,22 +44,16 @@ def self.location_answers_for_mission(mission, user = nil, options = {})
...
+        (`answers`.`latitude` IS NOT NULL AND `answers`.`longitude` IS NOT NULL)
+        OR (`choices`.`latitude` IS NOT NULL AND `choices`.`longitude` IS NOT NULL)
       })
       .select(:response_id,
-        %{COALESCE(`answers`.`value`,
-                   CONCAT(`options`.`latitude`, ' ', `options`.`longitude`),
-                   CONCAT(`choice_options`.`latitude`, ' ', `choice_options`.`longitude`)) AS `value`})
-      .distinct
+        %{COALESCE(CONCAT(`answers`.`latitude`, ' ', `answers`.`longitude`),
+                   CONCAT(`choices`.`latitude`, ' ', `choices`.`longitude`)) AS `value`})
```
hooverlunch: I'm not a huge fan of this one piece. I see that it is required in order to deprecate smoothly. So I guess it stays, but could we at least put a comment on it saying that it's temporary and perhaps listing the places in the app that depend on it.

Could we also refactor the dashboard map to use the new lat/lng cols instead of this? Shouldn't be hard. I think even that that code might be doing needless joins right now since it hasn't been touched at all in this PR.

hooverlunch:
Looking great. See my comments. Can we add just a few specs to ensure that:
* lat/lng get set properly for new location-type answer
* ditto for new select_one-geo answer
* the query works

hooverlunch:
PS You can go ahead and merge after that. Just let me know what you think about refactoring the dashboard map as I suggested.

app/models/answer.rb
lmatiolis: True! Thanks for pointing it out! I guess I just extracted the old code to DRY it up but didn't really look into it.

app/models/answer.rb
lmatiolis: Yep, that part wasn't changed to avoid possible collisions with old code. I'll add the commentary explaining that this should be temporary.

About the dashboard, I'm not sure i follow. This query is the one being used on the dashboard. Are we talking about the same thing?

lmatiolis: Oh, and about listing the places it's being used: A quick search here shows that the only place that it's calling it is on the welcome dashboard. (Unless it's being also called from dynamic segments...)

So maybe it's safe to change the query (and dashboard) to use the lat/long new columns... what do you think? Do you remember somewhere else that this could be in use?

hooverlunch: Nope go for it.

sent via wireless telephone machine.

lmatiolis:
@hooverlunch Can you please check the changes on the dashboard map and the new specs (to see if i didn't miss anything?) Thanks!

PR #144, opened 2015-10-01
3228: Make dashboard report use AJAX exclusively.

lmatiolis:
This change was done to improve load time on login and to show
the dashboard more quickly.

### DIFF
--- a/app/assets/javascripts/controllers/report/report_controller.js
+++ b/app/assets/javascripts/controllers/report/report_controller.js
@@ -3,43 +3,46 @@
 
   // constructor
   ns.ReportController = klass = function(init_data) {
-    this.dont_set_title = init_data.dont_set_title;
+    // Don't do anything if data wasn't provided.
+    if(init_data){
+      this.dont_set_title = init_data.dont_set_title;
 
-    // create supporting models unless in read only mode
-    if (!init_data.read_only) {
-      this.options = init_data.options;
-      this.menus = {
-        attrib: new ns.AttribMenu(this.options.attribs),
-        form: new ns.FormMenu(this.options.forms),
-        calc_type: new ns.CalcTypeMenu(this.options.calculation_types),
-        question: new ns.QuestionMenu(this.options.questions),
-        option_set: new ns.OptionSetMenu(this.options.option_sets)
+      // create supporting models unless in read only mode
+      if (!init_data.read_only) {
+        this.options = init_data.options;
+        this.menus = {
+          attrib: new ns.AttribMenu(this.options.attribs),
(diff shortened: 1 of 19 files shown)


app/assets/javascripts/controllers/report/report_controller.js
```diff
@@ -3,43 +3,46 @@
...
-    if (!init_data.read_only) {
-      this.options = init_data.options;
-      this.menus = {
-        attrib: new ns.AttribMenu(this.options.attribs),
-        form: new ns.FormMenu(this.options.forms),
-        calc_type: new ns.CalcTypeMenu(this.options.calculation_types),
-        question: new ns.QuestionMenu(this.options.questions),
-        option_set: new ns.OptionSetMenu(this.options.option_sets)
+    // Don't do anything if data wasn't provided.
+    if(init_data){
```
hooverlunch: Space after `if` please.

app/controllers/welcome_controller.rb
```diff
@@ -106,12 +106,11 @@ def dashboard_index
...
-        prepare_report
-      end
-
       # render JSON if ajax request
       if request.xhr?
+        # Load report asynchronously
+        prepare_report
+
         data = {
           recent_responses: render_to_string(partial: 'recent_responses'),
```
hooverlunch: This feels a bit hackish right now.

I think a better way to do this would be:
* To make a new endpoint on ReportsController called /data or so
* Make both the dashboard and reports#show use it, so reports#show only loads via AJAX also. This would be better also since having a loading indicator on part of the page is better than having the whole page take a long time  to load.
* Then we wouldn't need to load _main.html.erb at all on non AJAX requests, so the awkward if block in ReportController can go away, right?

hooverlunch:
Good start. See my comments above. Thanks.

app/assets/javascripts/views/report_ajax_loader.js
```diff
@@ -0,0 +1,80 @@
...
+// View model to load the report via ajax. Used on dashboard and report show.
+(function(ns, klass) {
+
+  // constructor
+  ns.ReportAjaxLoader = klass = function(params, dashboard) { var self = this;
+    self.dashboard = dashboard;
+    self.params = params;
+
+    // save the report id
+    if (params) {
```
hooverlunch: When would params ever not be given?

lmatiolis: I don't know... the params check is old code.

hooverlunch: Ok nrmind.

PR #146, opened 2015-10-02
3229: Add low level caching for reports.

lmatiolis:
This is used on the welcome dashboard and when showing a separate
report.

### DIFF
--- a/app/controllers/concerns/report_embeddable.rb
+++ b/app/controllers/concerns/report_embeddable.rb
@@ -28,15 +28,30 @@ module ReportEmbeddable
     @report_data[:report][:generated_at] = I18n.l(Time.zone.now)
   end
 
-  # runs the report and handles any errors, adding them to the flash
+  # Looks for a cached, populated report object matching @report.
+  # If one is found, stores it in @report. If not found,
+  # calls run on the existing @report
+  #
   # returns true if no errors, false otherwise
-  def run_and_handle_errors
+  def run_or_fetch_and_handle_errors
     begin
-      @report.run(current_ability)
+      @report = Rails.cache.fetch(self.cache_key_with_responses) do
+        @report.run(current_ability)
+        @report
+      end
+
       return true
     rescue Report::ReportError, Search::ParseError
       flash.now[:error] = $!.to_s
       return false
     end
+  end
+
+  def cache_key_with_responses
+    [
+      I18n.locale.to_s,
+      Response.per_mission_cache_key(current_mission),
+      @report.cache_key
+    ].join('-')
   end
 end
--- a/app/models/report/report.rb
+++ b/app/models/report/report.rb
@@ -78,10 +78,11 @@ class Report::Report < ActiveRecord::Base
   end
 
   # records a viewing of the form, keeping the view_count up to date
+  # It's using the update_column to avoid it updating the updated_at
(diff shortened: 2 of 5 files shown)


app/controllers/concerns/report_embeddable.rb
```diff
@@ -2,6 +2,7 @@
 module ReportEmbeddable
   # sets up the @report_data structure which will be converted to json
   def build_report_data(options = {})
+    @report = Rails.cache.read(@report.cache_key_with_responses(current_mission))
```
hooverlunch: I think cache_key_with_responses probably belongs in ReportEmbeddable. It seems like Report shouldn't need to know about Response and I18n.

app/controllers/concerns/report_embeddable.rb
```diff
@@ -32,7 +33,10 @@ def build_report_data(options = {})
   # returns true if no errors, false otherwise
   def run_and_handle_errors
     begin
-      @report.run(current_ability)
+      Rails.cache.fetch(@report.cache_key_with_responses(current_mission)) do
+        @report.run(current_ability)
+        @report
+      end
```
hooverlunch: I'm not super happy with the structure here. It seems like line 5 should not be necessary.

How about this: 
* we rename this method to run_or_fetch_and_handle_errors`
* add a comment saying `# Looks for a cached, populated report object matching @report. If one is found, stores it in @report. If not found, calls run on the existing @report`
* add `@report = ` to the front of line 36.

Does this work? Then we can remove line 5 and a lot of confusion (at least for me!)

We can also get rid of return true and return false. I don't think they are used.

app/models/report/report.rb
```diff
@@ -78,10 +78,11 @@ def run(current_ability = nil)
...
 
   # records a viewing of the form, keeping the view_count up to date
+  # It's using the update_column to avoid it updating the updated_at
+  # value ( which would invalidate the cache ). It also skips validations.
   def record_viewing
-    self.viewed_at = Time.now
-    self.view_count += 1
-    save(:validate => false)
+    update_column(:viewed_at, Time.now)
+    update_column(:view_count, self.view_count += 1)
```
hooverlunch: Should have been this way all along!

hooverlunch:
Generally good work here! See my suggestion for a simplifying refactor.

app/controllers/concerns/report_embeddable.rb
lmatiolis: @hooverlunch The problem is that `@report.run` doesn't return the `@report`. So what we would be caching instead is the return of @report.run if we removed the last line on the block.

I don't like the way it is right now too. If we could change the return of all `run` methods to return `self`, we could do the way you suggested, but i'm not sure if this would affect other parts of the app. It looks like (from a quick search) that the return of run is not being used, but you must know better.

Let me know what you think.

Thanks!

hooverlunch: I don't think that's necessary. The block should return @report due to line 38. Or if you want to change run to return self that's fine too. We could then get rid of line 38.

lmatiolis: Oh, i misinterpreted what you were asking for. (I though the problem was on the block code). I'll be doing these changes (including the change to self) It will make everything clearer.

Thanks!

PR #148, opened 2015-10-13
3014: Updates backbone view to prevent double form submission

meltheadorable:
Submitting the from with "enter" was both submitting the HTML form and triggering the save call in backbone causing the duplicate groups and the weird redirect.


hooverlunch:
Perfect, thanks.

PR #152, opened 2015-10-20
3390: Fixes option set export bug

meltheadorable:
Fixes bug where option set names longer than 31 characters or with special characters would cause the export to fail.

* Adds `worksheet_name` method to OptionSet model
* Updates export view to use `worksheet_name` for the export

### DIFF
--- a/app/models/option_set.rb
+++ b/app/models/option_set.rb
@@ -293,6 +293,23 @@ class OptionSet < ActiveRecord::Base
     end
   end
 
+  def worksheet_name
+    name = self.name
+    name = name.truncate(31) if self.name.size > 31
+    name = name.gsub(
+      /[\[\]\*\\?\:\/]/, {
+        '[' => '(',
+        ']' => ')',
+        '*' => '∗',
+        '?' => '',
+        ':' => '-',
+        '\\' => '-',
+        '/' => '-'
+      }
+    )
+    name
+  end
+
   # Returns a string representation, including options, for the default locale.
   def to_s
     "Name: #{name}\nOptions:\n#{root_node.to_s_indented}"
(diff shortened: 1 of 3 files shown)


spec/models/option_set_spec.rb
```diff
@@ -75,4 +75,17 @@
...
+      set = create(:option_set)
+      expect(set.worksheet_name).to eq set.name
+    end
+
+    it 'should return a replaced name if invalid' do
+      set = create(:option_set, name: 'My Options?: ~*[Yes/No:No\Yes]*~')
+      expect(set.worksheet_name.size).to be <= 31
+      expect(set.worksheet_name).to eq 'My Options- ~∗(Yes-No-No-Ye...'
+    end
+  end
```
hooverlunch: Nice.

app/models/option_set.rb
```diff
@@ -293,6 +293,23 @@ def as_json(options = {})
...
+        '*' => '∗',
+        '?' => '',
+        ':' => '-',
+        '\\' => '-',
+        '/' => '-'
+      }
+    )
+    name
+  end
+
```
hooverlunch: Did it not work to put this as a helper method? Just curious.

Typically I would say by LRP, OptionSet probably shouldn't know about worksheets, but it's not a big deal.

meltheadorable: I probably could find _somewhere_ to define it where that particular kind of view has access to it, I can try to relocate this piece, but the first couple of places I tried didn't work.

hooverlunch: I mean you could always do a PORO -- WorksheetBuilder or something. But like I say, not a big deal.

PR #154, opened 2015-10-22
3391: Fixes bug updating multilevel option sets

meltheadorable:
Updates to multi-level option sets with read-only levels tried to permit child attributes that weren't present. This checks to make sure that the key for children to permit is present before continuing.

* Updates `permit_children` method in parameters concern with check for key presence


hooverlunch:
:+1:

PR #155, opened 2015-10-22
3323: Fixes low activity users bug

meltheadorable:
Updates low activity query to properly constrain the results by mission id.


hooverlunch:
:+1:

PR #156, opened 2015-10-28
3268 odk hint

jpdimond:


hooverlunch:
Don't forget to put issue number in commit messages

app/models/qing_group.rb
```diff
@@ -1,9 +1,12 @@
...
   include Translatable
 
-  translates :group_name
+  translates :group_name, :group_hint
 
   replicable child_assocs: :children, backward_assocs: :form, dont_copy: [:form_id]
 
   alias_method :c, :children
+
+  scope(:child_groups, ->(children) { children.where(type: "QingGroup").all })
```
hooverlunch: I don't think you want the `.all` here. Typically you don't put that in a scope. I think it actually has no effect.

app/views/forms/_odk_group.xml.erb
```diff
@@ -4,7 +4,11 @@
   <% grid_mode = grid_mode?(children) %>
 
   <group appearance="field-list">
-    <label><%= group.group_name %></label>
+    <label ref="jr:itext('name<%= group.id %>:label')"/>
+
+    <input ref="<%= "/data/hint#{group.id}"%>">
+        <hint ref="jr:itext('hint<%= group.id %>:hint')"/>
```
hooverlunch: Too much tab.

app/views/forms/show.xml.erb
```diff
@@ -13,6 +13,13 @@
             <data id="<%= @form.id %>" version="<%= @form.current_version.sequence %>">
           <% end %>
 
+          <%# group name and notes %>
+          <% groups = QingGroup.child_groups(@form.root_group.children) %>
```
hooverlunch: `children` is delegated to root_group on form, so you can just do @form.children. LoD says you want to avoid chaining method calls wherever possible. So always look for delegators and create them where appropriate.

app/views/forms/show.xml.erb
```diff
@@ -40,6 +47,16 @@
           <% configatron.preferred_locales.each do |lang| %>
             <translation lang="<%= @style == 'commcare' ? lang : language_name(lang) %>">
 
+              <%# group hint translations %>
+              <% groups.each do |group| %>
+                <text id="name<%= group.id %>:label">
+                  <value><%= group.group_name(lang, string: false) %></value>
```
hooverlunch: I think you mean `strict: false`?

app/views/forms/show.xml.erb
```diff
@@ -40,6 +47,16 @@
...
+                </text>
+                <text id="hint<%= group.id %>:hint">
+                  <value><%= group.group_hint(lang, string: false) %></value>
```
hooverlunch: Here too.

spec/expectations/grid_form_odk.xml
```diff
@@ -31,13 +39,18 @@
           </text>
         </translation>
       </itext>
+      <bind nodeset="/data/name2" readonly="true()" type="string"/>
```
hooverlunch: It seems like this binding is not used at all. Is it needed? (And the name2 tag above?)

spec/expectations/grid_form_odk.xml
```diff
@@ -5,12 +5,20 @@
     <model>
       <instance>
         <data id="1" version="1">
+          <name2/>
+          <hint2/>
```
hooverlunch: Perhaps `group-hint2` or `grp-hint2`? hint alone is ambiguous, from a readability perspective.

spec/models/qing_group_spec.rb
```diff
@@ -0,0 +1,14 @@
+require "spec_helper"
+
+describe QingGroup do
+  context "One QingGroup" do
+    before do
+      @form = FactoryGirl.create(:sample_form, question_types: [['text', 'text', 'text']])
```
hooverlunch: The sample_form factory is deprecated. You can just use regular `:form`. Could you actually put a comment on it? It was a bad idea.

spec/models/qing_group_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+      group = create(:qing_group, form: @form, ancestry: @form.root_group.id)
```
hooverlunch: This should not be necessary I think. You can use the form factory directly to create a form with several groups.

hooverlunch:
image

I think the hint box should be a good bit bigger, perhaps a short text area the same size as the hint textarea in the question form. The group name field should also probably be the same width.

app/views/forms/_odk_group.xml.erb
```diff
@@ -4,7 +4,11 @@
   <% grid_mode = grid_mode?(children) %>
 
   <group appearance="field-list">
-    <label><%= group.group_name %></label>
+    <label ref="jr:itext('name<%= group.id %>:label')"/>
```
hooverlunch: In testing it looks like if the label is here, it is not bold. If you move it inside the input tag below, it will become bold and look better.

Then, however, this really becomes a "group header" device, not group hint. So I would change all the names to e.g. grp-header instead of hint. Make sense?

hooverlunch:
Note: you were missing group_hint in the permit params code for qing group controller, so the hints were not saving. I fixed and committed (needed to do so to be able to test). It's a good idea to always test the full form lifecycle including refreshing the page and verifying your data made it. Also checking the logs is good, since you would have seen an 'unpermitted param' notice.

hooverlunch:
Overall, this looks awesome! Great work! I would say you can merge once my comments are addressed. Please be careful when tweaking the modal form styles. They are finicky for some reason. And let's talk about the scope piece.

PR #161, opened 2015-11-02
3471: Adds searching on options and choices

meltheadorable:
* Updates answer model with methods for option names
* Updates excerpt code to check option name fields
* Updates answer index to search associations


hooverlunch:
VERY impressive. I am amazed at how simple this was to add. Using canonical_name means we are not searching option translations, but that should be ok. Thanks!

PR #162, opened 2015-11-02
3399: SMS Credentials Settings Fix

meltheadorable:
* Updates view with new 'Clear' checkboxes
* Updates backbone view with logic for checkboxes
* Fixes Setting model to incorporate logic


hooverlunch:
Great work! I fixed one minor bug. Other than that, looks great. Good solution. I will merge.

PR #165, opened 2015-11-11
3010: Repeat groups for ODK incoming

meltheadorable:
* Refactors Response model
* Creates XMLSubmission model
* Adds group number to answers
* Parse submission XML via Nokogiri instead of LibXML


app/models/answer.rb
```diff
@@ -200,6 +200,10 @@ def has_coordinates?
     end
   end
 
+  def from_group?
+    return questioning && questioning.parent && questioning.parent.type == 'QingGroup'
```
hooverlunch: Please remove `return` per style guide.

app/models/xml_submission.rb
```diff
@@ -0,0 +1,69 @@
+class XMLSubmission
+  attr_accessor :response, :data
+
+  def initialize(response: nil, data: nil)
+    @response = response
+    @data = data
+    populate_from_odk(@data) if @data.is_a? String
+    populate_from_j2me(@data) if @data.is_a? Hash
```
hooverlunch: I'm not sure about this heuristic. Didn't the old code have a check for an xmlns or something?

app/models/xml_submission.rb
```diff
@@ -0,0 +1,69 @@
...
+    # if form has no version, error
+    raise "xml submissions must be to versioned forms" if form.current_version.nil?
+
+    # if form version is outdated, error
+    raise FormVersionError.new("form version is outdated") if form.current_version.sequence > params[:version].to_i
+  end
+
+  def populate_from_odk(xml)
+    # Response mission should already be set
+    raise "Submissions must have a mission" if @response.mission.nil?
```
hooverlunch: Perhaps this should move to `populate_from_hash`?

app/models/xml_submission.rb
```diff
@@ -0,0 +1,69 @@
...
+      group = child if child.elements.present?
+      if group
+        group.elements.each do |c|
+          hash[c.name] ||= []
+          hash[c.name] << c.try(:content)
+        end
+      else
+        hash[child.name] = child.try(:content)
+      end
+    end
```
hooverlunch: Very elegant!

spec/models/xml_submission_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    @response = create(:response, form: @form)
+    @data = build_odk_submission(@form)
+  end
+
+  describe '.new' do
+    it 'creates a submission and parses it to populate response' do
+      submission = XMLSubmission.new(response: @response, data: @data)
+      response = submission.response
+      response.answers.each_with_index do |answer|
+        expect(answer.group_number).to eq nil unless answer.from_group?
```
hooverlunch: Should this spec be checking `group_number` a little more closely? Seems like it would be good to actually test this with some repeated groups in the submission XML? Or is that going to come later?

hooverlunch:
Overall, great work! I am impressed! I wonder if we need a little more test coverage of the group stuff. Let's discuss.

spec/models/xml_submission_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    @data = build_odk_submission(@form, repeat: true)
+  end
+
+  describe '.new' do
+    it 'creates a submission and parses it to populate response' do
+      submission = XMLSubmission.new(response: @response, data: @data)
+      response = submission.response
+      response.answers.each_with_index do |answer|
+        expect(answer.group_number).to eq nil unless answer.from_group?
+        expect(response.answers.where('group_number > ?', 1).count).to eq 2
```
meltheadorable: This is now checking for the presence of repeated group answers

PR #168, opened 2015-12-28
3007: Question groups are SMS-Able

meltheadorable:
* Makes `smsable_questionings` return an indexed hash
* Changes questionings object in SMS decoder to `smsable_questionings` hash
* Prints smsable_questionings index in SMS guide

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -175,9 +175,10 @@ class Form < ActiveRecord::Base
     questionings.reject{|q| q.hidden?}
   end
 
-  # returns questionings that work with sms forms and are not hidden
+  # returns hash of questionings that work with sms forms and are not hidden
   def smsable_questionings
-    questionings.reject{|q| q.hidden? || !q.question.qtype.smsable?}
+    smsable_questionings = questionings.reject{|q| q.hidden? || !q.question.qtype.smsable?}
+    smsable_questionings.index_by.with_index { |_, i| i + 1 }
   end
 
   def questioning_with_code(c)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -178,7 +178,7 @@ class Form < ActiveRecord::Base
   # returns hash of questionings that work with sms forms and are not hidden
   def smsable_questionings
     smsable_questionings = questionings.select(&:smsable?)
-    smsable_questionings.index_by.with_index { |_, i| i + 1 }
+    smsable_questionings.map.with_index { |q, i| [i + 1, q] }.to_h
   end
 
   def questioning_with_code(c)
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -177,7 +177,7 @@ class Form < ActiveRecord::Base
 
   # returns hash of questionings that work with sms forms and are not hidden
   def smsable_questionings
-    smsable_questionings = questionings.reject{|q| q.hidden? || !q.question.qtype.smsable?}
+    smsable_questionings = questionings.select(&:smsable?)
     smsable_questionings.index_by.with_index { |_, i| i + 1 }
   end
 
--- a/app/views/forms/sms_guide.html.erb
+++ b/app/views/forms/sms_guide.html.erb
@@ -42,7 +42,7 @@
     <%# loop over each question %>
     <% @form.smsable_questionings.each do |index, qing| %>
       <tr>
-        <td class="rank"><%= qing.full_dotted_rank %>.</td>
+        <td class="rank"><%= index %>.</td>
         <td class="question">
           <%= qing.question.name %>
           <%= reqd_sym if qing.required? %>
--- a/app/views/forms/sms_guide.html.erb
+++ b/app/views/forms/sms_guide.html.erb
@@ -40,7 +40,7 @@
     </tr>
 
     <%# loop over each question %>
-    <% @form.smsable_questionings.each do |qing| %>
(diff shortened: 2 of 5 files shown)


app/models/form.rb
```diff
@@ -175,9 +175,10 @@ def visible_questionings
     questionings.reject{|q| q.hidden?}
   end
 
-  # returns questionings that work with sms forms and are not hidden
+  # returns hash of questionings that work with sms forms and are not hidden
   def smsable_questionings
-    questionings.reject{|q| q.hidden? || !q.question.qtype.smsable?}
+    smsable_questionings = questionings.reject{|q| q.hidden? || !q.question.qtype.smsable?}
```
hooverlunch: Please add spaces around curlies per style guide (I know they weren't there before). The one below is good.

app/models/form.rb
```diff
@@ -175,9 +175,10 @@ def visible_questionings
...
+    smsable_questionings.index_by.with_index { |_, i| i + 1 }
```
hooverlunch: Cute!

app/views/forms/sms_guide.html.erb
```diff
@@ -48,7 +48,7 @@
           <%= reqd_sym if qing.required? %>
         </td>
         <td class="answers">
-          <%= answer_space(" #{qing.rank}.") %>
+          <%= answer_space(" #{index}.") %>
```
hooverlunch: Wow this worked out quite nicely...

app/views/forms/sms_guide.html.erb
```diff
@@ -40,7 +40,7 @@
     </tr>
 
     <%# loop over each question %>
-    <% @form.smsable_questionings.each do |qing| %>
+    <% @form.smsable_questionings.each do |index, qing| %>
       <tr>
         <td class="rank"><%= qing.full_dotted_rank %>.</td>
```
hooverlunch: I think this should also be changed to `index` since decoding error messages will refer to the bare integer rank, not the full dotted one, and will confuse the heck out of everyone.

hooverlunch:
Stellar work! Just a few small changes!

app/models/form.rb
md5: This is the first time I've seen this idiom and while I agree that it's cute, it's also sort of confusing.

meltheadorable: I'm open to suggestions. I mean, it's longer but there's always

```ruby
smsable_hash = {}
smsable_questionings.each_with_index do |qing, index|
  smsable_hash[index + 1] = qing
end
smsable_hash
```

md5: Don't get me wrong. I actually like this code and what it does is
straightforward. I still don't fully understand *how* it works though.

hooverlunch: Hmm. So index_by returns an enumerator that ... bleh I admit I'm stumped too. Would take some more digging through Rails source to figure it out.

How about `smsable_questionings.each_with_index{ |q, i| [i, q] }.to_h`

meltheadorable: `with_index` is a method on Enumerator which essentially adds an index to an existing enumerator, so `['cat', 'dog', 'bunny'].map.with_index{ |animal, index| "#{index}:#{animal}" }` would give you `["0:cat", "1:dog", "2:bunny"]`

`index_by` is more complicated than `map`, and ruby is doing some weird stuff behind the scenes to make Enumerators work that I don't *fully* understand because the entire class is written in raw C, but there's basically some kind of delegation back to the original Enumerator happening, and `with_index` is just some sort of a wrapper.

meltheadorable: I think this would need to be `smsable_questionings.each_with_index.map{ |q, i| [i + 1, q] }.to_h`, which could also be written `smsable_questionings.map.with_index { |q, i| [i + 1, q] }.to_h` which might clarify a little how `index_by` replacing `map` works in the current code.

md5: The part that I'm having a hard time grokking is how the block given to `with_index` ends up being the block used by the `yield` in `index_by`. I've looked through both the relevant Ruby code and the C code and I still don't get it, though that's clearly my problem since it works great :confounded: 

I definitely think the `smsable_questionings.map.with_index { |q, i| [i + 1, q] }.to_h` code is clearer.

meltheadorable: Yeah, I think part of the confusion might be that the C code shown in ruby's documentation site is severely clipped and more or less just farms out the work to methods defined elsewhere. I struggle reading C so it's still a mystery to me looking at the full source, but maybe you'll find https://github.com/ruby/ruby/blob/ruby_2_2/enumerator.c#L495 more interesting.

I'm fine with changing it to the map with index though, I have a preference for clear code over clever code.

md5: Thanks for the pointer @meltheadorable. I was actually looking at that code already and I'm still scratching my head since I too don't know much C.

meltheadorable: Neither do I :smile_cat: I'm just taking for granted that, however enumerators work, they manage to call the block correctly even when passed through a chain.

md5: I appreciate the exchange. I ended up picking up a digital copy of [*Ruby Under a Microscope*](http://patshaughnessy.net/ruby-under-a-microscope) since I'm not satisfied with my understanding of the MRI internals. Looking at the Rubinius implementation of `with_index` is helpful, but not helpful enough. :+1:

PR #169, opened 2015-12-28
3620: Fixes ODK option download bug

meltheadorable:


app/helpers/odk_helper.rb
```diff
@@ -99,7 +99,14 @@ def multi_level_option_nodeset_ref(qing, cur_subq)
...
   def odk_option_translations(form, lang)
-    form.all_first_level_option_nodes.map{ |on| %Q{<text id="on#{on.id}"><value>#{on.option.name(lang, strict: false)}</value></text>} }.join.html_safe
+    odk_options = form.all_first_level_option_nodes.collect do |on|
+      content_tag(:text, id: "on#{on.id}") do
+        content_tag(:value) do
+          on.option.name(lang, strict: false)
+        end
+      end
+    end
+    odk_options.join.html_safe
```
hooverlunch: I think if you use `odk_options.reduce(&:<<)` you shouldn't need `html_safe`. I like to avoid it wherever possible.

meltheadorable: Sure thing, I just didn't wanna go any further past the estimate unless you thought it was worthwhile to add a new spec when the issue was just an unescaped HTML entity.

PR #171, opened 2015-12-30
3468: Allow partial matches

meltheadorable:
* Changes several fields for search qualifiers to allow partial matching


hooverlunch:
Looks great, merge pls.

PR #172, opened 2015-12-30
3885 group hints and wrapping

meltheadorable:


hooverlunch:
This looks good but the hint icons are not totally in alignment still:

image

as you can see from the last one. I think the way to fix this is to make the `div.control` narrower for group members than for top level questions. (Could instead make the label narrower, but then the control wouldn't appear indented, which I think is a good cue that it's part of the group.)

Can we try it this way?

meltheadorable:
I'll give it a shot!

hooverlunch:
I don't want to fire this up again to look at it so I'm trusting that it does what I suggested. You can go ahead and close :)

PR #173, opened 2016-01-05
2185: Adds hostname & rails environment on error

meltheadorable:
* Adds hostname & rails environment to error reports


hooverlunch:
Lovely, merge please.

PR #174, opened 2016-01-20
3879: csv imports

meltheadorable:


app/models/user_batch.rb
```diff
@@ -38,22 +38,21 @@ def create_users(mission)
     end
 
     # parse the input file as a spreadsheet
-    @data = Roo::Spreadsheet.open(file)
+    @data = Roo::Spreadsheet.open(file).parse
 
-    validate_headers
+    validate_headers(@data.shift)
```
hooverlunch: Can we rename this method to `parse_headers`? Makes way more sense that way.

app/models/user_batch.rb
```diff
@@ -269,7 +257,7 @@ def check_uniqueness_on_db(objects, fields)
     add_results_errors_on_objects(results, objects, fields) unless results.nil?
   end
 
-  def add_results_errors_on_objects(results, objects, fields)
+  def add_results_errors_on_objects(results, _, fields)
```
hooverlunch: Perhaps just get rid of that param?

app/models/user_batch.rb
```diff
@@ -286,15 +274,16 @@ def correct_field_key(field)
...
-                                                  'assignments',
-                                                  'user_id',
-                                                  'users',
-                                                  'import_num')
+    DirectDBConn.new('assignments').insert_select(
+      users_batch,
+      'assignments',
+      'user_id',
+      'users',
+      'import_num')
```
hooverlunch: Could we just make this a two-liner?

hooverlunch:
That was quick! Nicely done!

hooverlunch:
Looks great. Two small comments in the first new commit, then you can merge. I don't need to see again.

PR #175, opened 2016-01-21
3906 media questions

meltheadorable:


.gitignore
```diff
@@ -27,3 +27,5 @@ config/*.sphinx.conf
 rake_tasks.txt
 routes.txt
 
+# file uploads
+uploads
```
hooverlunch: Prepend with a slash?

Gemfile
```diff
@@ -84,7 +85,7 @@ gem 'rails-backbone', github: 'codebrew/backbone-rails'
 gem 'rack-attack', github: 'sassafrastech/rack-attack'
 
 # reCAPTCHA support
-gem "recaptcha", :require => "recaptcha/rails"
+gem 'recaptcha', require: 'recaptcha/rails'
```
hooverlunch: Hooray for code niceness.

app/assets/javascripts/views/question_form.js
```diff
@@ -65,6 +65,23 @@
...
 
+    // show/hide key question
+    var hide_key_q = (
+      selected_type == "image" ||
+      selected_type == "annotated_image" ||
+      selected_type == "signature" ||
+      selected_type == "sketch" ||
+      selected_type == "audio" ||
+      selected_type == "video"
+    );
```
hooverlunch: I think we should bite the bullet and transfer the list of multimedia question types to a JS constant someplace. Or even better you could transfer the full set of QuestionTypes as JSON. They are used all over the place. Then here you could just do .index(..) != -1

app/controllers/api/v1/answers_controller.rb
```diff
@@ -17,6 +17,10 @@ def index
         return render json: { errors: ["access_denied"] }, status: 403
       end
 
+      if question.multimedia?
+        return render json: { errors: ["question_type_not_api_accessible"] }, status: 422
+      end
```
hooverlunch: Nice.

app/controllers/media/objects_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  # GET /media/objects/1
+  # GET /media/objects/1.json
+  def show
+    style = params[:style]
+    send_file @media_object.item.path(style), type: @media_object.item_content_type, disposition: 'inline'
+  end
+
+  private
+    # Use callbacks to share common setup or constraints between actions.
```
hooverlunch: I think we can get rid of all the comments in this file. Did this come along with a gem or something? I don't think this was part of the issue spec?

app/models/form.rb
```diff
@@ -172,7 +172,7 @@ def questions(reload = false)
   end
 
   def visible_questionings
-    questionings.reject{|q| q.hidden?}
+    questionings.reject{ |q| q.hidden? || q.multimedia? }
```
hooverlunch: Yay. I think the styleguide also calls for a space before the `{`.

app/models/media/audio.rb
```diff
@@ -0,0 +1,4 @@
+class Media::Audio < Media::Object
+  validates_attachment_content_type :item, content_type: [%r{\Aaudio/.*\Z}, 'video/webm']
+  validates_attachment_file_name :item, matches: /(mp3|ogg|webm|wav)\Z/i
```
hooverlunch: Is there a reason not to put a `\.` before the parenthetical in the regexp?

app/models/media/image.rb
```diff
@@ -0,0 +1,8 @@
+class Media::Image < Media::Object
+  has_attached_file :item, styles: {
+    normal: '720x720>',
+    thumb: '100x100#'
+  }
+  validates_attachment_content_type :item, content_type: %r{\Aimage/.*\Z}
+  validates_attachment_file_name :item, matches: /(png|jpe?g)\Z/
```
hooverlunch: Same here.

app/models/form.rb
meltheadorable: It might, but my linter doesn't catch that at the moment. Still tweaking its options. I can change it.

app/models/media/image.rb
```diff
@@ -0,0 +1,8 @@
+class Media::Image < Media::Object
+  has_attached_file :item, styles: {
+    normal: '720x720>',
+    thumb: '100x100#'
+  }
```
hooverlunch: I assume this call in addition to the one in the parent class doesn't cause issues? Another option would be to define a `styles` class method that returns the styles hash and then you'd call that from the parent...

app/models/media/audio.rb
meltheadorable: I'm not sure if there's any particular reason, it's omitted in the documentation examples for paperclip's file name validations, but it could be an oversight rather than intentional on their part.

app/models/question_type.rb
```diff
@@ -3,16 +3,22 @@ class QuestionType
...
+    { name: 'long_text', odk_name: 'string', properties: %w(form_printable smsable textual) },
+    { name: 'integer', odk_name: 'int', properties: %w(form_printable smsable numeric headerable) },
+    { name: 'decimal', odk_name: 'decimal', properties: %w(form_printable smsable numeric headerable) },
+    { name: 'location', odk_name: 'geopoint', properties: %w() },
+    { name: 'select_one', odk_name: 'select1', properties: %w(form_printable has_options smsable headerable) },
+    { name: 'select_multiple', odk_name: 'select', properties: %w(form_printable has_options smsable headerable) },
+    { name: 'datetime', odk_name: 'dateTime', properties: %w(form_printable temporal has_timezone smsable headerable) },
+    { name: 'date', odk_name: 'date', properties: %w(form_printable temporal smsable headerable) },
+    { name: 'time', odk_name: 'time', properties: %w(form_printable temporal smsable headerable) },
+    { name: 'image', odk_name: 'image/*', properties: %w(multimedia) },
```
hooverlunch: Good for you for looking up image/*, but I think the correct value for odk_name for all of these is `binary`. Ultimately this info should be moved out of this class as it violates LRP.

app/models/question_type.rb
```diff
@@ -39,42 +45,47 @@ def human_name
...
   # returns whether this type has a timezone
   def has_timezone?
-    properties.include?("has_timezone")
+    properties.include?('has_timezone')
   end
 
   # returns whether this type is temporal
   def temporal?
-    properties.include?("temporal")
+    properties.include?('temporal')
```
hooverlunch: I just remembered that our styleguide says to always use double quoted strings. Can you fix your linter to adhere to that?

app/models/media/image.rb
meltheadorable: No issues, as far as I can tell. Under the hood, this is just using `define_method` for a bunch of different methods, so the child overrides the parent. It successfully defines and generates the files for those attachments when they're created.

config/initializers/paperclip.rb
```diff
@@ -0,0 +1,17 @@
+Paperclip::Attachment.default_options.merge!(
+  path: ':rails_root/uploads/:class/:attachment/:id_partition/:style/:filename',
+  url: '/uploads/:class/:id/:style'
```
hooverlunch: I don't think the URL thing is right. I guess we can change later. Ultimately I'd like /en/m/foobar/media/6352. This comes up in issue 3908.

config/initializers/paperclip.rb
```diff
@@ -0,0 +1,17 @@
...
+)
+
+Paperclip.options[:content_type_mappings] = {
+  wmv: 'application/vnd.ms-asf',
+  opus: 'audio/x-opus+ogg'
+}
+
+module Paperclip
```
hooverlunch: Detailed comment here please explaining why this is necessary. Turning off a security feature so need to explain.

app/models/question_type.rb
meltheadorable: I had gotten that from https://opendatakit.org/help/form-design/body/ which made no mention of `binary` as a type and I couldn't find any other reference for what the valid types were. Can make that change, but I'm curious where I should have been looking to find that information if not there, for next time.

db/migrate/20160118184307_create_media_objects.rb
```diff
@@ -0,0 +1,11 @@
+class CreateMediaObjects < ActiveRecord::Migration
+  def change
+    create_table :media_objects do |t|
+      t.references :answer, index: true, foreign_key: true
+      t.text :annotation
```
hooverlunch: What is this for?

db/schema.rb
```diff
@@ -79,21 +79,12 @@
...
-  create_table "condition_option_nodes", force: :cascade do |t|
-    t.integer "condition_id", limit: 4, null: false
-    t.integer "option_node_id", limit: 4, null: false
-    t.integer "rank", limit: 4, null: false
-  end
-
-  add_index "condition_option_nodes", ["condition_id"], name: "index_condition_option_nodes_on_condition_id", using: :btree
-  add_index "condition_option_nodes", ["option_node_id"], name: "index_condition_option_nodes_on_option_node_id", using: :btree
-  add_index "condition_option_nodes", ["rank"], name: "index_condition_option_nodes_on_rank", using: :btree
-
```
hooverlunch: Oops, I must have added this in. My bad.

app/controllers/media/objects_controller.rb
meltheadorable: Just from the rails resource generator, the controller and associated route were necessary for testing paperclip's overrides because of the non-public file store. I can remove it, but I wouldn't have been able to figure out if the paperclip configuration was right if I hadn't built it first. Will remove the comments though.

spec/models/media/image_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+describe Media::Image do
+    include_examples 'accepts file extensions', %w(jpg png)
+    include_examples 'rejects file extensions', %w(tiff)
+    include_examples 'rejects file types', %w(audio video)
```
hooverlunch: Indent is off here.

app/models/question_type.rb
meltheadorable: Can do.

spec/support/media_spec_helpers.rb
```diff
@@ -0,0 +1,65 @@
...
+      File.open(path)
+    end
+  end
+
+  shared_context 'media helpers' do
+    def factory_name(described_class)
+      described_class.name.underscore.gsub('/', '_').to_sym
+    end
+
+    def file_for_type(file_type)
```
hooverlunch: If this method is only used to get fixtures for the reject case, perhaps you could remove it and just pass `mp3`, `jpg`, etc. directly to the shared example.

spec/models/media/video_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
+
+describe Media::Video do
+  include_examples 'accepts file extensions', %w(3gp mp4 webm mpeg wmv avi)
+  include_examples 'rejects file extensions', %w(ogv)
+  include_examples 'rejects file types', %w(image audio)
+end
```
hooverlunch: These look really spiffy.

hooverlunch:
Very nice work! Just a few comments and you can merge away!

app/models/question_type.rb
hooverlunch: https://opendatakit.org/help/form-design/examples/#video_prompts

image

config/initializers/paperclip.rb
meltheadorable: i'm not entirely sure how their interpolation works under the hood, but i'm not 100% sure if that's possible. I think we may have to denormalize the mission slug onto the media model if we're going to make that work with acceptable performance, and I'd have to figure out what we can do about the locale. You're not actually defining like, a proper rails routing string there, so I think all of the `:blah` things in the string need to be methods on the model with the attachment. We can try a few things, and I can dig deeper into their docs on the related issue, but it could be difficult.

hooverlunch: Let's wait till we get to that bridge before we cross it.

On Thu, Jan 21, 2016 at 2:53 PM, Melody <notifications@github.com> wrote:

> In config/initializers/paperclip.rb
> <https://github.com/thecartercenter/elmo/pull/175#discussion_r50453592>:
>
> > @@ -0,0 +1,17 @@
> > +Paperclip::Attachment.default_options.merge!(
> > +  path: ':rails_root/uploads/:class/:attachment/:id_partition/:style/:filename',
> > +  url: '/uploads/:class/:id/:style'
>
> i'm not entirely sure how their interpolation works under the hood, but
> i'm not 100% sure if that's possible. I think we may have to denormalize
> the mission slug onto the media model if we're going to make that work with
> acceptable performance, and I'd have to figure out what we can do about the
> locale. You're not actually defining like, a proper rails routing string
> there, so I think all of the :blah things in the string need to be
> methods on the model with the attachment. We can try a few things, and I
> can dig deeper into their docs on the related issue, but it could be
> difficult.
>
> —
> <https://github.com/thecartercenter/elmo/pull/175/files#r50453592>.
>



-- 
Tom Smyth

touchstonecohousing.org

app/models/question_type.rb
meltheadorable: Gotcha, so `odk_name` refers specifically to the part that goes into the `bind` and we'll need to incorporate the `mediatype` thing separately when we deal with the rendering portion?

config/initializers/paperclip.rb
meltheadorable: Yeah, I wasn't intending to dig into it now, just letting you know that based on what I know now, it isn't straightforward.

spec/support/media_spec_helpers.rb
meltheadorable: since they have different file names & directories, i'd either need to pass the file path into the shared example, or move that same `case` logic into the example group definition. 

The reason this works differently from the other specs is because I can't leverage the factories as much for it, with traits I can create a `:media_audio` with an `mp3` given only the type & the extension, but I can't create a `:media_video` with a `png` without setting the item directly, since the `png` trait is only valid & defined for the `:media_image` factory. I could make all of the traits completely generic to all media objects, in which case I'd be able to handle this based on extensions alone, but that makes the `webm` case weird since it's valid for multiple content types, and seems like a worse practice in general, since it loses the structure & most of those combinations wouldn't be valid.

hooverlunch: Ah ok nevermind. Looks great.

On Thu, Jan 21, 2016 at 3:07 PM, Melody <notifications@github.com> wrote:

> In spec/support/media_spec_helpers.rb
> <https://github.com/thecartercenter/elmo/pull/175#discussion_r50455761>:
>
> > @@ -0,0 +1,65 @@
> > +module MediaSpecHelpers
> > +  module FileHandling
> > +    def media_fixture(name)
> > +      path = File.expand_path("../../fixtures/media/#{name}", __FILE__)
> > +      File.open(path)
> > +    end
> > +  end
> > +
> > +  shared_context 'media helpers' do
> > +    def factory_name(described_class)
> > +      described_class.name.underscore.gsub('/', '_').to_sym
> > +    end
> > +
> > +    def file_for_type(file_type)
>
> since they have different file names & directories, i'd either need to
> pass the file path into the shared example, or move that same case logic
> into the example group definition.
>
> The reason this works differently from the other specs is because I can't
> leverage the factories as much for it, with traits I can create a
> :media_audio with an mp3 given only the type & the extension, but I can't
> create a :media_video with a png without setting the item directly, since
> the png trait is only valid & defined for the :media_image factory. I
> could make all of the traits completely generic to all media objects, in
> which case I'd be able to handle this based on extensions alone, but that
> makes the webm case weird since it's valid for multiple content types,
> and seems like a worse practice in general, since it loses the structure &
> most of those combinations wouldn't be valid.
>
> —
> <https://github.com/thecartercenter/elmo/pull/175/files#r50455761>.
>



-- 
Tom Smyth

touchstonecohousing.org

app/models/question_type.rb
hooverlunch: Right. That would actually perhaps be a good time to strip odk_name out of
here.

On Thu, Jan 21, 2016 at 2:55 PM, Melody <notifications@github.com> wrote:

> In app/models/question_type.rb
> <https://github.com/thecartercenter/elmo/pull/175#discussion_r50453932>:
>
> > -    {:name => "select_one", :odk_name => "select1", :properties => %w(form_printable has_options smsable headerable)},
> > -    {:name => "select_multiple", :odk_name => "select", :properties => %w(form_printable has_options smsable headerable)},
> > -    {:name => "datetime", :odk_name => "dateTime", :properties => %w(form_printable temporal has_timezone smsable headerable)},
> > -    {:name => "date", :odk_name => "date", :properties => %w(form_printable temporal smsable headerable)},
> > -    {:name => "time", :odk_name => "time", :properties => %w(form_printable temporal smsable headerable)}
> > +    { name: 'text', odk_name: 'string', properties: %w(form_printable smsable textual headerable) },
> > +    { name: 'long_text', odk_name: 'string', properties: %w(form_printable smsable textual) },
> > +    { name: 'integer', odk_name: 'int', properties: %w(form_printable smsable numeric headerable) },
> > +    { name: 'decimal', odk_name: 'decimal', properties: %w(form_printable smsable numeric headerable) },
> > +    { name: 'location', odk_name: 'geopoint', properties: %w() },
> > +    { name: 'select_one', odk_name: 'select1', properties: %w(form_printable has_options smsable headerable) },
> > +    { name: 'select_multiple', odk_name: 'select', properties: %w(form_printable has_options smsable headerable) },
> > +    { name: 'datetime', odk_name: 'dateTime', properties: %w(form_printable temporal has_timezone smsable headerable) },
> > +    { name: 'date', odk_name: 'date', properties: %w(form_printable temporal smsable headerable) },
> > +    { name: 'time', odk_name: 'time', properties: %w(form_printable temporal smsable headerable) },
> > +    { name: 'image', odk_name: 'image/*', properties: %w(multimedia) },
>
> Gotcha, so odk_name refers specifically to the part that goes into the
> bind and we'll need to incorporate the mediatype thing separately when we
> deal with the rendering portion?
>
> —
> <https://github.com/thecartercenter/elmo/pull/175/files#r50453932>.
>



-- 
Tom Smyth

touchstonecohousing.org

hooverlunch:
Lovely. All done. Thanks.

PR #176, opened 2016-01-22
3903: Fixes translation call

meltheadorable:


hooverlunch:
Bizarre. I wonder how that never came up until now... Anyway, all good!

PR #177, opened 2016-01-22
2962: Fixes SMS error when missing dot on first answer

meltheadorable:
* Raises an error if the first answer token doesn't look like the start of an answer

### DIFF
--- a/app/models/sms/decoder.rb
+++ b/app/models/sms/decoder.rb
@@ -35,10 +35,10 @@ class Sms::Decoder
     check_permission
 
     # create a blank response
-    @response = Response.new(:user => @user, :form => @form, :source => "sms", :mission => @form.mission)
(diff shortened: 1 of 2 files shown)


app/models/sms/decoder.rb
```diff
@@ -53,9 +53,9 @@ def decode
 
         # look up the questioning object for the specified rank and store it in the @qing instance variable
         find_qing
-
-      # otherwise, we add the token to the value variable and proceed
-      else
+      elsif index == 0 # if this is the first token, raise error
+        raise_decoding_error("question_doesnt_exist", rank: tok)
```
hooverlunch: Should we perhaps add a more descriptive error message for this one? e.g. "Sorry, but '#{token}' is not a valid answer. Be sure to include a period after the question number."

hooverlunch:
Just one comment. You can then merge. Nice work.

PR #179, opened 2016-01-28
Update production-setup.md

evilGmonkey:
Ruby fails to build if you don't install libffi-dev.


hooverlunch:
Thanks @evilGmonkey . What version of the system are you presently running on your servers?

evilGmonkey:
am using  v5.2.4

PR #181, opened 2016-02-11
3935: Fixes cache key for list reports by ensuring correct ordering for calculations

meltheadorable:


app/models/report/report.rb
```diff
@@ -59,7 +59,7 @@ def self.terminate_sub_relationships(report_ids)
   def cache_key
     chunks = [super]
     chunks << option_set_choices.map(&:option_set_id)
-    chunks << ("calcs-" << (calculations.order(updated_at: :desc).first.try(:cache_key) || "none"))
+    chunks << ("calcs-" << (calculations.reorder(updated_at: :desc).first.try(:cache_key) || "none"))
```
hooverlunch: Ah, clever. Can you also please add the total calculation count to the hash key. Currently if you delete a calculation (except the most recent one) the cache key doesn't change and so the report won't regenerate, right?

meltheadorable: Oh, you're right, yes I can add the count to the cache key.

PR #182, opened 2016-02-11
3908: Media objects in response view

meltheadorable:
* Adds thumbnails for audio/video
* Updates download filenames for media files
* Adds `media_object` association to `Answer`
* Delegates mission to response through answer
* Adds interpolations for mission and locale
* Updates media object show route
* Unhides multimedia questions for response view
* Updates initializer with new paperclip URL specification


app/controllers/media/objects_controller.rb
```diff
@@ -16,4 +22,9 @@ def set_media_object
   def media_object_params
     params.require(:media_object).permit(:answer_id, :annotation)
   end
+
+  def media_filename
+    extension = File.extname(@media_object.item_file_name)
+    "elmo-#{@response.id}-#{@answer.id}#{extension}"
```
hooverlunch: Very nice.

app/models/answer.rb
```diff
@@ -28,6 +29,7 @@ class Answer < ActiveRecord::Base
   delegate :question, :qtype, :required?, :hidden?, :multimedia?, :option_set, :options, :condition, to: :questioning
   delegate :name, :hint, to: :question, prefix: true
   delegate :name, to: :level, prefix: true, allow_nil: true
+  delegate :mission, to: :response
```
hooverlunch: Nice use of delegation.

config/initializers/paperclip.rb
```diff
@@ -1,6 +1,6 @@
 Paperclip::Attachment.default_options.merge!(
   path: ":rails_root/uploads/:class/:attachment/:id_partition/:style/:filename",
-  url: "/uploads/:class/:id/:style"
+  url: "/:locale/m/:mission/:class/:id/:style"
```
hooverlunch: I'm not sure if we need `:locale` in here. Answers are not locale dependent. What was your reasoning for including it?

app/models/media/object.rb
```diff
@@ -5,4 +5,13 @@ class Media::Object < ActiveRecord::Base
...
   delegate :url, to: :item
+  delegate :mission, to: :answer
+end
+
+Paperclip.interpolates :mission do |attachment, style|
+  attachment.instance.mission.compact_name
+end
+
+Paperclip.interpolates :locale do |attachment, style|
+  I18n.locale
```
hooverlunch: Do we need this? See below...

config/initializers/paperclip.rb
meltheadorable: Your spec:

> URLs should look like /en/m/foobar/media/6352

app/views/responses/_answer_read_only.html.erb
```diff
@@ -31,6 +31,18 @@
...
+  <%= link_to image_tag(answer.media_object.url(:thumb)), answer.media_object.url %>
+
+<% when 'audio' %>
+
+  <%= link_to image_tag('media/audio.png'), answer.media_object.url %>
+
+<% when 'video' %>
+
+  <%= link_to image_tag('media/video.png'), answer.media_object.url %>
+
```
hooverlunch: This looks great. May not have been clear from issue but these should also appear in editable answer view.

meltheadorable: Will add.

app/controllers/media/objects_controller.rb
```diff
@@ -4,11 +4,17 @@ class Media::ObjectsController < ApplicationController
 
   def show
     style = params[:style]
-    send_file @media_object.item.path(style), type: @media_object.item_content_type, disposition: "inline"
+    @answer = @media_object.answer
```
hooverlunch: I'm realizing that I think we need to authorize access to these objects. So will need to add to ability, etc.

config/initializers/paperclip.rb
meltheadorable: `media` is interpolated by `:class` in this case, it gives the fully qualified class name separated by slashes, so images will be in `media/images`, while audio is in `media/audio`, etc. The route expects `media` explicitly so that I don't need to parse the subclass from the params.

app/controllers/media/objects_controller.rb
meltheadorable: I wasn't sure if authorization was part of another sub-issue, or its own issue, I figured that we'd be doing something with authorizing access to media files though, otherwise there would be little reason not to serve them from accessible folders. I can do that here though.

hooverlunch:
These look amazing:

image


One small note: please add a 5px bottom margin to each thumbnail so the spacing is more even.

hooverlunch:
Note: I changed disposition to inline so that the browser will display what it can. Fewer steps to view something that way. If the browser doesn't know what to do with it it will download it anyway.

hooverlunch:
Fantastic work overall. Really pleased!

meltheadorable:
Hmm, my browser was still displaying the images in the image tags with `disposition: 'attachment'` but wasn't downloading audio/video (instead, it was trying to play it in a new tab), are you saying the browser opening the content in a tab is more desirable?

PR #185, opened 2016-03-02
1394: Generates SMS auth codes for users

meltheadorable:


app/assets/stylesheets/all/users.css.scss
```diff
@@ -3,12 +3,17 @@
...
+  }
+
   td.name {
-    width: 100px;
+    width: 150px;
   }
 
   td.value {
     font-weight: bold;
+    font-family: monospace;
```
hooverlunch: Did you mean to use the SCSS variable here?

app/models/user.rb
```diff
@@ -310,6 +311,14 @@ def regenerate_api_key
     save
   end
 
+  # regenerates sms auth code
+  def regenerate_sms_auth_code
+    begin
+      self.sms_auth_code = Random.alphanum(4)
+    end while User.exists?(sms_auth_code: sms_auth_code)
+    save
+  end
```
hooverlunch: Lovely.

app/models/user.rb
```diff
@@ -310,6 +311,14 @@ def regenerate_api_key
     save
   end
 
+  # regenerates sms auth code
+  def regenerate_sms_auth_code
+    begin
+      self.sms_auth_code = Random.alphanum(4)
+    end while User.exists?(sms_auth_code: sms_auth_code)
+    save
```
hooverlunch: `validate: false`?

app/models/user.rb
```diff
@@ -400,6 +409,8 @@ def phone_should_be_unique
       end
     end
 
+
+
```
hooverlunch: Why the extra blanks?

lib/tasks/db_add_sms_auth_code.rake
```diff
@@ -0,0 +1,9 @@
+namespace :db do
+  desc "Add SMS auth codes to users"
+  task :add_sms_auth_codes => :environment do
+    User.find_each do |user|
+      auth_code = Random.alphanum(4)
+      user.update_column(:sms_auth_code, auth_code)
+    end
+  end
+end
```
hooverlunch: Can this be in a migration instead please? Else I will forget to run it on all the different servers.

spec/requests/sms_auth_code_field_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    before(:all) do
+      login(@observer)
+      get("/en/m/#{get_mission.compact_name}/users/#{@user.id}/edit")
+    end
+
+    it 'should not be visible' do
+      assert_select('div.user_sms_auth_code', false)
+    end
+  end
+end
```
hooverlunch: Very nice.

app/models/ability.rb
```diff
@@ -147,6 +147,8 @@ def initialize(params)
             can :create, Sms::Test
           end
 
+          can :regenerate_sms_auth_code, User
+
```
hooverlunch: I'm realizing that it would probably make sense for an observer to be able to see and regenerate their own auth code (but only their own). Could you add this please?

hooverlunch:
Looks fantastic. A few requested changes and you're good to merge. Thanks.

app/assets/stylesheets/all/users.css.scss
meltheadorable: I originally had, but there's no actual way to use the defined variables in other files, then forgot to un-define the variable since I couldn't use it. If I had more time, I'd have tried to refactor the variables out into their own file that could be included in the other files and then fix all the places that manually re-define font-family stacks, but that seems like an issue for the backlog since it's a little more extensive.

lib/tasks/db_add_sms_auth_code.rake
meltheadorable: Sure, I'll move it.

app/models/ability.rb
meltheadorable: Yep, will do.

app/models/user.rb
meltheadorable: I think I might've moved the method from there to another part of the file and my linter didn't catch the extra blanks. I'll take care of it.

app/models/user.rb
meltheadorable: Could you explain why you don't want to validate here? This method has more or less identical behavior to the API key generation which doesn't add `validate: false` there, just curious if I should be changing both, or if you noticed something breaking that shouldn't be due to validation issues.

hooverlunch: Yes they should both be `validate: false`. In my book, calling `save` without `!` or `validate: false` should only be done if you're intending to check the errors object. Else you have a recipe for silent, hard-to-debug failure.

PR #187, opened 2016-03-17
3920: Adds media create/delete, cleanup job

meltheadorable:


app/controllers/media/objects_controller.rb
```diff
@@ -29,4 +45,17 @@ def media_filename
...
+  def media_class(type)
+    case type
+    when 'audio'
+      return Media::Audio
+    when 'video'
+      return Media::Video
+    when 'image'
+      return Media::Image
+    else
+      return Media::Object
```
hooverlunch: When would this ever happen? Should we be throwing an error here?

app/controllers/media/objects_controller.rb
```diff
@@ -15,6 +15,22 @@ def show
       filename: media_filename
   end
 
+  def create
+    media = media_class(params[:type]).new(item: params[:upload])
+    media.answer = Answer.find(params[:answer_id]) if params[:answer_id]
```
hooverlunch: Perhaps a comment here explaining why it's ok to leave answer_id blank. It's not super clear.

app/jobs/media_cleanup_job.rb
```diff
@@ -0,0 +1,9 @@
+class MediaCleanupJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    Media::Object.expired.find_each do |media|
+      media.destroy
+    end
```
hooverlunch: One liner using `(&:blah)` syntax?

config/routes.rb
```diff
@@ -69,8 +69,10 @@
       end
     end
 
-    namespace :media do
+    namespace :media, type: /audio|image|video/ do
       get ':type/:id(/:style)' => 'objects#show', defaults: { style: 'original' }
+      post ':type' => 'objects#create', as: :create
+      delete ':type/:id' => 'objects#delete', as: :delete
```
hooverlunch: Very nice, clean, lovely.

config/schedule.rb
```diff
@@ -15,4 +15,7 @@
 
   # make sure the daemon is running
   rake "ts:start"
-end
\ No newline at end of file
+
+  # Clean up expired media
+  runner "MediaCleanupJob.perform_later"
```
hooverlunch: I'm worried that this may end up spinning up two Rails processes at once. Could we perhaps create an 'hourly' rake task that does these things? Could perhaps use `ThinkingSphinx::RakeInterface#start` to do the Sphinx thing...

spec/requests/media/objects_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  describe 'delete media object' do
+    let(:media_file) { create(:media_image) }
+
+    it 'returns empty 204 on success' do
+      id = media_file.id
+      delete media_delete_path(media_file, mission_name: mission.compact_name, type: 'image')
+      expect(response).to have_http_status 204
+    end
+  end
+end
```
hooverlunch: Dandy!

hooverlunch:
This looks fantastic. The only thing I'd like to see again is the solution for running hour tasks. I think the 'hourly' rake task may be the way to go... Curious for your thoughts though. Just don't want to eat up a bunch of memory unnecessarily.

app/controllers/media/objects_controller.rb
meltheadorable: It shouldn't ever happen, the `type` comes from the URL params and the scope restricts to the valid types, so that condition should be more-or-less unreachable entirely. Raising an error there wouldn't be a bad idea.

hooverlunch: Ok let's do that. I generally prefer failing loudly when something that
shouldn't happen does.

On Thu, Mar 17, 2016 at 7:59 PM, Melody <notifications@github.com> wrote:

> In app/controllers/media/objects_controller.rb
> <https://github.com/thecartercenter/elmo/pull/187#discussion_r56597010>:
>
> > @@ -29,4 +45,17 @@ def media_filename
> >      extension = File.extname(@media_object.item_file_name)
> >      "elmo-#{@response.id}-#{@answer.id}#{extension}"
> >    end
> > +
> > +  def media_class(type)
> > +    case type
> > +    when 'audio'
> > +      return Media::Audio
> > +    when 'video'
> > +      return Media::Video
> > +    when 'image'
> > +      return Media::Image
> > +    else
> > +      return Media::Object
>
> It shouldn't ever happen, the type comes from the URL params and the
> scope restricts to the valid types, so that condition should be
> more-or-less unreachable entirely. Raising an error there wouldn't be a bad
> idea.
>
> —
> You are receiving this because you commented.
> <https://github.com/thecartercenter/elmo/pull/187/files/f4cd545baffd958d5a3e0f390a5944ca0edb251f#r56597010>
>



-- 
Tom Smyth

touchstonecohousing.org

config/schedule.rb
meltheadorable: I'm not sure about the rake task, the index option on the rake interface takes a few options from the task. I wouldn't want to duplicate & recreate that portion of TS's task code. Also not big on interacting with sphinx through an internal API, could be an update headache if they decide to refactor that code.

The choice to run that hourly was more or less entirely arbitrary, mostly because there was other stuff already in that block in the schedule. It could run less frequently, or more frequently (but with something other than cron). I know there's at least one option for scheduling ActiveJobs that uses the after callback to do the re-enqueuing, and that option wouldn't spin up an additional rails process.

hooverlunch:
Well done.

PR #188, opened 2016-03-17
3888: Implements auth code checking

meltheadorable:


app/models/sms/decoder.rb
```diff
@@ -97,6 +101,32 @@ def find_form
...
+    end
+
+    def log_authentication_failure
+      @msg.auth_failed = true
+      @msg.save(validate: false)
+      raise_decoding_error("user_not_found")
+    end
+
+    def check_for_brute_force
+      messages =  Sms::Message.where(user: @user).since(1.minute.ago).where(auth_failed: true)
```
hooverlunch: Extra space after `=`.

app/models/sms/decoder.rb
```diff
@@ -97,6 +101,32 @@ def find_form
...
+      raise_decoding_error("account_locked") if messages.count >= 3
```
hooverlunch: Can we move `3` and `1.minute` into constants? Also I realized that my spec was a little weird. I like the way you've implemented it. I think that the minimum interval for brute force tries would equal the lookback window divided by the threshold. Right now that is 1 minute / 3 == 20 seconds. Probably fine. But good to have these as constants: `BRUTE_FORCE_CHECK_WINDOW` and `BRUTE_FORCE_LOCK_THRESHOLD` or something. Perhaps also put the formula I'm suggesting in a comment above it. (If you think I'm correct!)

app/models/sms/message.rb
```diff
@@ -20,6 +20,7 @@ class Sms::Message < ActiveRecord::Base
 
   # order by id after created_at to make sure they are in creation order
   scope(:latest_first, ->{ order('created_at DESC, id DESC') })
+  scope :since, -> (time) { where('created_at > ?', time) }
```
hooverlunch: Nice touch.

db/migrate/20160304205331_add_auth_failed_to_sms_messages.rb
```diff
@@ -0,0 +1,5 @@
+class AddAuthFailedToSmsMessages < ActiveRecord::Migration
+  def change
+    add_column :sms_messages, :auth_failed, :boolean
```
hooverlunch: Can we have default: false and null: false please? I like to avoid nulls in boolean columns unless they have an explicit meaning.

spec/models/sms/sms_decoder_spec.rb
```diff
@@ -24,6 +24,38 @@
...
+  context 'with SMS Authentication enabled' do
+    before(:all) { create_form(questions: %w(integer), authenticate_sms: true) }
+
+    it 'should work with the correct code provided' do
+      auth_code = user.sms_auth_code
+      assert_decoding(body: "#{auth_code} #{@form.code} 1.17", answers: [17])
+    end
+
+    it 'should raise an error with an incorrect code provided' do
+      auth_code = 'n000'
```
hooverlunch: lol

app/models/sms/decoder.rb
meltheadorable: I'll file that one under "things my linter weirdly doesn't catch"

spec/models/sms/sms_decoder_spec.rb
```diff
@@ -24,6 +24,38 @@
...
+    end
+
+    it 'should raise an error if no code is provided' do
+      assert_decoding_fail(body: "#{@form.code} 1.17", error: "user_not_found")
+    end
+
+    it 'should lock account after 3 failed attempts' do
+      create_list(:sms_message, 3, user: user, auth_failed: true, type: "incoming")
+      assert_decoding_fail(body: "#{@form.code} 1.17", error: "account_locked")
+    end
```
hooverlunch: Can we get one additional test using Timecop that verifies that 3 failed attempts spaced out by more than window/threshold don't trigger the account locking?

spec/models/sms/sms_decoder_spec.rb
```diff
@@ -370,7 +402,8 @@
 
   def create_form(options)
     option_names = options[:default_option_names] ? nil : %w(Apple Banana Cherry Durian) + ["Elder Berry"]
-    @form = create(:form, smsable: true, question_types: options[:questions], option_names: option_names)
+    authenticate_sms = options[:authenticate_sms] ? options[:authenticate_sms] : false
+    @form = create(:form, smsable: true, question_types: options[:questions], option_names: option_names, authenticate_sms: authenticate_sms)
```
hooverlunch: Fantastic set of tests!

hooverlunch:
Really nice work!

hooverlunch:
Just a few comments, and then ready to go.

db/migrate/20160304205331_add_auth_failed_to_sms_messages.rb
meltheadorable: Will do.

app/models/sms/decoder.rb
meltheadorable: Sure thing, I think it makes sense to have those as constants, and I'll make a note of the formula at the same time.

PR #191, opened 2016-03-29
4130 option node shortcodes

meltheadorable:


app/models/option_node.rb
```diff
@@ -30,11 +30,16 @@ class OptionNode < ActiveRecord::Base
...
   def self.preload_child_options(roots)
-    ancestries = roots.map{ |r| "'#{r.id}'" }.join(',')
-    nodes_by_root_id = OptionNode.includes(:option).where("ancestry IN (#{ancestries})").order('rank').group_by{ |n| n.ancestry.to_i }
-    roots.each{ |r| r.child_options = (nodes_by_root_id[r.id] || []).map(&:option) }
+    ancestries = roots.map { |r| "'#{r.id}'" }.join(",")
+    nodes_by_root_id = OptionNode.includes(:option).
+      where("ancestry IN (#{ancestries})").
+      order("rank").
+      group_by { |n| n.ancestry.to_i }
+    roots.each { |r| r.child_options = (nodes_by_root_id[r.id] || []).map(&:option) }
```
hooverlunch: Thx for the cleanup!

app/models/option_set.rb
```diff
@@ -170,8 +178,9 @@ def form_smsable?
 
   def option_has_answers?(option_id)
     # Do one query for all and cache.
-    @option_ids_with_answers ||= Answer.where(questioning_id: questionings.map(&:id),
-      option_id: descendants.map(&:option_id)).pluck('DISTINCT option_id')
+    @option_ids_with_answers ||= Answer.where(
+      questioning_id: questionings.map(&:id),
+      option_id: descendants.map(&:option_id)).pluck("DISTINCT option_id")
```
hooverlunch: pluck should probably go to next line?

app/models/option_set.rb
```diff
@@ -207,7 +216,11 @@ def question_count
   # gets number of questions by which a copy of this option set is used
   def copy_question_count
     if is_standard?
-      respond_to?(:copy_question_count_col) ? copy_question_count_col || 0 : copies.inject(0){|sum, c| sum += c.question_count}
+      if respond_to?(:copy_question_count_col)
+        (copy_question_count_col || 0)
```
hooverlunch: No need for parens..

app/models/option_set_import.rb
```diff
@@ -144,11 +147,12 @@ def load_and_clean_data
       # Get special columns i18n values
       id_header = 'Id'
       coordinates_header = I18n.t('activerecord.attributes.option.coordinates')
+      shortcode_header = I18n.t('activerecord.attributes.option.shortcode')
 
       # Find any special columns
       special_columns = {}
       headers.each_with_index do |h,i|
-        special_columns[i] = h.downcase.to_sym if [id_header, coordinates_header].include?(h)
+        special_columns[i] = h.downcase.to_sym if [id_header, coordinates_header, shortcode_header].include?(h)
```
hooverlunch: Split line?

config/initializers/base_36.rb
```diff
@@ -0,0 +1,20 @@
+module Base36
```
hooverlunch: Why initializers?

app/models/option_set.rb
```diff
@@ -297,19 +321,41 @@ def worksheet_name
...
+
+  def shortcode_offset
+    @shortcode_offset ||= Base36.offset(shortcode_length)
+  end
+
+  def generate_sequence(batch: false)
+    return unless (options_added? || options_removed? || batch)
+    descendants.each_with_index do |option_node, i|
+      option_node.sequence = i + 1
+      option_node.save(validate: false)
```
hooverlunch: Perhaps use `update_attribute`?

app/models/option_set_import.rb
```diff
@@ -105,8 +107,9 @@ def create_option_set
         end
       end
 
-      raise ActiveRecord::Rollback if errors.present?
+      @option_set.generate_sequence(batch: true)
```
hooverlunch: I can see why you'd want to do it this way as it reduces duplication, but I think it would be much more efficient to assign the sequence during the initial creation of each option node above. Option Sets can have 100,000's of options, so need to be super performance conscious here.

app/models/option_set.rb
```diff
@@ -24,56 +24,64 @@ class OptionSet < ActiveRecord::Base
   # the dependent object doesn't know who destroyed it.
   before_destroy { report_option_set_choices.each(&:option_set_destroyed) }
 
-  scope :by_name, -> { order('option_sets.name') }
+  # Generate sequence
+  after_save :generate_sequence
```
hooverlunch: I'm not sure about this approach because it would potentially re-number existing nodes, correct? That could be an issue if e.g. a few nodes get added to an existing set and the form is already out in the wild.

I think sequence num should be fixed once it is assigned.

Perhaps in OptionSet.update_children you could get the current max sequence num for the option set right before you create each node. It won't be great performancewise but that code needs to go eventually anyway, and it isn't used for large option set imports...

db/migrate/20160311174929_generate_sequences_for_option_nodes.rb
```diff
@@ -0,0 +1,15 @@
+class GenerateSequencesForOptionNodes < ActiveRecord::Migration
+  def up
+    OptionSet.find_each do |option_set|
+      option_set.generate_sequence(batch: true)
```
hooverlunch: This is fine for a migration I think. So I guess we're keeping batch mode, but can you please change generate_sequence to use find_each somehow? Loading 200k options into memory at once will likely bring down the server.

spec/models/option_node_spec.rb
```diff
@@ -1,289 +1,298 @@
...
   end
 
-  describe 'option_level' do
+  describe "shortcode" do
+    let(:option_set) { create(:option_set, super_multilevel: true) }
+
+    it "should return shortcodes based on sequence" do
+      shortcodes = option_set.descendants.map(&:shortcode)
+      expect(shortcodes).to eq ["1", "2", "3", "4", "5", "6", "7", "8", "9", "a", "b", "c", "d", "e"]
+    end
```
hooverlunch: Might be good to have a spec demostrating generating shortcodes for large numbers. Perhaps a quick model spec for Base36 directly. (So perhaps making Base36 a model?)

hooverlunch:
Looking good. Just some performance tweaks and I'll take another quick look. Let me know if you have questions.

app/models/base36.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def self.offset(length)
+    if length > 1
+      36 ** (length - 1)
+    else
+      0
+    end
+  end
+end
```
hooverlunch: Much better I think.

meltheadorable: Yep, It was originally in initializers because at first I was just going to patch in the necessary conversion methods into Fixnum and String, but then I thought better of it and forgot to break it out of initializers into `lib` or `models`.

app/models/option_node.rb
```diff
@@ -296,6 +297,12 @@ def shortcode
     @shortcode = Base36.to_padded_base36(sequence, length: shortcode_length)
   end
 
+  def max_sequence
+    # for some reason ancestry scopes requests through the current node even if you don't
+    # call where through self, so you need to explicitly call unscoped here
```
hooverlunch: Nice comment.

hooverlunch:
Looks fantastic. Shut it down!

app/models/option_node.rb
meltheadorable: It wasn't so much a catch as a long, headache-inducing, painful debugging session -- the comment's there because calling `unscoped` is really weird and feels hacky and I wanted to make absolutely 100% sure it was clear why that was happening to save somebody else the trouble of removing it and then wondering why sequence assignment was broken.

PR #196, opened 2016-04-28
4339 admin option set import

meltheadorable:


app/mailers/admin_mailer.rb
```diff
@@ -8,9 +8,9 @@ def error(exception, session = nil, params = nil, env = nil, user = nil)
     @params = params
     @env = env
     @user = user
-    @hostname = env['SERVER_NAME'] || env['HTTP_HOST'] || env['HTTP_ORIGIN']
+    @hostname = env['SERVER_NAME'] || env['HTTP_HOST'] || env['HTTP_ORIGIN'] if env
     @rails_env = Rails.env
-    path = (env && env['REQUEST_URI']) ? (": " + env['REQUEST_URI']) : ""
+    path = (env && env['REQUEST_URI']) ? (": " + env['REQUEST_URI']) : "" if env
```
hooverlunch: Existence of `env` is already checked in the initial condition. Not setting path would result in an error below.

app/views/option_sets/export.xlsx.axlsx
```diff
@@ -4,7 +4,12 @@ xlsx_package.workbook.tap do |wb|
   header_style = wb.styles.add_style(b: true, alignment: { horizontal: :center })
   row_style = wb.styles.add_style(locked: false)
 
-  locale = @option_set.mission.setting.preferred_locales.first || :en
+  if @option_set.mission
+    locale = @option_set.mission.setting.preferred_locales.first || :en
```
hooverlunch: Ugh LoD! 

(I know this is not your code).

Could we refactor though by adding a `default_locale` method to Mission that would delegate to a method of the same name in Setting that would do `preferred_locales.first`?

app/views/option_sets/export.xlsx.axlsx
```diff
@@ -4,7 +4,12 @@ xlsx_package.workbook.tap do |wb|
...
+  else
+    locale = :en
```
hooverlunch: I would use `I18n.default_locale` here and above instead of `:en` hardcoded.

spec/models/option_set_import_spec.rb
```diff
@@ -19,7 +19,30 @@
...
       name: name,
-      geographic?: false)
+      geographic?: false,
+      is_standard: false)
+
+    expect(option_set.levels).to be_nil
+    expect(option_set.level_names).to include('en' => 'Province')
+
+    expect(option_set.total_options).to eq(26)
+    expect(option_set.all_options).to include(have_attributes(canonical_name: "Kinshasa"))
```
hooverlunch: This code should be moved into a helper method, like `expect_correct_option_set(option_set)` or something. Right now it's all repeated.

hooverlunch:
Looks smooth! As I said before I worry someone is going to import a flippin huge option set and then try to replicate it and it will not be pretty (I think -- never tested it). But I guess we'll deal with that when it happens...

PR #203, opened 2016-05-13
4277: Adds authorization token for media objects

meltheadorable:


app/controllers/media/objects_controller.rb
```diff
@@ -5,9 +5,15 @@ class Media::ObjectsController < ApplicationController
...
     @answer = @media_object.answer
-    @response = @answer.response
+    @response = @answer.try(:response)
     disposition = (params[:dl] == "1") ? "attachment" : "inline"
-    authorize! :show, @response
+
+    if @response
+      authorize! :show, @response
+    else
+      authorized = @media_object.token == params[:token]
```
hooverlunch: Why is it necessary to assign this to a var? Could just be an `unless` directly?

app/controllers/media/objects_controller.rb
```diff
@@ -47,7 +53,7 @@ def media_object_params
 
   def media_filename
     extension = File.extname(@media_object.item_file_name)
-    "elmo-#{@response.id}-#{@answer.id}#{extension}"
+    "elmo-#{@response.try(:id)}-#{@answer.try(:id)}#{extension}"
```
hooverlunch: This seems like an odd way to fail. This will generate an invalid filename if response or answer are nil, right? Shouldn't this be failing earlier on in some cleaner way if so?

app/controllers/media/objects_controller.rb
```diff
@@ -5,9 +5,15 @@ class Media::ObjectsController < ApplicationController
...
+    @response = @answer.try(:response)
     disposition = (params[:dl] == "1") ? "attachment" : "inline"
-    authorize! :show, @response
+
+    if @response
+      authorize! :show, @response
+    else
+      authorized = @media_object.token == params[:token]
+      raise CanCan::AccessDenied.new("Not authorized to show media", :view, :media_object) unless authorized
+    end
```
hooverlunch: Won't CanCan complain in this case that no auth check has been performed? Shouldn't there be an explicit skip or something?

app/controllers/media/objects_controller.rb
meltheadorable: The filename will be `elmo--.extension` which is weird but not broken and won't break anything -- I'm not sure how friendly we need to make the user experience of re-downloading a media file that you just uploaded to an invalid response, but if response and/or answer are nil, that method fails for the normal  `show` case for thumbnails which _does_ need to work when the response has not been saved yet.

db/schema.rb
```diff
@@ -197,6 +200,7 @@
     t.string "item_file_name", limit: 255
     t.integer "item_file_size", limit: 4
     t.datetime "item_updated_at"
+    t.string "token", limit: 255
```
hooverlunch: I know the db schema is kind of borked but can you please remove all the other changes from this branch so that it's just the addition of token.

app/models/media/object.rb
```diff
@@ -7,12 +7,25 @@ class Media::Object < ActiveRecord::Base
...
   scope :expired, -> { where(answer_id: nil).where("created_at < ?", 12.hours.ago) }
 
   def download_url
     dl_url = url
     separator = (dl_url =~ /\?/) ? "&" : "?"
-    "#{dl_url}#{separator}dl=1"
+    "#{dl_url}#{separator}token=#{token}&dl=1"
+  end
+
+  def token_url(style: nil)
```
hooverlunch: Can there be a comment on this method explaining why this is necessary. The bug that necessitated this feature obviously wasn't obvious to us initially, so it would be good to capture the reasoning here.

app/controllers/media/objects_controller.rb
meltheadorable: It makes for a very long line, but it could just be an unless directly

hooverlunch:
Looking great! Please be sure this is adequately tested manually. If I'm right about the missing skip auth check, it would seem that that should have shown up in a manual test. Or if I'm wrong about it, then so much the better!

app/controllers/media/objects_controller.rb
meltheadorable: The explicit skip is already at the top of the controller because we aren't using the auto-authorizers since we're auth-ing the response and didn't write custom logic for media objects.

app/models/media/object.rb
meltheadorable: Sure thing

app/controllers/media/objects_controller.rb
hooverlunch: I meant unless on separate line, so block style, not one-liner.

app/controllers/media/objects_controller.rb
hooverlunch: I see, I didn't realize what this method was for. Nevermind.

PR #206, opened 2016-05-26
4040 user group index

meltheadorable:


app/assets/javascripts/backbone/views/user_groups_modal_view.js.coffee
```diff
@@ -0,0 +1,12 @@
...
+  el: '#user-groups-modal'
+
+  initialize: (params) ->
+    this.params = params
+    this.set_body(params.html)
+
+  set_body: (html) ->
+    # console.log(html)
+    @modal_body = $(@el).find('.modal-body')
+    console.log(@modal_body)
```
hooverlunch: Let's get rid of these console's please.

app/assets/javascripts/backbone/views/user_groups_modal_view.js.coffee
```diff
@@ -0,0 +1,12 @@
+class ELMO.Views.UserGroupsModalView extends Backbone.View
+  el: '#user-groups-modal'
+
+  initialize: (params) ->
+    this.params = params
+    this.set_body(params.html)
```
hooverlunch: Always s/this/@/ in coffeescript

app/assets/javascripts/backbone/views/user_groups_modal_view.js.coffee
```diff
@@ -0,0 +1,12 @@
...
+
+  set_body: (html) ->
+    # console.log(html)
+    @modal_body = $(@el).find('.modal-body')
```
hooverlunch: s/$(@el)/@$el/

app/assets/javascripts/backbone/views/user_groups_modal_view.js.coffee
```diff
@@ -0,0 +1,12 @@
...
+    console.log(@modal_body)
+    $('.modal-body').html(html)
```
hooverlunch: Should you be using the @modal_body var here?

All this could really be one line: `@$el.find('.modal-body').html(html)`

app/assets/javascripts/backbone/views/user_list_view.js.coffee
```diff
@@ -0,0 +1,21 @@
+class ELMO.Views.UserListView extends Backbone.View
+
+  el: '.link_set'
+
+  events:
+    'click .list-groups': 'show_groups_modal'
+
+  initialize: (params) ->
+    this.params = params
```
hooverlunch: @ instead of this

app/assets/javascripts/backbone/views/user_list_view.js.coffee
```diff
@@ -0,0 +1,21 @@
...
+
+  show_groups_modal: (event) ->
+    event.preventDefault()
+    ELMO.app.loading(true)
+    $.ajax
+      url: ELMO.app.url_builder.build('user_groups'),
+      method: "get",
+      data: {},
```
hooverlunch: I think you can leave this out. You also can remove the trailing commas. Coffee adds them based on indent.

app/assets/javascripts/backbone/views/user_list_view.js.coffee
```diff
@@ -0,0 +1,21 @@
...
+      success: (html) =>
+        new ELMO.Views.UserGroupsModalView({html: html})
+        ELMO.app.loading(false)
+        $("#user-groups-modal").modal('show')
```
hooverlunch: Seems like this should instead be a method invocation on the object you just built?

Also, general tip: try to avoid using `$` instead of `@$`. Doing so ensures your DOM access is properly scoped to the view's `el`.

app/helpers/user_groups_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+    case field_name
+    when "name"
+      group.name
+    when "users"
+      group.users.count
+    when "created_at"
+      l(group.created_at)
+    when "filter"
+      filter_link(group)
+    when "actions" then table_action_links(group)
```
hooverlunch: Seems incongruent that this line uses `when`?

app/helpers/user_groups_helper.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+  end
+
+  def filter_link(group)
+    qualifier = t('search_qualifiers.group')
+    link_text = t('activerecord.action_links.user_group.members')
```
hooverlunch: Best to use `I18n.t` here as `t` may produce enclosing <span>s and junk if translation is missing.

app/models/user.rb
```diff
@@ -9,17 +9,17 @@ class User < ActiveRecord::Base
...
+  has_many :assignments, autosave: true, dependent: :destroy, validate: true, inverse_of: :user
   has_many :missions, -> { order "missions.created_at DESC" }, through: :assignments
-  has_many :operations, :inverse_of => :creator, :foreign_key => :creator_id, :dependent => :destroy
-  has_many :reports, :inverse_of => :creator, :foreign_key => :creator_id, :dependent => :nullify, class_name: 'Report::Report'
-  has_many :user_groups, :dependent => :destroy
-  has_many :groups, :through => :user_groups
+  has_many :operations, inverse_of: :creator, foreign_key: :creator_id, dependent: :destroy
+  has_many :reports, inverse_of: :creator, foreign_key: :creator_id, dependent: :nullify, class_name: 'Report::Report'
+  has_many :user_group_assignments, dependent: :destroy
+  has_many :user_groups, through: :user_group_assignments, dependent: :destroy
```
hooverlunch: Have you tested if these delete cleanly on mission delete? Added to mission delete spec? If so, great, disregard.

app/models/user.rb
```diff
@@ -97,10 +97,11 @@ def self.find_by_credentials(login, password)
...
     [
-      Search::Qualifier.new(:name => "name", :col => "users.name", :type => :text, :default => true),
-      Search::Qualifier.new(:name => "login", :col => "users.login", :type => :text, :default => true),
-      Search::Qualifier.new(:name => "email", :col => "users.email", :type => :text, :default => true),
-      Search::Qualifier.new(:name => "phone", :col => "users.phone", :type => :text)
+      Search::Qualifier.new(name: "name", col: "users.name", type: :text, default: true),
+      Search::Qualifier.new(name: "login", col: "users.login", type: :text, default: true),
+      Search::Qualifier.new(name: "email", col: "users.email", type: :text, default: true),
+      Search::Qualifier.new(name: "phone", col: "users.phone", type: :text),
+      Search::Qualifier.new(name: "group", col: "user_groups.name", type: :text, assoc: :user_groups)
```
hooverlunch: Does this really work? Might be good to add a quick spec for this modeled off of question_search_spec?

app/models/user_group_assignment.rb
```diff
@@ -0,0 +1,4 @@
+class UserGroupAssignment < ActiveRecord::Base
```
hooverlunch: Nice naming here.

app/views/users/_user_groups_modal.html.erb
```diff
@@ -0,0 +1,25 @@
+<div id="user-groups-modal" class="modal fade" tabindex="-1" role="dialog" aria-labelledby="" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close group-cancel" data-dismiss="modal" aria-hidden="true">&times;</button>
+        <h4 class="modal-title">
+          <%= I18n.t("activerecord.action_links.user_group.index") %>
```
hooverlunch: Is `I18n.` needed here?

app/views/users/form.html.erb
```diff
@@ -20,6 +20,17 @@
...
 
+  <% if @user.user_groups.present? %>
+    <div class="form_field user_groups">
+      <div class="label_and_control">
+        <label class="main"><%= I18n.t('activerecord.attributes.user.user_groups') %>:</label>
+        <div class="control read_only">
+          <%= @user.user_groups.map(&:name).join(', ') %>
+        </div>
+      </div>
+    </div>
```
hooverlunch: I think there should be a way to do this using `f.field`. How about `f.field(:group_names, read_only: true)`, and you add a `group_names` method to user?

db/migrate/20160520200712_recreate_user_groups.rb
```diff
@@ -0,0 +1,10 @@
+class RecreateUserGroups < ActiveRecord::Migration
+  def change
+    create_table :user_groups do |t|
+      t.string :name
```
hooverlunch: Should this be null: false?

I am a huge stickler for db constraints. I think mission reference can also be non null too right, since you can't have standard-mode user groups?

db/migrate/20160520200852_create_user_group_assignments.rb
```diff
@@ -0,0 +1,11 @@
+class CreateUserGroupAssignments < ActiveRecord::Migration
+  def change
+    create_table :user_group_assignments do |t|
+      t.references :user, index: true, foreign_key: true
+      t.references :user_group, index: true, foreign_key: true
```
hooverlunch: These should both be null false also

spec/factories/user_groups.rb
```diff
@@ -1,8 +1,6 @@
-# Read about factories at https://github.com/thoughtbot/factory_girl
-
 FactoryGirl.define do
   factory :user_group do
-    user_id 1
-    group_id 1
+    name { Faker::Team.name }
+    mission
```
hooverlunch: I believe the project convention is to do `mission { get_mission }`.

spec/requests/ability_spec.rb
```diff
@@ -1,65 +1,27 @@
...
-    g.mission = get_mission
-    u = FactoryGirl.create(:user, :name => "Ada Nu User")
-    expect(@ability.cannot?(:create, UserGroup)).to be true
-  end
-
   def create_user_and_ability(options)
-    @user = FactoryGirl.create(:user, :role_name => options[:role])
-    @ability = Ability.new(:user => @user, :mission => get_mission)
+    @user = FactoryGirl.create(:user, role_name: options[:role])
+    @ability = Ability.new(user: @user, mission: get_mission)
```
hooverlunch: Should we have some UserGroup ability specs?

hooverlunch:
image

I think this means the model name translation for UserGroup is not defined. If you add it, this should read 'No User Groups found'

hooverlunch:
We need to add the new group qualifier to the search help box. I think we don't need to add an example though as people can access it through your nice filter link.

hooverlunch:
image

Should we sort alphabetically by name?

hooverlunch:
Do we need these buttons?

image

I'm not sure anything is being saved unless I'm missing something? Perhaps we can just remove the footer and let folks use the x?

hooverlunch:
Let's just make this 'User Groups' please:

image

hooverlunch:
Overall, great work! Looking very clean and tight. Thanks so much. I don't need to see it again.

app/views/user_groups/index.html.erb
```diff
@@ -0,0 +1 @@
+<%= render "index_table" %>
```
hooverlunch: Do we need this file? I don't think it's ever used, is it?

PR #215, opened 2016-06-30
4738: fixes broken tagging feature

meltheadorable:
* fixes regression caused by issues with activemodel serializers


app/controllers/tags_controller.rb
```diff
@@ -1,10 +1,10 @@
 class TagsController < ApplicationController
-  skip_authorization_check :only => [:suggest]
+  skip_authorization_check only: [:suggest]
 
   def suggest
     @tags = Tag.suggestions(current_mission, params[:q])
     respond_to do |format|
-      format.json { render json: @tags.as_json }
+      format.json { render json: @tags, root: "tags" }
```
hooverlunch: 👍

PR #250, opened 2016-08-17
4936: Fixing Google Maps issues

meltheadorable:


hooverlunch:
Looks good. Do we have a plan for updating all the instances to add google map api keys? I usually keep a deploy checklist for things like this as it's easy to forget.

meltheadorable:
This didn't actually fix the issue consistently and sometimes the issue was fixed without this being included, the buggy behavior requires further investigation as it might be browser-specific or otherwise intermittent, I'm gonna go ahead and close this until it's been investigated further, as managing the API keys has a number of other complications and an alternative solution might be preferred when the source of the bug is actually identified.

PR #253, opened 2016-08-18
4647: Creates new reviewer role

meltheadorable:


app/helpers/elmo_form_builder.rb
```diff
@@ -15,28 +17,41 @@ def field(field_name, options = {})
     # options[:read_only] must be true if form_mode is show
     # it may optionally be true if specified by the user
     # else it is false
-    options[:read_only] ||= @template.read_only
+    if form_mode == :show
+      options[:read_only] = read_only?
+    else
+      options[:read_only] ||= read_only? if options[:read_only].nil?
```
hooverlunch: Shouldn't this just be `options[:read_only] = read_only? if options[:read_only].nil?`. If the form_mode is :show we still want to obey the field-level read_only option if it is given, right?

spec/models/ability/response_ability_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+
+describe Response do
+  let(:user) { create(:user, role_name: "observer") }
+
+  subject(:ability) do
+    Ability.new(user: user, mission: get_mission)
+  end
+
+  let(:response) { create(:response, form: create_form, answer_values: create_answers) }
+  let(:own_response) { create(:response, form: create_form, answer_values: create_answers, user: user) }
```
hooverlunch: Is there a reason not to just do a `let(:form) { ...` here? Using `create_form` in this way will create two separate forms I think..?

spec/models/ability/response_ability_spec.rb
```diff
@@ -0,0 +1,71 @@
+require "spec_helper"
+
+describe Response do
```
hooverlunch: Nice to add this spec!

PR #254, opened 2016-08-19
852 new broadcast workflow

hooverlunch:


app/controllers/broadcasts_controller.rb
```diff
@@ -113,11 +106,14 @@ def possible_recipients
 
   private
 
-    def set_medium_options
-      @medium_options = configatron.to_h[:outgoing_sms_adapter] ? Broadcast::MEDIUM_OPTIONS : Broadcast::MEDIUM_OPTIONS_WITHOUT_SMS
-    end
+  def prep_form_vars
+    @medium_options = configatron.to_h[:outgoing_sms_adapter] ?
```
meltheadorable: Is there a reason to use `to_h` here instead of `has_key?` ?

hooverlunch: I think so. Configatron might have improved but I remember fighting with it like crazy trying to check for nils. has_key? won't work I think because outgoing_sms_adapter might be nil.

meltheadorable:
In the broadcast index, if you select "All users in mission" or "All observers in mission" the count shows as `0 Users` on the broadcasts index instead of the actual recipient count, it probably makes sense to calculate and store that count on the model when it's sent, since the number of users in these virtual groupings could change from the point when the broadcast was originally sent to them.

Otherwise, I think this looks good, though I'm unclear on the exact behavior when you combine "Select all" with a search, I understand at present that what it does is sets `specific_users` and then enumerates them, but I think we might want to set another type for "searched users" as the combination of the two things could potentially be quite large and as an end user it might be less confusing to see something like "All searched users" or "Search: <search string>" than a specific list of names.

hooverlunch:
Great catches here. I'll fix those two things and merge.

PR #257, opened 2016-08-23
4648 broadcast groups

hooverlunch:


meltheadorable:
Everything with this feature specifically looks great! Really excellent work, I don't need any changes to what you have here.

I did notice a couple of bugs though that might have been already happening in one of the other refactors that touched these systems:

1) The SMS test console fails when sending a forward because it tries to send it out through the real outgoing SMS adapter -- i think we need to pass the test adapter through to the mass messenger somehow

2) The broadcasts index shows SMS forwards but doesn't distinguish them in any way, we should either show broadcasts only or add a column to show whether they are a forward or a manual broadcast

If you wanna file separate bugs for these we can address them separately, but this is the first issue in a while that has touched forwards & broadcasts simultaneously and so I hadn't noticed them until now and they should still be fixed before we release. We can discuss the best approach to tackling these.

PR #270, opened 2016-09-08
5083: Fixes parser error for SMS

meltheadorable:


app/models/sms/decoder.rb
```diff
@@ -229,7 +229,7 @@ def add_answer
         idxs = split_options.map do |l|
 
           # make sure it's a letter
-          if l =~ /[a-z]/
+          if l =~ /\A[a-z]\z/i
```
hooverlunch: AHA! Nice work.

PR #271, opened 2016-09-09
5096: Removes IntelliSMS

meltheadorable:


db/migrate/20160909161815_remove_intelli_sms_default.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveIntelliSmsDefault < ActiveRecord::Migration
+  def up
+    Setting.where(default_outgoing_sms_adapter: "IntelliSms").update_all(default_outgoing_sms_adapter: nil)
```
hooverlunch: Please prefer raw SQL wherever possible in migrations. Relying on AR models can break if the models change in the future.

spec/features/sms_settings_form_spec.rb
```diff
@@ -8,37 +8,36 @@
...
       end
 
       scenario "filling in both should work" do
         visit("/en/m/#{mission.compact_name}/settings")
-        fill_in("setting_intellisms_username", with: "abc")
-        fill_in("setting_intellisms_password1", with: "jfjfjfjf")
-        fill_in("setting_intellisms_password2", with: "jfjfjfjf")
+        fill_in("setting_twilio_account_sid", with: "abc")
+        click_link("Change Auth Token")
+        fill_in("setting_twilio_auth_token1", with: "jfjfjfjf")
```
hooverlunch: Nice refactor of this test!

hooverlunch:
Looks like a bunch of tests are failing because default adapter is not recognized. Can you fix?

Other than that I think it looks fantastic!

PR #276, opened 2016-11-03
use dir=rtl on Arabic pages

mapmeld:
Last year before there was an Arabic locale, I added some CSS which supports right-to-left language layout: #131 

This would add a field to the locale file, allowing Arabic and other languages to set the dir attribute on the top-level HTML element.

Default value is dir=ltr

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -1,7 +1,7 @@
 <% body = yield %>
 <% ctlr_action_name = "#{route_key}_#{controller.action_name.gsub('update', 'edit').gsub('create', 'new')}" %>
 <!DOCTYPE html>
-<html>
+<html dir="<%= t('locale_dir', default: 'ltr') %>">
 <head>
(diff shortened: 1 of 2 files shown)


hooverlunch:
How does this actually look when enabled? We had looked at RTL support earlier but found we would need to do more work than this to add reasonable support. Namely:

* Install bootstrap RTL support (https://github.com/morteza/bootstrap-rtl) 
* Audit stylesheets for left/right specific things and use https://github.com/tysonmatanich/directional-scss to make any left/right references dynamic

Specific things that will need attention:

* Header/footer/navbar
* Dashboard
* Forms
* Index tables
* Form drag and drop
* Tags sometimes getting squat
* Printable form (labels on correct side of boxes)
* SMS guide (just a few right/left issues)
* RTF editor
* Location lookup modal (small issues)

Then do run-through of regression test steps, watching out for small issues.

If you think it's decent without all this stuff and want to move ahead, can you please add a local config setting (defaulting to false) that enables RTL support. Add a comment saying RTL support is minimal at the moment.

Thoughts?

PR #277, opened 2016-11-19
4821 sms form translation

meltheadorable:


app/views/forms/sms_guide.html.erb
```diff
@@ -2,65 +2,64 @@
...
 <% content_for(:alerts) { appendix_alert } %>
 
-<% locale_was = I18n.locale; I18n.locale = @lang %>
 <div class="sms_guide">
 
   <div class="controls">
     <form>
       <input type="hidden" name="sms_guide" value="1">
-      <%= t("common.language") %>:
+      <%= t("common.language", locale: @lang) %>:
```
meltheadorable: I will change the name of the instance variable (I was just re-using the one already here), but I think the additional helper would be a bad move despite the current repetition -- relying on the existence of an instance variable like that makes for some pretty opaque and tight coupling between the helpers and specific variables set in the controller, I'd rather have the extra keystrokes on these calls than increase the cognitive overhead for the next dev working on this stuff (whether that be me, in the future, or somebody else). 

I'm open to finding another solution to the repetition here, it's a bit of a code smell, but we already have this issue where some other helpers are relying on sort of hidden information and it makes the execution path harder to follow, I don't want to compound that right now.

PR #278, opened 2016-12-01
5505 sms wo mission

hooverlunch:


app/models/sms/decoder.rb
```diff
@@ -86,6 +86,21 @@ def find_form
     # if version not found, raise error
     raise_decoding_error("form_not_found", form_code: code) unless v
 
+    if @msg.mission
+      # if we already know the mission (it may or may not be already stored on the message)
+      # and it doesn't match the form's mission, complain
+      if v.form.mission != @msg.mission
+        raise_decoding_error("wrong_mission", form_code: code)
```
meltheadorable: Should we be giving away this information? Seems like saying that the form code is valid in another mission is a data leak. I'd prefer just to give whatever invalid message we would usually give if the form code does not exist in the mission.

app/models/sms/decoder.rb
```diff
@@ -86,6 +86,21 @@ def find_form
...
+      end
+    else
+      # If the mission is not stored on the message, set it based on form.
+      # This is allowed due to situations where multiple missions may want to use the same phone
+      # number or same gateway provider that doesn't support different submit URLs
+      # based on incoming phone number.
+      @msg.mission = v.form.mission
```
meltheadorable: Is there a way to, trivially, make this contingent on a setting somewhere enabling this behavior? This is definitely a security risk and I'd prefer this behavior be opt-in on a mission-by-mission basis -- just some kind of setting like "allow responses via universal SMS url"

Maybe just something in the backlog to handle this as a TODO since this is critical right now, but I think that's a reasonable enhancement.

app/models/sms/decoder.rb
hooverlunch: Yeah good point. I'll remove.

app/models/sms/decoder.rb
hooverlunch: Yeah good idea, I'll add to backlog

PR #279, opened 2016-12-01
5505: Adds global SMS key

meltheadorable:


app/controllers/settings_controller.rb
```diff
@@ -35,7 +35,11 @@ def regenerate_incoming_sms_token
   end
 
   def using_incoming_sms_token_message
-    url = mission_sms_submission_url(@setting.incoming_sms_token, locale: nil)
+    if params[:missionless].present?
+      url = missionless_sms_submission_url(@setting.universal_sms_token, locale: nil, mission_name: nil, mode: nil)
+    else
+      url = mission_sms_submission_url(@setting.incoming_sms_token, locale: nil)
+    end
```
hooverlunch: Ah, nice that you included this modal stuff!

app/controllers/sms_controller.rb
```diff
@@ -40,7 +40,7 @@ def create
     end
 
     # Mission is guaranteed to be set by this point.
-    if params[:token] != current_mission.setting.incoming_sms_token
+    unless [current_mission.setting.incoming_sms_token, configatron.universal_sms_token].include? params[:token]
```
hooverlunch: What if setting.incoming_sms_token is nil and params[:token] is nil?. Perhaps add a compact?

config/initializers/local_config.rb.example
```diff
@@ -37,6 +37,7 @@ ActionMailer::Base.smtp_settings = {
 # ** FrontlineCloud **
 # configatron.default_settings.outgoing_sms_adapter = "FrontlineCloud"
 # configatron.default_settings.frontlinecloud_api_key = "xxx"
+# configatron.universal_sms_token = "50b488c71be4e9cd0ba34a7bf0b7053f"
```
hooverlunch: This looks like it's part of the frontline cloud settings. Perhaps put in a separate paragraph and add a comment?

config/locales/en.yml
```diff
@@ -546,6 +547,8 @@ en:
         incoming_sms_numbers_str: "Enter the phone number(s) (including country code, separated by commas) to which incoming SMSes for SMS forms should be sent. Only needed if using SMS submissions."
         incoming_sms_token: |-
           This token is included in the URL used by the incoming SMS provider to prevent the submission of unauthorized messages.
+        universal_sms_token: |-
+          This token is included in the URL used by the incoming SMS provider to prevent the submission of unauthorized messages. Use this URL if you need to allow multiple missions to submit via the same URL.
```
hooverlunch: Nice.

spec/support/incoming_sms_support.rb
```diff
@@ -53,10 +53,11 @@ def do_incoming_request(params)
     req_env = {}
 
     url_prefix = defined?(missionless_url) && missionless_url ? "" : "/m/#{get_mission.compact_name}"
+    url_token = defined?(missionless_url) && missionless_url ? configatron.universal_sms_token : get_mission.setting.incoming_sms_token
 
     params[:sent_at] ||= Time.now
     params[:incoming][:adapter] ||= "TwilioSms"
-    params[:url] ||= "#{url_prefix}/sms/submit/#{get_mission.setting.incoming_sms_token}"
+    params[:url] ||= "#{url_prefix}/sms/submit/#{url_token}"
```
hooverlunch: I think we should have a quick spec that checks that two different missions (i.e. forms from two different missions) can submit using the missionless URL. Not checking this was why I missed this issue in the first place. Could you add?

app/controllers/sms_controller.rb
meltheadorable: Good thinking.

spec/support/incoming_sms_support.rb
meltheadorable: Yep.

PR #282, opened 2016-12-10
5253 print logo override

meltheadorable:


config/deploy.rb
```diff
@@ -107,10 +107,12 @@
   task :theme_config, role: :app do
     theme = fetch(:theme, "app/assets/stylesheets/all/variables/_default_theme.scss")
     logo = fetch(:logo, "app/assets/images/logo.png")
+    logo_dark = fetch(:logo_dark, "app/assets/images/logo-dark.png")
     run "mkdir -p #{shared_path}/app/assets/stylesheets/all/variables/"
     run "mkdir -p #{shared_path}/app/assets/images/"
     top.upload theme, "#{shared_path}/app/assets/stylesheets/all/variables/_theme.scss"
     top.upload logo, "#{shared_path}/app/assets/images/logo-override.png"
+    top.upload logo_dark, "#{shared_path}/app/assets/images/logo-dark-override.png"
```
hooverlunch: Now that the deploy config is in a different repo, we can either just copy the same folder structure over or put these three files in a simpler, gitignored subdir called `theme` or something else... Not sure what is best.

PR #284, opened 2016-12-12
5354: Fixed broken birth year import

meltheadorable:


app/views/users/form.html.erb
```diff
@@ -12,7 +12,8 @@
   <%= f.field(:email) %>
   <%= f.field(:phone) %>
   <%= f.field(:phone2) %>
-  <%= f.field(:birth_year, type: :select, options: birth_year_options) %>
+  <%= f.field(:birth_year, type: :select, options: birth_year_options,
+        read_only_content: @user.birth_year.present? ? @user.birth_year.to_s : nil ) %>
```
hooverlunch: `birth_year.try(:to_s)`?

meltheadorable: oh gosh, i use `try` all the time, too -- brain fuzz!

PR #286, opened 2017-01-16
5582: Changes to translatable, fixes schema

meltheadorable:


app/models/concerns/translatable.rb
```diff
@@ -215,4 +222,20 @@ def available_locales(options = {})
 
     locales
   end
-end
\ No newline at end of file
+end
+module Translatable
+  class TranslatableLengthValidator < ActiveModel::Validations::LengthValidator
+
+    def tokenize(value)
```
hooverlunch: I don't really understand what this method is for. I trust it's correct, but can you put a comment so future visitors are not as confused as me :)

PR #288, opened 2017-02-18
5757: Locale URLs for ODK

meltheadorable:


spec/controllers/odk_form_download_spec.rb
```diff
@@ -18,54 +18,58 @@
       @form3.publish!
     end
 
-    describe 'listing forms' do
-      it 'should succeed' do
-        get_s(odk_form_list_path)
+    describe "listing forms" do
+      it "should succeed" do
+        get_s(odk_form_list_path, format: :xml)
```
hooverlunch: Should we perhaps hard-code the paths in this spec instead of using the helpers? At first glance I assumed this would construct a path that ends in `.xml`, which nobody actually uses. This may not be true but I can't be sure and it seems that we should be testing the real paths publish in our documentation instead of what the helper generates?

spec/controllers/odk_form_download_spec.rb
```diff
@@ -74,21 +78,21 @@
...
     before do
       @mission.update_attributes!(locked: true)
       @form1 = create(:form, mission: @mission, question_types: %w(integer integer))
       @form1.publish!
     end
 
-    it 'listing forms should return 403' do
-      get(odk_form_list_path)
+    it "listing forms should return 403" do
+      get(odk_form_list_path, format: :xml)
```
hooverlunch: Same thing here. I'm not at all convinced this isn't adding a .xml...

config/routes.rb
```diff
@@ -100,7 +100,7 @@
   scope ":locale/:mode(/:mission_name)", locale: /[a-z]{2}/, mode: /m|admin/, mission_name: /[a-z][a-z0-9]*/ do
 
     # the rest of these routes can have admin mode or not
-    resources :forms do
+    resources :forms, constraints: lambda { |req| req.format == :html } do
```
hooverlunch: Can we standardize on the `->` notation?

spec/controllers/odk_form_download_spec.rb
meltheadorable: It's not adding a `.xml`, it's setting content type and/or accepts headers -- if we hardcoded the paths we'd still need the `format: :xml` there

hooverlunch: Oh oh oh! Nice. My bad. Maybe we should hard code them anyway?

PR #292, opened 2017-02-23
3594 option reference corruption

hooverlunch:


app/models/condition.rb
```diff
@@ -30,41 +29,26 @@ class Condition < ActiveRecord::Base
...
-  end
+  replicable backward_assocs: [:questioning, :ref_qing, {name: :option_node, skip_obj_if_missing: true}],
+    dont_copy: [:ref_qing_id, :questioning_id, :option_node_id]
 
-  def option
-    options.try(:first)
+  # We accept a list of OptionNode IDs as a way to set the option_node association.
+  # This is useful for forms, etc. We just pluck the last non-blank ID off the end.
+  # If all are blank, we set the association to nil.
+  def option_node_ids=(ids)
```
meltheadorable: Neat!

db/migrate/20170109183447_populate_condition_option_node_reference.rb
```diff
@@ -0,0 +1,61 @@
...
+              sequence: condition.ref_qing.option_set.descendants.maximum(:sequence)
+            )
+          elsif node.try(:option_id) != option.id
+            puts "Condition ##{condition.id}: OptionNode ##{node.try(:id)} didn't match Option ##{option.id}"
+            check_failed = true
+          end
+        end
+      end
+
+      raise "OptionNode <=> Option check failed" if check_failed
```
meltheadorable: What would happen if this error was raised in, say, a deploy? What would the resolution steps look like? Just trying to understand the implications of a failure to map this here.

db/schema.rb
```diff
@@ -144,9 +136,9 @@
     t.integer "form_id", limit: 4, null: false
     t.text "group_hint_translations", limit: 65535
     t.text "group_name_translations", limit: 65535
-    t.integer "group_rank", limit: 4
```
meltheadorable: Any idea what this was used for and why it isn't anymore? Appreciate the cleanup, just making sure there's no chance this is called by some metaprogramming somewhere that makes it hard to search.

spec/features/response_condition_spec.rb
```diff
@@ -4,7 +4,21 @@
   before do
     @user = create(:user)
     @form = create(:form, name: 'Foo',
-      question_types: %w(long_text text integer text decimal select_one multilevel_select_one multilevel_select_one select_multiple datetime date time text))
+      question_types: [
+        "long_text",             #  0
```
meltheadorable: These little hints are great, eventually I'd like to try doing this with maybe a hash (using `.keys` to get the array for the form builder), or try to find some other way to make these sorts of specs a little more human-readable but the comment hints are a really great start.

spec/models/condition_spec.rb
```diff
@@ -69,227 +49,194 @@
     # expect(c.to_odk).to eq("not(selected(/data/q#{q.previous[0].question.id}, '#{c.option_id}'))")
     # q, c = build_condition(:question_types => %w(datetime integer), :op => 'neq', :value => '2013-04-30 2:14pm')
     # expect(c.to_odk).to eq("format-date(/data/q#{q.previous[0].question.id}, '%Y%m%d%H%M') != '201304301414'")
+    let(:qing) { form.questionings.first }
```
meltheadorable: hooray for `let`! 💯

db/migrate/20170109183447_populate_condition_option_node_reference.rb
hooverlunch: It's designed to fail loudly in a deploy. We would go in and try to fix things manually as quickly as possible. I have run the production DB through the migration and it works smoothly (after some tweaking).

db/schema.rb
hooverlunch: I honestly can't remember and I didn't mean to remove it. Can you put it back in please. This must have happened by accident when I was playing around with the production DB. Thank you for catching!

PR #295, opened 2017-03-01
5758: Remember context

meltheadorable:


app/controllers/concerns/application_controller/crud.rb
```diff
@@ -45,30 +45,46 @@ def set_success_and_redirect(obj, options = {})
...
-      session[:last_page_numbers][last_page_number_hash_key] = params[:page]
-    end
+  def remember_context
+    remember_search(params[:search]) if params[:search]
+    remember_page(params[:page]) if params[:page]
+  end
+
+  def remember_search(search)
+    session[:last_search] ||= {}
+    session[:last_search][last_context_hash_key] = search
```
hooverlunch: Shouldn't it be `last_searches` b/c there are multiple of them and to be consistent with `last_page_numbers`?

app/controllers/concerns/application_controller/crud.rb
```diff
@@ -45,30 +45,46 @@ def set_success_and_redirect(obj, options = {})
...
-    controller_name + current_mission.try(:id).to_s
+  # builds a simple hash key for remembering page context
+  def last_context_hash_key
+    "#{controller_name}#{current_mission.try(:id)}"
+  end
+
+  def get_last_context
+    page = get_last_page_number if session[:last_page_numbers]
+    search = get_last_search if session[:last_search]
+    { page: page, search: search }
```
hooverlunch: Can we do no spaces around curlies for hashes, only for procs?

app/controllers/concerns/application_controller/crud.rb
```diff
@@ -45,30 +45,46 @@ def set_success_and_redirect(obj, options = {})
...
-    if session[:last_page_numbers]
-      session[:last_page_numbers][last_page_number_hash_key]
-    else
-      nil
-    end
+    session[:last_page_numbers][last_context_hash_key].presence
+  end
+
+  def get_last_search
+    session[:last_search][last_context_hash_key].presence
```
hooverlunch: Using `.presence` here can still end up causing a nil error if `session[:last_search]` is nil. You could do `session[:last_search].try(:[], last_context_hash_key)` though?

Also above.

app/controllers/users_controller.rb
```diff
@@ -157,7 +157,7 @@ def handle_printable_instructions
       flash[:password] = @user.password
       redirect_to(action: :login_instructions, id: @user.id)
     else
-      redirect_to(index_url_with_page_num)
+      redirect_to(index_url_with_context)
```
hooverlunch: Howz about a quick happy path feature spec? remember_context_spec.rb:

1. Create N+2 responses where N is the per_page for responses. All submitted by the same user.
1. Go to response index
1. search by submitter:Your Fake user
1. click page 2
1. delete first response (don't forget to click confirm)
1. should be redirected back to page 2 with search intact

hooverlunch:
BTW here is the spec failure:

```
  1) Sms::Decoder authentication with SMS Authentication enabled should lock account after 3 failed attempts
     Failure/Error: expect(error.type).to eq(options[:error])
       
       expected: "account_locked"
            got: "user_not_found"
       
       (compared using ==)
     # ./spec/models/sms/sms_decoder_spec.rb:631:in `assert_decoding_fail'
     # ./spec/models/sms/sms_decoder_spec.rb:27:in `block (4 levels) in <top (required)>'
```

Flapper?

meltheadorable:
Must be a flapper, since the PR build passed.

app/controllers/concerns/application_controller/crud.rb
meltheadorable: Those are only called in

```ruby
def get_last_context
    page = get_last_page_number if session[:last_page_numbers]
    search = get_last_searches if session[:last_searches]
    {page: page, search: search}
  end
```

so `last_searches` should never be `nil`

PR #296, opened 2017-03-06
5868 add reviewer field

DevneyHamilton:


DevneyHamilton:
This PR includes small updates to the dev enviro setup instructions.

app/helpers/responses_helper.rb
```diff
@@ -26,6 +26,7 @@ def format_responses_field(resp, field)
       when "incomplete" then tbool(resp.incomplete?)
       when "reviewed" then reviewed_status(resp)
       when "user_id" then can?(:read, resp.user) ? link_to(resp.user.name, resp.user) : resp.user.name
+      when "reviewer_id" then resp.reviewer.present? ? (can?(:read, resp.reviewer) ? link_to(resp.reviewer.name, resp.reviewer) : resp.reviewer.name) : ""
```
meltheadorable: I think we should maybe split this out so that it's not one line -- chaining the ternary operator like this can be a little bit confusing. The parenthesis help but I think it'd be better for this to take up a couple of lines.

spec/models/report/answer_tally_report_spec.rb
```diff
@@ -151,7 +151,7 @@
       expect(report).to have_data_grid(nil)
     end
 
-    it "counts of options across a select one question and select multiple question" do
+    it "counts of options across a select one question and select multiple question", :investigate do
```
meltheadorable: Thanks for doing this! We will be keeping a good eye on these moving forward and eventually start fixing them.

Gemfile
```diff
@@ -32,6 +32,7 @@ gem "therubyracer", "~> 0.12.2", platforms: :ruby
 gem "draper", "~> 2.1.0"
 gem "attribute_normalizer", "~> 1.2.0"
 gem "responders", "~> 2.3.0"
+gem "thor", "0.19.1"
```
meltheadorable: Maybe we should add a comment to this explaining that newer versions of thor produce command line argument errors, so this is a temporary workaround. We will want to remove this constraint once either rails or thor fixes the issue.

PR #299, opened 2017-03-14
5866 opt set modal header

DevneyHamilton:


DevneyHamilton:
I'm sorry, a few front end tests are failing so I'm looking into that.

app/assets/javascripts/legacy/views/question_form.js
```diff
@@ -106,16 +106,23 @@
     var question_type_param = '?adding_to_question_type=' + self.field_value('qtype_name'),
         loadUrl = self.params.new_option_set_path + question_type_param;
 
+    var modal_header = 'option_set.create_for_question';
+    var question_code = $('#question_code').val().trim();
+    if (question_code === '') {
+      modal_header = 'option_set.create_for_new_question';
+    }
+    $('#create-option-set .modal-title').html(I18n.t(modal_header, {code: question_code}));
```
hooverlunch: Very nice.

hooverlunch: One thing: can you please change to `.text` instead of `.html` so that we get XSS protection.

rgalusca:
I found a bug on this branch. I cannot open the "Create New Option Set" modal. @DevneyHamilton, can you please check what is going on on Monday morning? Thank you!

DevneyHamilton:
@rgalusca, Can you please give me a screenshot of where you can’t open the option set modal? If I'm creating a new question, or editing an existing question, and have 'Select One' or 'Select Multiple,' I have the orange "+ Create New Option Set" link (in screen shot below) and it opens up the modal. If I create a new option set from the main menu, instead of from creating a question, then it takes me to a page to create the option set.
screen shot 2017-03-27 at 10 31 18 am

DevneyHamilton:
Oh actually, let me push the rebased branch. . . .

DevneyHamilton:
@rgalusca I have updated. On my local it's working when I make a new question from editing a form or creating a new form as well as from creating a new question. It turns out there is different HTML for the new question form in these different parts of the application, something to keep in mind for future chances to move towards React and/or refactoring.

PR #301, opened 2017-03-15
4957 external sql

hooverlunch:
The spec failure is a very strange flapper. Runs fine individually.

### DIFF
--- a/app/assets/stylesheets/all/components/_settings.css.scss
+++ b/app/assets/stylesheets/all/components/_settings.css.scss
@@ -1,0 +1,21 @@
+form.setting_form {
+  .adapter_settings .credential_fields {
+    display: none;
+  }
+
+  #external_sql .control {
+    pre {
+      width: 415px;
+      height: 100px;
+      overflow: auto;
+      border: 1px solid #ccc;
+      margin: 0 0 3px 0;
+      font-size: 8pt;
+      white-space: normal;
+    }
+
+    a {
+      font-size: 9pt;
+    }
+  }
+}
--- a/app/models/results/sql_generator.rb
+++ b/app/models/results/sql_generator.rb
@@ -6,10 +6,11 @@ class Results::SqlGenerator
   # We accept a mission parameter here for scoping. Obviously, anybody able to run queries on the
   # DB has at least read access to all data, so this is not a security measure, just a convenience one.
   def initialize(mission)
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/all/components/_settings.css.scss
```diff
@@ -0,0 +1,21 @@
+form.setting_form {
```
meltheadorable: YAY! Love seeing the CSS getting broken out more as we move forward.

app/models/results/sql_generator.rb
```diff
@@ -0,0 +1,42 @@
...
+      select("answers.time_value AS time_value").
+      select("answers.latitude AS latitude").
+      select("answers.longitude AS longitude").
+      select("IFNULL(ao.canonical_name, co.canonical_name) AS choice_name").
+      select("option_sets.name AS option_set").
+      joins(Results::Join.list_to_sql(:users, :forms, :answers, :questionings,
+        :questions, :option_sets, :options, :choices)).
+      where("responses.mission_id = ?", mission.id).
+      order("responses.created_at DESC").
+      to_sql
```
meltheadorable: Just wondering if there is any benefit to letting this query stay an activerecord relation and calling `to_sql` when we need SQL specifically. I guess if that need presents itself we could always refactor later, I was just trying to think through whether there's anywhere else where we're producing a complete result set like this where we would benefit from being able to re-use the relation.

spec/models/sms/sms_decoder_spec.rb
```diff
@@ -6,25 +6,25 @@
 
   describe "authentication" do
     context "with SMS Authentication enabled" do
-      before(:all) { create_form(questions: %w(integer), authenticate_sms: true) }
+      let(:form) { create_form(questions: %w(integer), authenticate_sms: true) }
```
meltheadorable: Thanks for making these changes, fingers crossed that we stop getting these spec errors on the SMS decoding! 🤞

app/models/results/sql_generator.rb
hooverlunch: Yeah probably not. Originally I think the mission scope was coming in this way, but not sure why, just trying to be cute I guess :o

meltheadorable: That's what I thought, just checking 😄

PR #302, opened 2017-03-21
5948 mission in print instructions

DevneyHamilton:
Asked Roxana about what to display. She said only 'name' (not compact name) until Dottie asks for something else.


meltheadorable:
You can't produce the ODK url without the compact name, so I really think we should display both if the purpose of this feature is, as I understand it, for observers with no access to the web interface trying to set up ELMO on their mobile devices.

PR #303, opened 2017-03-21
5961 dont notify admin for 404

DevneyHamilton:
This will not send emails to admins for record not found. Tested manually that without changes, an email is sent on an ActiveRecord::RecordNotFound error and with changes, the email is not sent. Instead it just renders nothing with status 404. It also has a rescue block for InvalidAuthenticityToken but this is not tested. We'll see if the emails stop in production.


DevneyHamilton:
To Mel: I ended up just raising the error instead of rendering nothing. Right now, raising the error should maintain the existing user-facing behavior. Having these separate handlers 1) skips the annoying email and 2) makes it easy to update behavior in the future. If you think it's better to collapse the two new handler functions into one function since they are identical, I can do that very quickly.

PR #305, opened 2017-03-28
5916: Test and update to sanitize response answer values when showing…

DevneyHamilton:
… a response.


DevneyHamilton:
Note: I'm nervous about the fact that we mark a string with user input as 'html_safe' in excerpt_to_html (line 39 of [answers_helper.rb | https://github.com/thecartercenter/elmo/blob/develop/app/helpers/answers_helper.rb]). Do we need to mark it html_safe there at all?

PR #308, opened 2017-04-03
6152: Fixed login instructions bug

meltheadorable:


DevneyHamilton:
After chat w/ Tom: printable instructions should not be available in admin mode at all. I'll address that.

PR #310, opened 2017-04-05
3016 repeat group csv

DevneyHamilton:


spec/expectations/response_csv/repeat_groups.csv
```diff
@@ -0,0 +1,11 @@
+Form,Submitter,DateSubmitted,ResponseID,Question1,Question2,Question3,Question4,Question5,Question6,RepeatLevel
```
DevneyHamilton: I updated these to have hardcoded ids. Previously this file was relying on FactoryGirl's sequence, which is messy with more than one test, and declared an anti-pattern on a couple blogs.

spec/models/response_csv_spec.rb
```diff
@@ -63,8 +63,68 @@
...
+      expected = File.read(File.expand_path('../../expectations/response_csv/responses.csv', __FILE__))
       expect(ResponseCSV.new(responses).to_s).to eq expected
     end
   end
-end
\ No newline at end of file
+
+  context "with repeat groups" do
+    let(:repeat_form) do
+      create(:form, question_types: ["integer", [:repeating, "text", "integer", "select_multiple"], "integer", [:repeating, "text", "geo_multilevel_select_one",  "integer"]]).tap do |f|
```
meltheadorable: this is interesting, but it looks so much like a hash that i'm wondering if this would be a better API for this:

`create(:form, question_types: ["integer", {repeating: ["text", "integer", "select_multiple"]}, "integer", {repeating: ["text", "geo_multilevel_select_one",  "integer"]}])`

Benefits include that you can just do a simple `hash[:repeating]` to get the question types rather than trying to delete the magic from the array. I think it also delineates the boundary a little better. This is a really good way of eliminating our dependency on plucking the specific group manually out of the code by its position in the array to apply the transformation on its attributes.

DevneyHamilton:
I wonder if we should wait to merge this until after the critical path changes for UUIDs are merged. That way if there are conflicts, I can fix them without blocking  the critical path.

DevneyHamilton:
Update: when UUID work is merged in I will rebase, resolve conflicts and merge.

PR #311, opened 2017-04-13
6237: Adds UUIDs to EVERYTHING

meltheadorable:
🍷🎉👩‍💻


spec/models/user_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'spec_helper'
 
 describe User do
+  it_behaves_like "has a uuid"
```
hooverlunch: I'm not sure about adding this to every spec. Seems a bit redundant to test behavior of a parent class on all the subclasses. I'd rather have an ApplicationRecord spec that tests it once. I will make this change since you're out.

db/migrate/20170413020922_add_uuid_fields_to_tables.rb
```diff
@@ -0,0 +1,14 @@
+class AddUuidFieldsToTables < ActiveRecord::Migration
+  TABLES_FOR_MIGRATION = [:answers, :assignments, :broadcast_addressings, :broadcasts, :choices, :conditions,
+    :form_forwardings, :form_items, :form_versions, :forms, :media_objects, :missions, :operations,
+    :option_nodes, :option_sets, :options, :questions, :report_calculations, :report_option_set_choices,
+    :report_reports, :responses, :settings, :sms_messages, :taggings, :tags,
+    :user_group_assignments, :user_groups, :users, :whitelistings]
+
+  def change
+    TABLES_FOR_MIGRATION.each do |table_name|
+      add_column table_name, :uuid, :string, null: false
```
hooverlunch: I have a feeling this won't work if there is existing data. I can fix it to add this after the next migration runs.

hooverlunch: Interesting -- seems to have worked in MySQL but I know it doesn't work in Postgres...

hooverlunch:
Another thing I had to change was to use update_column (no callbacks) instead of save! because save was triggering sphinx index stuff on every Answer save!

This is why it's good to have a decent amount of data in your DB when testing this stuff..

db/migrate/20170413020922_add_uuid_fields_to_tables.rb
meltheadorable: I did test this on existing data, though not enough of it to catch the other problem you mentioned with sphinx.

spec/models/user_spec.rb
meltheadorable: This was really useful during development because it helped me catch cases where individual subclasses were treating it differently for one reason or another, and also caught one case where the column had not actually been added to the class in question. Like, `Setting` was filtering out attributes in admin mode so the uuid was getting stripped before save. I never would have caught that with a single spec and we'd never catch regressions like that if they're removed, which is why I went for the shared example group. We can remove this now that we do know it's working if you want, this special case shouldn't be around long enough for us to introduce regressions once the uuid becomes the primary key, but it was worthwhile to have it there.

hooverlunch: Interesting. I will leave it in and add a comment explaining why we're diverging from the norm in this case.

PR #314, opened 2017-04-21
5920: Expand password keyspace, refactor

meltheadorable:


meltheadorable:
I don't think we have any minimum requirements for user-selected passwords and I think a goal of this functionality was to improve the generation of new passwords without changing the requirements for existing ones in a way that might invalidate them. We can talk about also setting more stringent requirements for user-selected passwords but I think that would be its own issue.

DevneyHamilton:
okay, that makes sense.

PR #319, opened 2017-04-27
5992 convert to postgres

hooverlunch:


app/models/question.rb
```diff
@@ -36,29 +36,56 @@ class Question < ApplicationRecord
...
-      LEFT OUTER JOIN forms ON forms.id = questionings.form_id
-      LEFT OUTER JOIN answers ON answers.questioning_id = questionings.id
-      LEFT OUTER JOIN questions copies ON questions.is_standard = 1 AND questions.id = copies.original_id
-      LEFT OUTER JOIN form_items copy_questionings ON copy_questionings.question_id = copies.id AND copy_questionings.type = 'Questioning'
-      LEFT OUTER JOIN forms copy_forms ON copy_forms.id = copy_questionings.form_id
-      LEFT OUTER JOIN answers copy_answers ON copy_answers.questioning_id = copy_questionings.id
-    }).group('questions.id') })
+  # Fetches association counts along with the questions.
+  # Accounts for copies with standard questions.
+  # - form_published_col contains true if any associated forms are published, false or nil otherwise
```
DevneyHamilton: I'm really curious what this scope is for?

spec/models/sms/sms_decoder_spec.rb
```diff
@@ -351,7 +351,7 @@
       assert_decoding(body: "#{@form.code} 1.15 2.11,12,13,14,15,20,1a", answers: [15, %w(1 2 3 4 5 36 10)])
     end
 
-    it "decimal question should work", :investigate do
+    it "decimal question should work" do
```
DevneyHamilton: The :investigate tag was for flapping specs that usually pass but not always. Do these always pass now?

docs/production-setup.md
```diff
@@ -156,6 +150,25 @@ When new versions of ELMO are released, you will want to upgrade. To do so, ssh
 
 Then load the site in your browser. You should see the new version number in the page footer.
 
+### Converting an Existing Instance from MySQL to PostgreSQL
+
+1. Install PostgreSQL (see above).
+1. First upgrade to version v5.99 (`git checkout v5.99`), which is the last MySQL compatible version.
```
meltheadorable: We'll need to use the right version number here when we have one. Maybe we can use `release-5.X` and keep that up to date with the latest version + patch?

hooverlunch:
The solution to searching translatable fields was converting them to JSONB and using `->>`. [So awesome](https://github.com/thecartercenter/elmo/blob/5992_convert_to_postgres/app/models/option_suggester.rb#L10).

PR #321, opened 2017-05-05
6320 uuid shortcode in response csv

DevneyHamilton:
I put the expectations (that used to be in csv files) into a PORO, and pass in the uuids and shortcodes to be interpolated into the expectation string.


spec/expectations/response_csv_expectation_maker.rb
```diff
@@ -0,0 +1,26 @@
...
+Sample Form 1,A User 1,2015-11-20 06:30 CST,#{uuids[0]},#{shortcodes[0]},Fruit,1,1,Banana,2,Cat,2,,,,,,\r
+Sample Form 1,A User 1,2015-11-20 06:30 CST,#{uuids[0]},#{shortcodes[0]},Vegetable,1,1,,,,2,Asparagus,Ghana,Accra,5.55,0.2,3\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},,0,3,,,,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Xigua,10,Dog,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Yuzu,9,Cat;Dog,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Ugli,8,Cat,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Vegetable,1,3,,,,4,Zucchini,Canada,Calgary,51.045,-114.057222,7\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Vegetable,1,3,,,,4,Yam,Canada,Ottawa,45.429299,-75.629883,6\r\n")
+  end
+end
```
hooverlunch: Cute!

The general pattern for adding stuff like this is to add a file to spec/support with a module inside of it, and then add e.g. 

```
config.include ResponseCSVExpectations, type: :request
```

to spec_helper.rb. Then you would just call `make_expectation_with_repeat_groups` without instantiating a class.

spec/models/response_csv_spec.rb
```diff
@@ -63,7 +64,10 @@
 
     it "should generate correct CSV" do
       responses = Response.unscoped.with_associations.order(:created_at)
-      expected = File.read(File.expand_path('../../expectations/response_csv/responses.csv', __FILE__))
+      #expected = File.read(File.expand_path('../../expectations/response_csv/responses.csv', __FILE__))
```
hooverlunch: Remove plz.

spec/models/response_csv_spec.rb
```diff
@@ -77,12 +81,12 @@
...
-                  {repeating: {q_types: ["text", "integer", "select_multiple"], name: "Fruit"}},
-                  "integer",
-                  {repeating: {q_types: ["text", "geo_multilevel_select_one",  "integer"], name: "Vegetable"}}
-                ]).tap do |f|
+      question_types:
+      ["integer",
+        {repeating: {q_types: ["text", "integer", "select_multiple"], name: "Fruit"}},
+        "integer",
+        {repeating: {q_types: ["text", "geo_multilevel_select_one",  "integer"], name: "Vegetable"}}
+        ]).tap do |f|
```
hooverlunch: I like getting rid of the huge indents, but our typical style would be to have these one level further in than line 83.

spec/expectations/response_csv_expectation_maker.rb
hooverlunch: Probably would also make sense to tweak the method names a little like `response_csv_expectation_with_repeat_groups`.

hooverlunch: Perhaps to DRY things up a bit, you could also make these one method, pass `responses` instead of `uuids` and `shortcodes`, and move the `map` calls into the new unified method.

spec/models/response_csv_spec.rb
```diff
@@ -125,7 +129,9 @@
       end
 
       responses = Response.order(:id)
-      expected = File.read(File.expand_path('../../expectations/response_csv/repeat_groups.csv', __FILE__))
+      uuids = responses.map(&:uuid)
+      shortcodes = responses.map(&:shortcode)
+      expected = ResponseCSVExpectationMaker.make_expectation_with_repeat_groups(uuids, shortcodes)
```
hooverlunch: So this would become `expected = response_csv_expectation(responses, repeat_groups: true)`

The `repeat_groups` named param would default to false.

spec/models/response_csv_spec.rb
```diff
@@ -63,7 +64,10 @@
 
     it "should generate correct CSV" do
       responses = Response.unscoped.with_associations.order(:created_at)
-      expected = File.read(File.expand_path('../../expectations/response_csv/responses.csv', __FILE__))
+      #expected = File.read(File.expand_path('../../expectations/response_csv/responses.csv', __FILE__))
+      uuids = responses.map(&:uuid)
+      shortcodes = responses.map(&:shortcode)
+      expected = ResponseCSVExpectationMaker.make_expectation_without_repeat_groups(uuids,shortcodes)
```
hooverlunch: This could be `expected = response_csv_expectation(responses)`

spec/support/response_csv_expectations.rb
```diff
@@ -0,0 +1,30 @@
...
+Sample Form 1,A User 1,2015-11-20 06:30 CST,#{uuids[0]},#{shortcodes[0]},Fruit,1,1,Banana,2,Cat,2,,,,,,\r
+Sample Form 1,A User 1,2015-11-20 06:30 CST,#{uuids[0]},#{shortcodes[0]},Vegetable,1,1,,,,2,Asparagus,Ghana,Accra,5.55,0.2,3\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},,0,3,,,,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Xigua,10,Dog,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Yuzu,9,Cat;Dog,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Fruit,1,3,Ugli,8,Cat,4,,,,,,\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Vegetable,1,3,,,,4,Zucchini,Canada,Calgary,51.045,-114.057222,7\r
+Sample Form 1,A User 2,2015-11-20 06:30 CST,#{uuids[1]},#{shortcodes[1]},Vegetable,1,3,,,,4,Yam,Canada,Ottawa,45.429299,-75.629883,6\r\n")
+  end
+end
```
DevneyHamilton: I left ordering in the spec, since the ordering is different for each one. I changed these to take responses so extracting the uuids and shortcodes happens in the helper instead of the spec. I left these as separate functions because the logic is different (e.g. 1 line per response w/out repeat, but not w/ repeat). I didn't try to iterate of responses because making the csv file lines from the responses require reproducing a lot of the the logic this is trying to test.

PR #323, opened 2017-05-18
6036 remove default scopes

hooverlunch:
This one turned out to be pretty simple. Just one reviewer is enough I think.

I cancelled the last test build b/c it was for a very small commit. I guess I shouldn't have but I feel bad for poor Travis working so hard all the time...


hooverlunch:
This one turned out to be pretty simple. Just one reviewer is enough I think.

I cancelled the last test build b/c it was for a very small commit. I guess I shouldn't have but I feel bad for poor Travis working so hard all the time...

PR #326, opened 2017-05-24
6000 soft delete

hooverlunch:
I didn't have time to add specs for every single bit of SQL that got a `deleted_at IS NULL`. If there was an existing test I think I mostly added the expectation that deleted stuff is ignored.

I do think there is enough coverage to detect syntax errors on pretty much everything though. Just not maybe logic errors, but I'm not super worried about those as the logic is quite simple.


hooverlunch:
Merging this without review because Mel is out, Devney is swamped, the sprint is late, the tests are passing, and I feel good about it. :/

PR #327, opened 2017-05-31
6007 password reset with username

hooverlunch:


spec/requests/password_reset_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+describe PasswordResetsController, type: :request do
+  let(:user) { create(:user) }
+
+  context "if not logged in" do
+    it "new password reset page should load" do
+      get new_password_reset_path
+      expect(response).to be_success
+    end
+
+    it "submitting password reset form should reset perishable token and send email" do
```
DevneyHamilton: Can we add a context block for 'user with email'? took me a while to figure that out.

spec/requests/password_reset_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+    before do
+      login(user)
+      user2.reset_perishable_token!
+    end
+
+    it "attempting to load edit should logout existing user" do
+      get edit_password_reset_path(id: user2.perishable_token)
+      expect(controller.current_user).to be_nil
+    end
+  end
```
DevneyHamilton: So is it correct that there are 2 ways to get a new password: 1) when logged out, request a new password with your login or email and 2) the admin resets it for you?

spec/requests/password_reset_spec.rb
hooverlunch: Sure.

spec/requests/password_reset_spec.rb
hooverlunch: Correct. And the admin can see it on the screen (observers only) or have it emailed when they are resetting it.

PR #329, opened 2017-06-13
6461 response csv performance

hooverlunch:


app/models/results/csv/generator.rb
```diff
@@ -0,0 +1,168 @@
+module Results
```
hooverlunch: This file shows as totally new because I split the classes out into separate files. See the individual commits if you want to see what got changed. Mostly caching and eager loading stuff.

app/controllers/responses_controller.rb
```diff
@@ -65,10 +65,11 @@ def index
...
         end
 
-        # get the response, for export, but not paginated
-        @responses = @responses.with_associations.order(:created_at)
+        # Get the response, for export, but not paginated.
+        # We deliberately don't eager load as that is handled in the Results::Csv::Generator class.
+        @responses = @responses.order(:created_at)
 
-        @csv = ResponseCSV.new(@responses)
+        @csv = Results::Csv::Generator.new(@responses)
```
hooverlunch: FYI Results is the namespace that I'm envisioning containing stuff like Response, Answer, Choice, etc. someday.

PR #332, opened 2017-06-19
6559 remove phone uniqueness

meltheadorable:


app/models/sms/decoder.rb
```diff
@@ -118,11 +126,41 @@ def find_form
...
+    possible_users = User.by_phone(@msg.from).active
+    return log_authentication_failure unless possible_users.present?
+
+    # check auth code
+    auth_code = @tokens.first if @tokens.first =~ AUTH_CODE_FORMAT
+    @user ||= possible_users.find_by(sms_auth_code: auth_code)
+
+
+    # check assigned to mission
+    possible_users = possible_users.assigned_to(@msg.mission)
```
hooverlunch: Technically this line and line 136 should not be run if `@user` is already set. This line will fire an unnecessary query. Can you wrap these blocks in `if`s instead of using `||=`?

app/models/sms/decoder.rb
```diff
@@ -118,11 +126,41 @@ def find_form
...
+  end
+
+  def set_mission
+    form_code = (@tokens.first =~ AUTH_CODE_FORMAT) ? @tokens.second : @tokens.first
+    form_version = FormVersion.find_by(code: form_code)
+    # if version not found, raise error
+    raise_decoding_error("form_not_found", form_code: form_code) unless form_version
+    @mission = form_version.mission
+    @msg.mission = @mission
+    @msg.save!
```
hooverlunch: Nice.

app/models/sms/incoming.rb
```diff
@@ -21,4 +18,12 @@ def recipient_numbers
...
     [user: User::SITE, phone: to]
   end
+
+  private
+
+  def set_user
+    matching_users = User.by_phone(from).active
+    self.user ||= matching_users.first if matching_users.count == 1
+    true
+  end
```
hooverlunch: Do we need this? I thought when I looked that the only place this is needed is in sms receive, so we could remove it. If it is still needed, could you please add a comment explaining why it's in two places.

spec/requests/sms/incoming_sms_spec.rb
```diff
@@ -328,4 +328,52 @@
...
+    end
+
+    it "should get the oldest user otherwise" do
+      reply = assert_sms_response(
+        from: oldest_user.phone,
+        incoming: { body: "#{oldest_form.code} 1.5 2.Y" },
+        outgoing: /#{oldest_form.code}.+thank you/i,
+        url: submission_url)
+      user = reply.user
+      expect(user.id).to eq oldest_user.id
```
hooverlunch: Exquisite

app/models/sms/incoming.rb
meltheadorable: This is here because it saves a lot of annoyance & some querying in the simplest (most common?) case where we know exactly which user we're fetching going in because there's only one with that number. Adding a comment.

PR #338, opened 2017-07-12
6168 cookie session store

DevneyHamilton:


DevneyHamilton:
Not sure why the erd file is here - I ran migrations after merging develop, no updates to schema.rb.

PR #341, opened 2017-08-02
6810: Refactored users controller to remove bug

hooverlunch:
Cleaning up the users controller filters was enough to fix this I think. (The problem was that the admin checkbox was getting ignored on create.)

We don't have great feature spec coverage here and I didn't have time to add some. But I tested new and edit in the following cases:

* Observer editing profile
* Mission mode (where even admins have permission to create/update users for that mission only)
* Admin mode

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1,17 +1,15 @@
 class UsersController < ApplicationController
   include BatchProcessable, Searchable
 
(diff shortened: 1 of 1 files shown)


hooverlunch:
Hmm, I cancelled some Travis builds as I was just updating comments. The one I didn't cancel actually passed.

PR #342, opened 2017-08-03
6840 preserve files

hooverlunch:


app/models/upload_saver.rb
```diff
@@ -0,0 +1,33 @@
+require "fileutils"
+
+# Saves uploaded files.
+class UploadSaver
+  STORAGE_PATH = Rails.root.join(*%w(tmp uploads))
```
meltheadorable: Could we change this to `Rails.root.join("tmp", "uploads")`? It's actually the same number of characters and is a little less mechanically complex.

app/jobs/cleanup_job.rb
```diff
@@ -0,0 +1,19 @@
+# Cleans up various bits left around by the app.
+class CleanupJob < ApplicationJob
```
meltheadorable: Like this refactor 👍

app/controllers/option_set_imports_controller.rb
```diff
@@ -1,6 +1,4 @@
 class OptionSetImportsController < ApplicationController
-  include UploadProcessable
```
meltheadorable: Nice

PR #343, opened 2017-08-04
6808 non-one-screen groups

hooverlunch:


app/helpers/odk_helper.rb
```diff
@@ -163,31 +163,74 @@ def odk_option_translations(form, lang)
...
+      content_tag(:input, ref: "#{xpath}/#{node.odk_code}-header") do
+        tag(:hint, ref: "jr:itext('#{node.odk_code}-header:hint')")
+      end
     end
   end
 
-  def odk_group_inner(node, xpath)
-    render("forms/odk/group_inner", node: node, xpath: xpath)
+  def odk_group_body(node, xpath)
+    render("forms/odk/group_body", node: node, xpath: xpath)
```
DevneyHamilton: this is a better name :)

app/models/form_item.rb
```diff
@@ -97,7 +97,7 @@ def as_json(options = {})
     result = super(options)
   end
 
-  def has_group_child?
+  def group_children?
```
DevneyHamilton: this seems more ambiguous to me, thinking about other people reading the code. I wouldn't know what group_children means - it could mean are children grouped? should children be grouped?

spec/requests/form_odk_rendering_spec.rb
```diff
@@ -189,20 +261,61 @@
...
-    prepare_expectation("odk/forms/#{filename}",
+    xml = prepare_expectation("odk/forms/#{filename}",
+      formname: [form.name],
       form: [form.id],
       formver: [form.code],
       itemcode: items.map(&:odk_code),
       optcode: nodes.map(&:odk_code),
       optsetid: items.map(&:option_set_id).compact.uniq
     )
+    if save_expectations
```
DevneyHamilton: This is so helpful! Thank you!

DevneyHamilton:
A lot of these changes are really helpful. I suggest leaving 'has_group_child?' instead of 'group_children?' because the latter is really ambiguous. 'has_any_group_children?' would be nice too, but it's even longer.

PR #344, opened 2017-08-08
5946 barcode

DevneyHamilton:


DevneyHamilton:
On my local when I run specs, I'm getting a lot of `undefined method `one_screen=' for #<QingGroup:0x007fe182ac2630>` errors. I ran migrations. I think something might have been lost in the merge with develop that has 6808. Tom, I'll check with you tomorrow morning in case you can see what went wrong.

app/views/users/_login_instruction_box.html.erb
```diff
@@ -28,6 +28,10 @@
         <td class="value"><%= "#{@current_mission.name}" %></td>
       </tr>
     <% end %>
+    <tr>
+      <td class="name"><%=I18n.t("activerecord.models.odk_qr_code.one")%></td>
+      <td class="value qrcode"><%= raw @config_qr.as_html %></td>
```
hooverlunch: Cool!

spec/models/odk_config_generator_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      ODKConfigGenerator.new().generate_odk_config("test", "test", "test")
+    end
+
+    it "encodes the string" do
+      return_value = Base64.strict_encode64(Zlib::Deflate.deflate("string"))
+      expect(Base64).to receive(:strict_encode64).and_return(return_value)
+
+      ODKConfigGenerator.new().generate_odk_config("test", "test", "test")
+    end
+  end
```
hooverlunch: These specs are nice and thorough but really I don't think you need this much. 

All you really need to test is that you're generating the correct base64 string for a given url, username, and password. We can assume the QRCode library is correct.

What I would do is compute outside of Ruby a correct base64 string and put it in here. (Since you already have this all working, you can cheat a bit and dump out a valid one from the Rails console and use that as your ground truth since you know it's correct.) Then add an accessor method to your class that returns the computed base64 string you're generating. Then compare the two in the spec.

I'm not sure that we gain anything by testing that deflate gets called or that encode gets called or etc. I think that's a bit too fine grained and if you always write tests at that level it will really slow you down. If you get the test I described above passing, you know all of those things must be happening.

Also in the above test you're not actually testing e.g. if zip and encode are called in the right order. I think if you flipped them in your code, the tests would still pass.

I know this is a lot. Picking out the right things to test is one of the hardest things to do in TDD and so I want to be sure we're on the same page on this. Happy to discuss it tomorrow.

app/controllers/users_controller.rb
```diff
@@ -127,6 +127,7 @@ def bulk_destroy
   # shows printable login instructions for the user
   def login_instructions
     @site_url = admin_mode? ? basic_root_url : mission_root_url
+    @config_qr = ODKConfigGenerator.new().generate_odk_config(@user.login, flash[:password], @site_url)
```
hooverlunch: (Also for future reference you don't need parens if you're calling a method with no args.)

DevneyHamilton: I was going for a more functional approach. Maybe it doesn't work well with Ruby OO? A benefit is easier testing and easier reasoning about code because its more of a pure function. But consistency is probably more important, so I can change it to use ivars.

app/models/odk_config_generator.rb
```diff
@@ -0,0 +1,20 @@
+class ODKConfigGenerator
```
DevneyHamilton: oooh good will do.

app/models/odk_config_generator.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def generate_odk_config(username, password, site_url)
+    string = generate_string(username, password, site_url)
+    processed_string = Base64.strict_encode64(Zlib::Deflate.deflate(string))
+    qrcode = RQRCode::QRCode.new(processed_string)
+  end
+
+  def generate_string(username, password, site_url)
```
DevneyHamilton: I want to be able to test this method - can I do that if its private?

app/views/users/_login_instruction_box.html.erb
```diff
@@ -28,6 +28,10 @@
         <td class="value"><%= "#{@current_mission.name}" %></td>
       </tr>
     <% end %>
+    <tr>
+      <td class="name"><%=I18n.t("activerecord.models.odk_qr_code.one")%></td>
```
DevneyHamilton: oh cool

config/locales/en.yml
```diff
@@ -202,6 +202,13 @@ en:
         few: "Missions"
         many: "Missions"
         other: "Missions"
+      odk_qr_code:
+        zero: "ODK QR Codes"
+        one: "ODK QR Code"
+        two: "ODK QR Codes"
+        few: "ODK QR Codes"
+        many: "ODK QR Codes"
+        other: "ODK QR Codes"
```
DevneyHamilton: ok

app/controllers/users_controller.rb
hooverlunch: Interesting. I'd be keen to discuss this offline. But for now yeah maybe best to keep consistent?

DevneyHamilton:
Updated, but with different refactoring than suggested. Changed class to ODK:UserConfigEncoder. The users controller passes the encoding to the QRCode gem.

PR #345, opened 2017-08-09
6103: Added counter type

hooverlunch:


hooverlunch:
Tested on ODK with counter app, works like a charm 😸

PR #347, opened 2017-08-15
5960 read only

DevneyHamilton:


app/assets/javascripts/backbone/views/question_form_view.js.coffee
```diff
@@ -3,13 +3,27 @@ class ELMO.Views.QuestionFormView extends ELMO.Views.ApplicationView
   initialize: (options) ->
     @prefillableTypes = options.prefillableTypes # Will be null for Question form
     @toggleFields()
+    @toggleRequired(@$('#questioning_read_only')[0].checked)
+    @toggleReadOnly(@$('#questioning_prefill_pattern')[0].value)
```
DevneyHamilton: indexing into array seems like not a safe thing to do; is there a better way?

DevneyHamilton:
I tested in the browser, but I haven't tested that the ODK readonly attribute is rendering correctly on a phone.

app/assets/javascripts/backbone/views/question_form_view.js.coffee
hooverlunch: Super common pattern in jQuery, totally fine.

PR #354, opened 2017-09-06
6931 login rtl

DevneyHamilton:


DevneyHamilton:
Screenshot: screen shot 2017-09-06 at 9 26 02 am

Tested in Chrome, Firefox, Safari.

app/assets/stylesheets/all/components/_rtl.css.scss
```diff
@@ -1,9 +1,34 @@
 html[dir="rtl"] {
-  #userinfo {
-    float: left;
 
-    .icon-btn {
+  #content {
+    h1.title{
       float: right;
+      margin-right: 0;
```
hooverlunch: So this is the kind of stuff I'd like to see us use:

http://matanich.com/2013/09/06/rtl-css-with-sass

I'd rather not put all of these separate styles for RTL in a separate file. It will be difficult to remember to keep it up to date.

DevneyHamilton:
When this is ready, I will merge into the header branch. We should put them in together so we can get rid of the rtl css file altogether. The failing is related to the rank assignment bug (destroy_questioning doesn't work).

PR #364, opened 2017-10-10
7276: Submitter usernames in reports

meltheadorable:


config/locales/en.yml
```diff
@@ -1094,6 +1094,7 @@ en:
   attrib_fields:
     form: "Form"
     submitter: "Submitter"
+    submitter_username: "Submitter Username"
```
hooverlunch: Can we change the above label to 'submitter name' for consistency?

app/models/report/attrib_field.rb
```diff
@@ -64,6 +64,13 @@ def title
       name: :submitter,
       name_expr_params: {sql_tplt: "__TBL_PFX__users.name", name: "name", clause: :select},
       value_expr_params: {sql_tplt: "__TBL_PFX__users.name", name: "value", clause: :select},
+        joins: [:users],
+        data_type: :text,
+        groupable: true},
+    submitter_username: {
+      name: :submitter_username,
+      name_expr_params: { sql_tplt: "__TBL_PFX__users.login", name: "name", clause: :select },
+      value_expr_params: {sql_tplt: "__TBL_PFX__users.login", name: "value", clause: :select},
```
hooverlunch: Nice

PR #365, opened 2017-10-11
7170 conditionals field react

DevneyHamilton:


app/assets/javascripts/components/conditions_form_field.jsx
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  getFieldData(ref_qing_id) {
+    ELMO.app.loading(true)
+    var self = this;
+    var url = this.buildUrl(ref_qing_id)
+    $.ajax(url)
+      .done(function(response) {
+          self.setState(response);
+        })
```
DevneyHamilton: Should I do more to check that the response is valid state here, or wait for an error in render? As in, if there's a 200 response but not the right structure for json?

app/controllers/ping_controller.rb
```diff
@@ -9,6 +9,11 @@ def show
     render layout: nil, formats: :text, status: @ok ? 200 : 503
   end
 
+
+  def json
```
DevneyHamilton: not sure if I should leave this for the future or not.

DevneyHamilton:
This can't be merged until we add back in cascading selects.

app/serializers/condition_view_serializer.rb
```diff
@@ -0,0 +1,28 @@
...
+  def questioning_id
+    object.questioning_id
+  end
+
+  def refable_qing_options
+    object.refable_qings.map{ |q| {code: q.question.code, rank: q.full_dotted_rank, id: q.id} }
+  end
+
+  def operator_options
+    object.applicable_operator_names.map { |n| {name: I18n.t(n, scope: [:condition, :operators]), id: n} }
```
hooverlunch: By convention we usually do this as `I18n.t("condition.operators.#{n}")`. Both work though.

app/views/conditions/_form_fields.html.erb
```diff
@@ -1,28 +1,5 @@
...
-    <% if @condition.ref_qing.present? && @condition.ref_qing_has_options? %>
-      <%= render('option_sets/cascading_selects',
-            index: 1,
-            option_node_path: @condition.option_node_path,
-            input_name_template: 'questioning[condition_attributes][option_node_ids][]') %>
-    <% else %>
-      <%= f.text_field(:value, class: 'text form-control') %>
-    <% end %>
-  </div>
+  <%= react_component("ConditionsFormField", ConditionViewSerializer.new(@condition)) %>
```
hooverlunch: !!!!!!

hooverlunch: ❤️

config/locales/en.yml
```diff
@@ -359,6 +359,10 @@ en:
         which_phone: "Which Phone(s)"
         recipient_selection: "Recipients"
         recipient_ids: "Users/Groups"
+      condition:
+        ref_qing_id: "Question"
+        op: "Comparison"
+        value: "Value"
```
hooverlunch: 👍

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
   end
+
+  describe "condition_form_data" do
+
+    let(:form) { create(:form, :published, question_types: %w(integer text integer text)) }
+    let(:qing) { form.questionings[2] }
+
+    context "without ref_qing_id" do
+      it "returns json with ref qing id options, no operator options, and no value options" do
+        expected_ref_qing_options = form.questionings.select{|q| q.id < qing.id }.map{ |q| { code: q.question.code, rank: q.full_dotted_rank, id: q.id } }
```
hooverlunch: Styleguide calls for space after `map`, `select` etc., before block {.

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
+        }.to_json
+
+        get "/en/m/#{get_mission.compact_name}/questionings/condition-form",
+          { ref_qing_id: nil,
+            form_id: form.id,
+            questioning_id: qing.id
+          }
+
+        expect(response).to have_http_status(200)
+        expect(response.body).to eq expected
```
hooverlunch: I think it would be fine and clear to just inline the expected JSON here, like 

```
expect(response.body).to eq {
    ...
}
```

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
+          value: nil,
+          form_id: form.id,
+          questioning_id: qing.id,
+          refable_qing_options: expected_ref_qing_options,
+          operator_options: [],
+          value_options: nil
+        }.to_json
+
+        get "/en/m/#{get_mission.compact_name}/questionings/condition-form",
+          { ref_qing_id: nil,
```
hooverlunch: I think styleguide calls for this brace to be at end of prev line, and then one level of indent for the attribs.

spec/requests/questionings_spec.rb
hooverlunch: Selecting questionings by comparing IDs won't work after we switch to UUIDS. I think you can just do: `form.c[0..1].map { |q| ... }`.

`form.c` is a shortcut alias for `form.sorted_children`, good for use in specs, especially when you have nested groups like `form.c[2].c[1].c[3]`

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
       end
     end
   end
+
+  describe "condition_form_data" do
+
+    let(:form) { create(:form, :published, question_types: %w(integer text integer text)) }
+    let(:qing) { form.questionings[2] }
```
hooverlunch: This can just be `form.c[2]`.

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
+      it "returns json with operator options" do
+        expected_operator_options = [
+          { "name":"is equal to","id":"eq" },
+          { "name":"is less than","id":"lt" },
+          { "name":"is greater than","id":"gt" },
+          { "name":"is less than or equal to","id":"leq" },
+          { "name":"is greater than or equal to","id":"geq" },
+          { "name":"is not equal to","id":"neq" }
+        ]
+        expected_ref_qing_options = form.questionings.select{ |q| q.id < qing.id}.map{ |q| { code: q.question.code, rank: q.full_dotted_rank, id: q.id } }
```
hooverlunch: If this is the same as above, it should be in a `let` block?

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
+
+        expect(response).to have_http_status(200)
+        expect(response.body).to eq expected
+      end
+    end
+
+    context "with ref_qing_id" do
+      it "returns json with operator options" do
+        expected_operator_options = [
+          { "name":"is equal to","id":"eq" },
```
hooverlunch: Style adjustment: `{name: "is equal to", id: "eq"}`

Note that `{"a": "b"}` still results in `{:a => "b"}` (symbol key). To get string keys you need to do `{"a" => "b"}`.

spec/requests/questionings_spec.rb
```diff
@@ -52,4 +52,71 @@
...
+          { "name":"is less than","id":"lt" },
+          { "name":"is greater than","id":"gt" },
+          { "name":"is less than or equal to","id":"leq" },
+          { "name":"is greater than or equal to","id":"geq" },
+          { "name":"is not equal to","id":"neq" }
+        ]
+        expected_ref_qing_options = form.questionings.select{ |q| q.id < qing.id}.map{ |q| { code: q.question.code, rank: q.full_dotted_rank, id: q.id } }
+        expected = {
+          id: nil,
+          ref_qing_id: form.questionings[0].id,
```
hooverlunch: `f.c[0].id`. Same elsewhere.

app/controllers/ping_controller.rb
```diff
@@ -22,4 +27,5 @@ def seconds_since_file_modified(path)
   rescue Errno::ENOENT
     1_000_000
   end
+
```
hooverlunch: Remove?

app/assets/javascripts/components/form_label.jsx
```diff
@@ -0,0 +1,5 @@
+class FormLabel extends React.Component {
+  render() {
+    return <label htmlFor={this.props.for} key={this.props.for}> {this.props.text} </label>
```
hooverlunch: Can we get rid of the extra space inside the tag? Worried it might cause styling issues.

app/assets/javascripts/components/form_select.jsx
```diff
@@ -0,0 +1,17 @@
+class FormSelect extends React.Component {
+  render() {
+    let options = [<option value="" value={this.props.id} key="blank"></option>]
```
hooverlunch: Why two value attribs?

app/serializers/condition_view_serializer.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def value_options
+    nil
+  end
```
hooverlunch: Great looking serializer! I guess the value options will come when we do the cascading select stuff?

app/assets/javascripts/components/form_field.jsx
```diff
@@ -0,0 +1,10 @@
+class FormField extends React.Component {
```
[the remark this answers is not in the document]
DevneyHamilton: was a mistake to have specific name in there. Not sure how generalizable this will be, but wanted to steer it in the reusable direction.

PR #373, opened 2017-10-23
7198: Added multiple condition support to replication

hooverlunch:


app/models/questioning.rb
```diff
@@ -17,7 +17,7 @@ class Questioning < FormItem
 
   scope :visible, -> { where(hidden: false) }
 
-  replicable child_assocs: [:question, :condition], backward_assocs: :form,
+  replicable child_assocs: [:question, :display_conditions], backward_assocs: :form,
```
DevneyHamilton: Is there a way to document this as a place for potential complexity the next time this gets updated? Seems like :display_conditions just got more complex.

spec/models/replication/form_replication_spec.rb
```diff
@@ -131,6 +131,32 @@
       end
     end
 
+    context "with a question with multiple conditions" do
```
DevneyHamilton: using the keys means this will hold up with switch to uuids, right?

hooverlunch: Yes.

app/models/questioning.rb
hooverlunch: I think it will be captured by the replication for skip logic issue. And ultimately this old replication system will be subsumed by the new, so there's that...

PR #374, opened 2017-11-06
7336 i18n allowed locales

hooverlunch:


app/models/concerns/translatable.rb
```diff
@@ -2,17 +2,19 @@
...
+      options = default_options.merge(options)
+
+      # Tidy options if given.
+      options[:locales] = options[:locales].map(&:to_s) if options[:locales].is_a?(Array)
 
-      # save the list of translated fields
+      class_variable_set('@@translate_options', options)
       class_variable_set('@@translated_fields', args)
 
       # Setup an accessor if not present.
```
DevneyHamilton: Should this comment still be here?

PR #376, opened 2017-11-10
7385 conditions in repeats

hooverlunch:


hooverlunch:
I think we're good on QA. The feature spec coverage is solid I think. Thanks!

PR #377, opened 2017-11-13
5993: convert to uuids

meltheadorable:


app/models/form_item.rb
```diff
@@ -100,8 +100,9 @@ def preordered_descendants
   end
 
   def sorted_children
-    children.order(:rank)
+    sorted = children.order(:rank)
   end
+  alias_method :sc, :sorted_children
```
hooverlunch: `c` is already aliased to sorted_children so you can remove this.

config/initializers/ancestry.rb
```diff
@@ -0,0 +1,4 @@
+module Ancestry
+  send :remove_const, :ANCESTRY_PATTERN
+  const_set :ANCESTRY_PATTERN, /\A[\w\-]+(\/[\w\-]+)*\z/
+end
```
hooverlunch: Nice one. Maybe a quick comment here?

db/migrate/20171112003313_migrate_primary_keys_to_uuid.rb
```diff
@@ -0,0 +1,234 @@
...
+          puts "NULLIFY ORPHAN"
+          orphan.update_columns(foreign_key_column => nil)
+        else
+          puts "DELETE ORPHAN"
+          orphan.delete
+        end
+      end
+    end
+  end
+end
```
hooverlunch: 😮

db/schema.rb
```diff
@@ -26,106 +27,119 @@
     t.integer "inst_num", default: 1, null: false
     t.decimal "latitude", precision: 8, scale: 6
     t.decimal "longitude", precision: 9, scale: 6
-    t.integer "option_id"
-    t.integer "questioning_id"
+    t.integer "old_id", default: "nextval('answers_id_seq'::regclass)"
```
hooverlunch: This is interesting. What is the purpose of this?

script/table_test.rb
```diff
@@ -0,0 +1,52 @@
+require 'active_support/core_ext/digest/uuid'
```
hooverlunch: Can you add a comment please summarizing what this script is for?

spec/spec_helper.rb
```diff
@@ -10,6 +10,7 @@
 require 'capybara/poltergeist'
 require 'capybara-screenshot/rspec'
 require 'paperclip/matchers'
+require 'cancan/matchers'
```
hooverlunch: Nice.

db/schema.rb
meltheadorable: actually i need to fix that, ids retain their "create the next thing in the sequence" behavior even after being removed as primary keys and their name change, i forgot to reset the defaults for those fields, `schema.rb` doesn't actually support setting them this way so if you tried to load this schema and then do another migration, the next dump would have no default value for those columns, but as it stands, `old_id` will be non-null and keep counting up for new records post-migration for existing installations

script/table_test.rb
meltheadorable: deleting this, it was a rough draft of navigating the structure i had ironed out for the table info and for the general process, doesn't do anything anymore

PR #382, opened 2017-12-01
7253 printable hint

DevneyHamilton:


DevneyHamilton:
This is what it looks like: 
screen shot 2017-12-01 at 12 54 45 pm

DevneyHamilton:
This is what it looks like: screen shot 2017-12-01 at 12 56 28 pm

app/decorators/printable/qing_decorator.rb
```diff
@@ -0,0 +1,17 @@
...
+  class QingDecorator < ::ApplicationDecorator
+    delegate_all
+
+    def name_and_rank
+      str = "#{full_dotted_rank}. "
+      str << h.reqd_sym if required?
+      str << name
+    end
+
+    def question_type
```
hooverlunch: This method looks good. The name is a little misleading though I think because I would expect this method (if calling as `qing.question_type` to just return the QuestionType object associated with the questioning.

In the context of just the decorator it might make sense but from the caller's perspective there is no way to tell one method is a decorator method vs a core object method.

A better name might be `selection_instructions`, which makes it clear that this is really meant as a cue to the user on how many boxes to check.

PR #387, opened 2017-12-11
6581 observer to enumerator

DevneyHamilton:
Not sure about schema.rb updates.

### DIFF
--- a/app/models/broadcast.rb
+++ b/app/models/broadcast.rb
@@ -23,7 +23,7 @@ class Broadcast < ApplicationRecord
   WHICH_PHONE_OPTIONS = %w(main_only alternate_only both)
 
   # options for recipients
-  RECIPIENT_SELECTION_OPTIONS = %w(all_users all_observers specific)
+  RECIPIENT_SELECTION_OPTIONS = %w(all_users all_enumerators specific)
 
   scope :manual_only, -> { where(source: "manual") }
 
@@ -107,7 +107,7 @@ class Broadcast < ApplicationRecord
 
   private
 
-  # Returns the recipients of the message. If recipient_selection is set to all_users or all_observers,
+  # Returns the recipients of the message. If recipient_selection is set to all_users or all_enumerators,
   # this will be different than `recipients`.
   def actual_recipients
     @actual_recipients ||= case recipient_selection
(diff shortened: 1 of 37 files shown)


app/models/broadcast.rb
```diff
@@ -23,7 +23,7 @@ def self.receivable_association
   WHICH_PHONE_OPTIONS = %w(main_only alternate_only both)
 
   # options for recipients
-  RECIPIENT_SELECTION_OPTIONS = %w(all_users all_observers specific)
+  RECIPIENT_SELECTION_OPTIONS = %w(all_users all_enumerators specific)
```
hooverlunch: I think we will need a migration for this one, no?

config/locales/en.yml
```diff
@@ -1134,7 +1134,7 @@ en:
         both: "Both SMS and email"
     recipient_selection_options:
       all_users: "All users in mission"
-      all_observers: "All observers in mission"
+      all_enumerators: "All enumerators in mission"
```
hooverlunch: I think key changes will need to be made in all the translation files, right?

db/schema.rb
```diff
@@ -124,11 +124,10 @@
     t.string "op", limit: 255
     t.uuid "option_node_id"
     t.integer "option_node_old_id"
-    t.uuid "questioning_id"
+    t.uuid "questioning_id", null: false
     t.integer "questioning_old_id"
-    t.uuid "ref_qing_id"
+    t.uuid "ref_qing_id", null: false
     t.integer "ref_qing_old_id"
-    t.string "role", null: false
```
hooverlunch: I think what's going on here is that your DB has been migrated for one of the condition branches that haven't been merged yet. Can you please remove this from the PR (ideally using rebase to keep the history clean). Then I think most of those test failures will go away.

config/locales/en.yml
DevneyHamilton: I wasn't sure about this - how will translators for other languages know those ones should be updated?

db/schema.rb
DevneyHamilton: thanks!

config/locales/en.yml
hooverlunch: They will appear as new strings that need to be translated I believe. (If the key changes).

If the key doesn't change, not sure. Presumably Transifex should be smart enough to see there was a change to the value and prompt the translator to doublecheck?

app/models/broadcast.rb
DevneyHamilton: yes!

config/locales/en.yml
DevneyHamilton: updated keys in all the files

PR #394, opened 2017-12-19
7611: Delete Condition Draft

meltheadorable:
This has absolutely no styling work done on it, that should come with #7612


app/assets/javascripts/components/ConditionsFormField.jsx
```diff
@@ -99,14 +104,32 @@ class ConditionsFormField extends React.Component {
...
     let value_field_props = this.buildValueProps(name_prefix, id_prefix);
-    return (
-      <div className="condition-fields">
-        <input {...condition_field_props}/>
-        <FormField {...ref_qing_field_props} />
-        <FormField {...operator_field_props} />
-        <FormField {...value_field_props} />
-      </div>
-    );
+    if(this.state.destroy == true) {
```
hooverlunch: I think there is functionally no difference between `x == true` and `x` in this context. If you want to match `true` exactly you need `===`

PR #395, opened 2017-12-19
7530 skip rule validation

hooverlunch:


app/models/form.rb
```diff
@@ -71,7 +71,6 @@ def self.receivable_association
   def self.terminate_sub_relationships(form_ids)
     Form.where(id: form_ids).update_all(original_id: nil)
     FormVersion.where(form_id: form_ids).delete_all
-    Questioning.where(form_id: form_ids).delete_all
```
meltheadorable: Why this change?

app/models/mission.rb
```diff
@@ -34,7 +34,7 @@ class Mission < ApplicationRecord
 
   scope(:sorted_by_name, -> { order("name") })
   scope(:sorted_recent_first, -> { order("missions.created_at DESC") })
-  scope(:for_user, ->(u) { where("missions.id IN 
+  scope(:for_user, ->(u) { where("missions.id IN
```
meltheadorable: Is there a way to write this query in ActiveRecord?

Something like `where(id: Assignment.where(user_id: u.id).pluck(:mission_id))` (this should pick up paranoia's default queries)?

I think we should be working to avoid hand-crafted SQL when possible since it helps us avoid a lot of the types of migration pains we experienced switching to  postgres and then to UUIDs.

db/migrate/20171219154111_convert_skip_rule_to_uuid.rb
```diff
@@ -0,0 +1,23 @@
+class ConvertSkipRuleToUuid < ActiveRecord::Migration
```
meltheadorable: Good call. Do you know if there's a way to make UUID the new default on rails 4? Might be worth that config change if it can be done. Otherwise we should put something in the backlog to make sure we set that up on rails 5 as I'm pretty sure it _can_ be done there.

spec/models/skip_rule_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        rule = create(:skip_rule, source_item: qing2, conditions_attributes: [
+          {ref_qing_id: qing1.id, op: "eq", value: "5"},
+          {ref_qing_id: "", op: "", value: ""}
+        ])
+        expect(rule.conditions.count).to eq 1
+        expect(rule.conditions[0].ref_qing).to eq qing1
+      end
+    end
+
+    describe "skip_if" do
```
meltheadorable: these specs look great 💯

spec/models/mission_destroy_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+  let(:deleted_counts) { expected_counts.map { |k, v| [k, 0] }.to_h }
+
+  it "should delete all objects in mission" do
+    expect(actual_counts).to eq expected_counts
+    mission.destroy
+    expect(actual_counts).to eq deleted_counts
+  end
+
+  def actual_counts
```
meltheadorable: Great refactor here.

app/models/form.rb
hooverlunch: Specs pass without it. Questionings get deleted by Mission directly. This code has a lot of quirks, just shaving off a few.

app/models/mission.rb
hooverlunch: Sure I can try it. Or even `where(id: user.assignments.pluck(:id))`, right? 

This was just a whitespace change but I guess that's my rotten luck :P

db/migrate/20171219154111_convert_skip_rule_to_uuid.rb
hooverlunch: http://www.mccartie.com/2015/10/20/default-uuid's-in-rails.html

I will add.

spec/models/mission_destroy_spec.rb
hooverlunch: Thanks!

app/models/mission.rb
hooverlunch: Actually turns out this method is a crappier version of user.missions and is only used once so I got rid of it :D

PR #396, opened 2017-12-27
7200 skip for odk

DevneyHamilton:
Need to deal with tests failing because of new eagerload keyword.


app/controllers/forms_controller.rb
```diff
@@ -78,7 +78,7 @@ def show
 
         # xml style defaults to odk but can be specified via query string
         @style = params[:style] || 'odk'
-
+        setup_condition_computer
```
DevneyHamilton: @meltheadorable I made the condition computer available here in the controller to be used in the odk views, in case it's relevant to your web view work.

app/decorators/odk/condition_group_decorator.rb
```diff
@@ -0,0 +1,21 @@
+module Odk
+  class ConditionGroupDecorator < BaseDecorator
+
+    def to_odk
+      conjuction = (true_if == "all_met") ? I18n.t("common.and") : I18n.t("common.or")
+      result = decorated_members.map{ |m| "(#{m.to_odk})"}.join(" #{conjuction} ")
+      if negate?
+        result = "not(#{result})"
+      end
+      decorated_members.count == 0 ? nil : result
```
hooverlunch: This would typically go as a guard clause at the top instead of at the bottom so you save the effort of all the computation.

app/decorators/odk/form_item_decorator.rb
```diff
@@ -48,16 +48,8 @@ def xpath_to(dest)
...
-          display_conditions.first.to_odk
-        else
-          operator = display_if == "all_met" ? " and " : " or "
-          display_conditions.map { |c| "(#{c.to_odk})" }.join(operator)
-        end
-      else
-        nil
-      end
+      computed_condition_group = context[:condition_computer].condition_group_for(object)
+      ConditionGroupDecorator.decorate(computed_condition_group).to_odk
```
hooverlunch: Woop woop!

app/models/form_item.rb
```diff
@@ -156,6 +159,14 @@ def display_conditionally?
     display_if != "always" && display_conditions.any?
   end
 
+  def condition_group
+    @condition_group ||= Forms::ConditionGroup.new(true_if: display_if, members: display_conditions)
+  end
+
+  def computed_condition_group
```
hooverlunch: We ended up not using this method, right, due to performance concerns? And the delegator above? I think we should remove it because it's too tempting.

spec/decorators/odk/condition_group_decorator_spec.rb
```diff
@@ -0,0 +1,74 @@
+require "spec_helper"
+
+module Odk
+  describe ConditionGroupDecorator, :odk, :reset_factory_sequences, database_cleaner: :truncate do
```
hooverlunch: Are these spec flags required given that you're mocking?

spec/requests/form_odk_rendering_spec.rb
```diff
@@ -188,6 +194,28 @@
...
+        ["integer", "integer", "integer"],
+        "integer",
+        ["integer", "integer", ["integer", "integer"]],
+        "integer"
+      ])
+    end
+
+    it "should render proper xml" do
+      form.c[0].skip_rules.build(destination: "item", dest_item: form.c[6])
+    end
```
hooverlunch: What is this? Doesn't seem like it does anything.

spec/decorators/odk/condition_group_decorator_spec.rb
DevneyHamilton: I'm making a form in the spec still.

spec/requests/form_odk_rendering_spec.rb
DevneyHamilton: False start I lost track of and will remove. Actual spec is farther down the file.

PR #398, opened 2018-01-03
7180 computed conditions

hooverlunch:
Adds ConditionComputer and specs and supporting classes.

Also adds support for display conditions on groups for ODK form rendering. This was a cleaner way to apply skip rules (skip the whole group with one computed condition instead of adding computed conditions to every single member of the group). It's also a nice feature that we really should have separate from this. I also created issues for adding condition capability to groups in other areas (e.g. web responses, form builder).

### DIFF
--- a/app/models/form_item.rb
+++ b/app/models/form_item.rb
@@ -101,9 +101,9 @@ class FormItem < ApplicationRecord
   end
 
   def preordered_descendants(eager_load: nil)
-    rel = descendants.order(:rank)
-    rel = rel.includes(eager_load) if eager_load
-    self.class.sort_by_ancestry(rel) { |a, b| a.rank <=> b.rank }
+    items = descendants.order(:rank)
+    items = items.includes(eager_load) if eager_load
+    self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }
   end
 
   def sorted_children
--- a/app/models/form_item.rb
+++ b/app/models/form_item.rb
@@ -100,7 +100,7 @@ class FormItem < ApplicationRecord
     end
   end
 
-  def preordered_descendants(eager_load:)
+  def preordered_descendants(eager_load: nil)
     rel = descendants.order(:rank)
     rel = rel.includes(eager_load) if eager_load
     self.class.sort_by_ancestry(rel) { |a, b| a.rank <=> b.rank }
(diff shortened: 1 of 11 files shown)


app/models/form_item.rb
```diff
@@ -100,8 +100,10 @@ def descendant_questionings(nodes = nil)
     end
   end
 
-  def preordered_descendants
-    self.class.sort_by_ancestry(descendants.order(:rank)) { |a, b| a.rank <=> b.rank }
+  def preordered_descendants(eager_load: nil)
+    rel = descendants.order(:rank)
```
DevneyHamilton: I don't know what 'rel' stands for here? Could you comment or update for future devs?

app/models/forms/condition_computer.rb
```diff
@@ -0,0 +1,69 @@
+# Computes display conditions that are implied from SkipRules.
```
DevneyHamilton: thanks for this comment! so succinct and helpful

app/models/forms/condition_computer.rb
```diff
@@ -0,0 +1,69 @@
...
+    def condition_group_for(item)
+      table[item] || empty_group
+    end
+
+    def build_table
+      self.table = {}
+      self.active_rules = []
+      self.last_item_cache = {}
+      form_items.each do |item|
+        add_display_conditions(item)
```
DevneyHamilton: Doing this work in the same loop depends on the items being preordered, correct? I think a comment about that would be helpful for the future.

spec/models/forms/condition_computer_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+        "integer",
+        ["integer", "integer", "integer"],
+        "integer",
+        ["integer", "integer", ["integer", "integer"]],
+        "integer"
+      ])
+    end
+    let(:form_items) { computer.form_items }
+
+    # ConditionGroup doesn't really contain a _name attrib, but this can be useful
```
DevneyHamilton: Thanks for this comment.

hooverlunch:
@DevneyHamilton Thanks! I ended up addressing your comments by just making variable names more verbose or explanatory which I think is even better than a comment! Thanks for pointing out unclear bits!

PR #399, opened 2018-01-04
7532: Update gems

meltheadorable:


Gemfile
```diff
@@ -156,6 +155,7 @@ group :development do
   # misc
   gem "fix-db-schema-conflicts", "~> 2.0.0"
   gem "letter_opener", "~> 1.4.1"
+  # gem "rubocop", "~> 0.52.1", require: false
```
hooverlunch: Curious why this is commented. Can we just delete this, or if you think we should leave it, can you put a comment as to why we have a commented line in the gemfile?

.ruby-version
```diff
@@ -1 +1 @@
-2.2.0
```
hooverlunch: How did you choose this version? Just curious.

I will add installing new ruby to the deploy checklist for 7.3.

meltheadorable: Honestly, it was the one I already had installed and ruby-build takes forever on the macbook air. In a vacuum I'd suggest we go all the way to 2.4.3.

hooverlunch: I think that is best if you don't mind. Run it the bg while you work on sth else? The fewer times we have to update ruby on the farm of ELMOs out there, the better.

Gemfile
meltheadorable: Wasn't supposed to be commented, I think I had been troubleshooting something with the gem installation process with the axlsx update and I forgot to put it back in, it's supposed to be there.

.ruby-version
meltheadorable: Already running in the background now. There's probably a way to make capistrano run `rbenv install` for us, do you want me to look into adding that to the deploy project?

PR #401, opened 2018-01-11
7172 skip logic component

hooverlunch:
I didn't have time to write feature specs for the skip rule form. On the other hand, I did write a bunch of specs for multiple conditions under display logic, which came in handy for skip logic too. I broke writing of feature specs out into another issue.


spec/requests/form_items_spec.rb
```diff
@@ -42,4 +42,143 @@
       end
     end
   end
+
```
hooverlunch: These were moved from questioning_spec without much change.

PR #406, opened 2018-02-07
7896 option levels bug

jpdimond:


app/assets/javascripts/legacy/views/option_set_form.js
```diff
@@ -166,9 +166,6 @@
 
     // show/hide the option levels field
     self.option_levels_field.show(checked)
-
-    // enable/disable nested options
-    self.options_field.list.allow_nesting(checked)
```
jpdimond: This isn't something that should depend on whether levels are enabled or not.

app/assets/javascripts/legacy/views/option_levels_field.js
```diff
@@ -24,6 +24,9 @@
         'edit': I18n.t('option_set.edit_option_level')
       }
     });
+
+    // No nesting for option levels
+    self.list.allow_nesting(false);
```
hooverlunch: So I'm assuming this fixes it?

app/assets/javascripts/legacy/views/option_set_form.js
hooverlunch: Won't this disable nested options? Note this is being called on `options_field` and not `option_levels_field`.

app/assets/javascripts/legacy/views/option_levels_field.js
jpdimond: yup!

app/assets/javascripts/legacy/views/option_set_form.js
jpdimond: ah you are right, i put it back in, and the other fix still works!

PR #415, opened 2018-02-23
7178 printable skip rule

DevneyHamilton:
screen shot 2018-02-23 at 5 41 19 pm
Current status


hooverlunch:
Good start! Couple quick notes: needs some I18n and a feature spec in addition to the style work. Should all be quick though. Then decorator can be reused in the read-only mode issue!

DevneyHamilton:
screen shot 2018-03-06 at 2 48 05 pm

PR #428, opened 2018-04-09
8182 3 answer heirarchy

DevneyHamilton:


DevneyHamilton:
This is not done yet. We should pair on either 1) checking if existing code is enough for answer_hierarchies to work for all subclasses of response_node or 2) renaming answers table to response_node.

smoyth:
Looks solid to me! Good work!

PR #430, opened 2018-04-11
8254 username uniqneness error

arisah-ijeoma:


spec/models/user_spec.rb
```diff
@@ -94,6 +94,16 @@
       expect(user).to be_valid
       expect(user.login).to eq "foobar"
     end
+
+    describe "uniqueness" do
+      before { @user = create(:user, login: "jayita") }
```
smoyth: We generally don't use ivars in rspec tests. It's kind of a code smell because `let` is a better way to accomplish this. So you can do `let(:user) { create(...) }`.

spec/models/user_spec.rb
```diff
@@ -94,6 +94,16 @@
...
+
+      it "can create a user with the same login after deleting" do
+        @user.destroy
+        @user2 = build(:user, login: "jayita")
+        expect(@user2).to be_valid
+      end
+    end
```
smoyth: I would also add a spec to test the main functionality. You want to be sure you didn't break _all_ uniqueness validation by adding the deleted_at scope.

So the same as above but take out the `destroy` part. This should go above the one you've already done. Be sure to have an expectation for the error message as we've done before.

app/models/user.rb
```diff
@@ -36,7 +36,7 @@ class User < ApplicationRecord
     c.logged_in_timeout(SESSION_TIMEOUT)
 
     c.validates_format_of_login_field_options = {with: /\A[a-zA-Z0-9\._]+\z/}
-    c.merge_validates_uniqueness_of_login_field_options(unless: :batch_creation?)
+    c.merge_validates_uniqueness_of_login_field_options(unless: :batch_creation?, scope: :deleted_at)
```
smoyth: Glad this worked out so cleanly!

PR #433, opened 2018-04-19
8253 form deletion bug

arisah-ijeoma:


spec/models/response_spec.rb
```diff
@@ -97,4 +97,22 @@
...
+      it "on response deletion" do
+        response.destroy
+        expect(form.responses_count).to eq(0)
+      end
+
+      it "on response creation" do
+        create(:response, user: user, form: form, answer_values: %w[1])
+        expect(form.responses_count).to eq(2)
+      end
+    end
```
smoyth: In general though, terrific job on these specs!

db/migrate/20180419135452_correct_form_responses_count.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Update old form response counts
+class CorrectFormResponsesCount < ActiveRecord::Migration
+  def up
+    execute("UPDATE forms SET responses_count = (
+      SELECT COUNT(*) FROM responses WHERE responses.form_id = forms.id AND responses.deleted_at IS NULL
+    )")
```
smoyth: Fantastic!

spec/models/response_spec.rb
```diff
@@ -97,4 +97,22 @@
...
+
+  context "responses count" do
+    let!(:user) { create(:user) }
+    let!(:form) { create(:form, question_types: %w[integer]) }
+    let!(:response) { create(:response, user: user, form: form, answer_values: %w[1]) }
+
+    describe "updates response count correctly" do
+      it "on response deletion" do
+        response.destroy
+        expect(form.responses_count).to eq(0)
```
smoyth: There is one key issue here.. You should add `reload` after `form`. If you do, I think you'll find that the spec fails. That's because you're not saving the form in the after_destroy callback.

This is one of the most important things to consider when doing ActiveRecord development and when writing specs for it.

Sometimes it is even best to test before AND after a reload. So I would leave this line and add a second one with a test of the reloaded value.

This stuff isn't easy but you'll get it eventually. Just always try to keep in mind what's in memory vs. what's persisted to the DB.

app/models/response.rb
```diff
@@ -27,6 +27,10 @@ class Response < ApplicationRecord
   before_save(:normalize_answers)
   before_create :generate_shortcode
 
+  # rails counter_cache increments on creation but does not decrement on deletion
+  # since we need the counter cache, we'll manually decrement on deletion
```
smoyth: Please mention that acts_as_paranoid is the reason for this, and provide a link to the open Github issue. That way if it gets resolved in future we might could take this out.

smoyth: https://github.com/ActsAsParanoid/acts_as_paranoid/issues/39

PR #437, opened 2018-05-01
8152 fix css

arisah-ijeoma:
I used a generic classname

As I am not fully familiar with the codebase, I didn't want to change too much too fast and risk breaking a styling I didn't even know existed

@smoyth

### DIFF
--- a/app/assets/stylesheets/print/_form.scss
+++ b/app/assets/stylesheets/print/_form.scss
@@ -44,7 +44,7 @@
     }
   }
 
-  tr.qing-group, tr.question {
+  tr.depth {
     &.depth-2 td.name {
       border-left: 20px solid $light-border-color;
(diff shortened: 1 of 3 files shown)


arisah-ijeoma:
screen shot 2018-05-01 at 1 08 37 pm

smoyth:
The code here looks great! `depth` is a good choice of class name. The CI failure is just a flapping spec so you can ignore that, and I've cleared the one hound error.

One thing I ask in future is for you to be a bit more descriptive in your commit names. `fix css` is quite generic. It helps to have the issue number of course, but if you can write a bit more than that it helps me when I'm doing stuff like backporting bugs and such. So for this one, you could write e.g. `fix indentation of skip logic in printable form`. Cool?

PR #441, opened 2018-05-17
7196 question batch delete

arisah-ijeoma:


spec/models/batch_destroy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      let(:user) { create(:user, mission: create(:mission)) }
+      let(:batch) { users }
+
+      it "deletes everyone but the current user" do
+        users << current_user
+        destroyer.destroy!
+
+        # current user
+        expect(User.count).to eq(1)
+        expect(current_user.active).to be_truthy
```
smoyth: Shouldn't we be checking that `current_user` is not deleted? So I think `expect(current_user).not_to be_destroyed` should work. Or you could combine both of these expectations into `expect(User.all.to_a).to contain_exactly(current_user)`

spec/models/batch_destroy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        destroyer.destroy!
+
+        # users that exist but are not the current user should be deactivated
+        User.where.not(email: "current@user.com").each do |u|
+          expect(u.active).to be_falsey
+        end
+
+        # current user and user with a mission
+        expect(User.count).to eq(2)
+        expect(current_user.active).to be_truthy
```
smoyth: Same thing here. The more exacting test is to check the exact members, not just the count. So `expect(User.all.to_a).to contain_exactly(...)`

spec/models/batch_destroy_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      end
+    end
+
+    describe "question" do
+      let(:questions) { create_list(:question, 3) }
+      let(:batch) { questions }
```
smoyth: This seems unnecessary?

app/controllers/users_controller.rb
```diff
@@ -86,41 +86,18 @@ def destroy
...
-      success <<  t("user.bulk_destroy_deleted", count: destroyed_users.count) unless destroyed_users.empty?
-      success <<  t("user.bulk_destroy_deactivated", count: deactivated_users.count) unless deactivated_users.empty?
-      success <<  t("user.bulk_destroy_skipped", count: skipped_users.count) unless skipped_users.empty?
-      flash[:success] = success.join(" ") unless success.empty?
+    destroyer = BatchDestroy.new(@users, current_user, current_ability)
+    results = destroyer.destroy!
+
+    success = []
+    success <<  t("user.bulk_destroy_deleted", count: results[:destroyed]) if results[:destroyed].positive?
+    success <<  t("user.bulk_destroy_deactivated", count: results[:deactivated]) if results[:deactivated].positive?
```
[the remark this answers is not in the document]
arisah-ijeoma: @smoyth, it felt weird having just this line be in a block

app/controllers/questions_controller.rb
```diff
@@ -66,11 +66,11 @@ def destroy
   def bulk_destroy
     @questions = load_selected_objects(Question)
     destroyer = BatchDestroy.new(@questions, current_user, current_ability)
-    results = destroyer.destroy!
+    out = destroyer.destroy!
```
arisah-ijeoma: var name looks a little weird

would `output` suffice instead?

PR #442, opened 2018-05-17
5996: option creation widget

scttnlsn:


smoyth:
Regarding the drag bug, does a google search turn up anything? See https://stackoverflow.com/questions/28633221/document-body-scrolltop-firefox-returns-0-only-js for instance. I wonder if upgrading JQUI or the plugin would fix it? I will make a separate issue in any case.

scttnlsn:
@smoyth Added some revisions based on your comments

PR #444, opened 2018-05-22
8438 response tree

jpdimond:


app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+
+module Results
+  # Builds a response tree from a form object
+  class ResponseTreeBuilder
+    def initialize(form)
+      @form = form
+    end
+
+    def build
+      @rn = AnswerGroup.new
```
smoyth: Better to not use an abbreviated var name like this. `root` works. Also I don't think it needs to be an ivar.

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+      add_children(@form.root_group.children, @rn)
+      @rn
+    end
+
+    def add_children(form_node, response_node)
```
smoyth: The var name form_node suggests it's a single node but in fact it's an array of nodes. Perhaps refactor so you're actually passing in a node and call children inside the method.

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+      form_node.each do |q|
+        if q.class == QingGroup && q.repeatable?
```
smoyth: It's more typical and less brittle to use `.is_a?` instead of `.class ==`

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+          response_node.children << AnswerGroupSet.new(questioning_id: q.id, new_rank: q.rank)
+          add_repeat_groups(q, response_node.children.last)
```
smoyth: I think it's better to assign the AnswerGroupSet to a var and use that in both the above lines than to use `children.last`. The latter makes me a bit nervous. Sometimes methods like `children` will sort things or do other magic things. Better not to trifle with it.

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+        elsif q.class == QingGroup
+          response_node.children << AnswerGroup.new(questioning_id: q.id, new_rank: q.rank)
+          add_children(q.children, response_node.children.last) if q.children.present?
```
smoyth: Same as above re `last`

smoyth: Also, remember that  `present?` is the same as not blank which is the same as not nil and not empty. There is no need to check both here. The best method to use is `children.any?` I would suggest, which is the opposite of `.empty?`. 

General tip: please pay attention to the subtle and exact meaning of these boolean methods.

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+        elsif q.multilevel?
+          add_multilevel(q, response_node)
+        else
+          response_node.children << Answer.new(questioning_id: q.id, new_rank: q.rank)
+        end
+      end
+    end
+
+    def add_repeat_groups(qing_group, response_node)
+      2.times do |i|
```
smoyth: Only one repeat item is needed plz, so no need for loop.

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+        response_node.children << AnswerGroup.new(questioning_id: qing_group.id, new_rank: i)
+        add_children(qing_group.children, response_node.children.last) if qing_group.children.present?
+      end
+    end
+
+    def add_multilevel(questioning, response_node)
+      response_node.children << AnswerSet.new(questioning_id: questioning.id, new_rank: questioning.rank)
+      questioning.levels.each_with_index do |_level, index|
```
smoyth: You can avoid the unused var by doing `questioning.levels.size.times do |index|`

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+        response_node.children.last.children << Answer.new(
```
smoyth: Same thing here with `.last`

app/models/results/response_tree_builder.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    def add_repeat_groups(qing_group, response_node)
+      2.times do |i|
+        response_node.children << AnswerGroup.new(questioning_id: qing_group.id, new_rank: i)
+        add_children(qing_group.children, response_node.children.last) if qing_group.children.present?
+      end
+    end
+
+    def add_multilevel(questioning, response_node)
```
smoyth: This class is looking really smooth! To make it even awesomer, I'd like you to think a bit more about the names and functions of the methods. This is a way you could get rid of the 'too many lines' hound thing above. Let's discuss.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe "response tree" do
+  # form with repeat group
```
smoyth: Remove?

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe "response tree" do
```
smoyth: For model specs like this you should do `describe Results::ResponseTreeBuilder`

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+  # form with repeat group
+  context "simple form" do
+    let(:form1) { create(:form, question_types: %w[text text text]) }
+
+    it "should produce a simple tree from a form with three children" do
+      response_tree = Results::ResponseTreeBuilder.new(form1).build
+      # same number of leaves at the root
+      expect(response_tree.root_answers.count).to eq form1.root_questionings.count
```
smoyth: Let's not add this root_answers method. It seems like kind of a smell to me. And let's not use root_questionings either. Instead, you can just do `response_tree.children` and `form_tree.children`. (Note that response_tree is really a root node just like root_group).

You then would add like a `let(:form_tree) { form1.root_group }`.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      # expect to be the same qing_ids
+      expect(response_tree.root_answers.map(&:questioning_id))
+        .to include(form1.root_questionings[0].id, form1.root_questionings[1].id,
+          form1.root_questionings[2].id)
```
smoyth: You should actually just do `expect(response_tree.children.map(&:questioning_id)).to eq(form_tree.children.map(&:id))` b/c you know the order should match also.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      # same ranks
+      expect(response_tree.root_answers.map(&:new_rank))
+        .to include(form1.root_questionings[0].rank, form1.root_questionings[1].rank,
+          form1.root_questionings[2].rank)
```
smoyth: Similar thing here.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    end
+  end
+
+  context "forms with groups" do
+    let(:form2) { create(:form, question_types: ["text", %w[select_one text], "text"]) }
```
smoyth: You should just do `let(:form)` throughout. No need to add different suffixes. In fact it's better not to, b/c then the `let(:form_tree)` thing can live **at the top level**!! 💥

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    it "should produce the correct tree" do
+      response_tree = Results::ResponseTreeBuilder.new(form2).build
```
smoyth: This should also live at the top level as a `let` block.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+      puts response_tree.children.inspect
```
smoyth: Baleeted.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      # same number of children at the root
+      expect(response_tree.children.length).to eq form2.root_group.children.length
+
+      # expect to include one answer group
+      expect(response_tree.children.map(&:class)).to include(AnswerGroup)
```
smoyth: You should be more specific here. Like `response_tree.children[1].class` or whatevs.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+      # expect to be the same qing_ids
+      expect(response_tree.children.map(&:questioning_id))
+        .to eq(form2.root_group.children.map(&:id))
+
+      # expect to be the same qing_ids
+      expect(response_tree.children.map(&:new_rank))
+        .to eq(form2.root_group.children.map(&:rank))
+
+      # expect appropriate number of children in the group
+      expect(response_tree.children.descendants.length).to eq 2
```
smoyth: You should be testing the ranks and all the same other stuff in the 2nd level children.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    it "should create the appropriate multilevel answer tree" do
+      response_tree = Results::ResponseTreeBuilder.new(form3).build
+
+      expect(response_tree.children[0].class).to eq AnswerSet
+      expect(response_tree.children[1].class).to eq Answer
+
+      expect(response_tree.children[0].children[0].class).to eq Answer
+      expect(response_tree.children[0].children[1].class).to eq Answer
+      expect(response_tree.children[0].children.length).to eq 2
```
smoyth: Please add and use a `c` alias for stuff like this.

smoyth: Also need to test questionings and ranks here.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    it "should create the appropriate repeating group tree" do
+      response_tree = Results::ResponseTreeBuilder.new(form4).build
+
+      expect(response_tree.children[0].class).to eq AnswerGroupSet
+      expect(response_tree.children[0].children[0].class).to eq AnswerGroup
+      expect(response_tree.children[0].children.length).to eq 2
+      expect(response_tree.children[0].children[0].children.length).to eq 2
+      expect(response_tree.children[0].children[0].children[0].class).to eq Answer
+      expect(response_tree.children[0].children[0].children[1].class).to eq Answer
```
smoyth: Nice. Let's talk about a way to dry this stuff up perhaps.

spec/models/results/response_tree_builder_spec.rb
jpdimond: done

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+        form.children[0].children.map(&:id))
+    end
+  end
+
+  private
+
+  def expect_children(node, types, qing_ids)
+    expect(node.children.map(&:type)).to eq types
+    expect(node.children.map(&:questioning_id)).to eq qing_ids
+    expect(node.children.map(&:new_rank).sort).to eq((1..node.children.size).to_a)
```
smoyth: There is a big potential problem here I am realizing. You need to sort the nodes by rank before you do _all_ the comparisons. Right now if they are out of order you are comparing the wrong things on the first two lines. Let me know if you need me to talk this through with you.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  context "repeat group forms" do
+    let(:form) { create(:form, question_types: ["text", {repeating: {items: %w[text text]}}]) }
+
+    it "should create the appropriate repeating group tree" do
+      response_tree = Results::ResponseTreeBuilder.new(form).build
+
+      expect_children(response_tree, %w[AnswerGroupSet Answer], form.children.map(&:id))
```
smoyth: These are in the wrong order. See comment below on `expect_children`.

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  context "multilevel answer forms" do
+    let(:form) { create(:form, question_types: %w[text multilevel_select_one]) }
+
+    it "should create the appropriate multilevel answer tree" do
+      response_tree = Results::ResponseTreeBuilder.new(form).build
+
+      expect_children(response_tree, %w[AnswerSet Answer], form.children.map(&:id))
```
smoyth: These are in the wrong order. See below comment.

app/models/results/response_tree_builder.rb
jpdimond: done

spec/models/results/response_tree_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      puts form.children[1].children.inspect
```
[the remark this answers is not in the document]
jpdimond: done

spec/models/results/response_tree_builder_spec.rb
jpdimond: done

PR #449, opened 2018-05-31
6800 rescue phone number error

arisah-ijeoma:
This is Phase 1 of the implementation

Displaying a default
screen shot 2018-05-31 at 1 29 51 pm
 error message on the table


app/helpers/sms_helper.rb
```diff
@@ -25,6 +25,12 @@ def format_sms_messages_field(sms, field)
       recips
     when "from" then
       user_with_phone(sms.sender, sms.from)
+    when "body" then
+      content_tag(:span) do
+        "".html_safe <<
+          sms.body <<
+          content_tag(:div, "#{I18n.t('sms.error')}: #{sms.error_message}", class: "error-msg")
```
smoyth: TERRIFIC. Nailed it!

smoyth: Oop! One small thing -- you mustn't add the error message if there is no error, right!?

arisah-ijeoma:
with new styling...

screen shot 2018-06-12 at 7 03 41 pm

app/assets/stylesheets/all/components/_sms.scss
```diff
@@ -21,4 +21,13 @@
...
\ No newline at end of file
+
+  .bold {
+    font-weight: bold;
+  }
+
+  .error-msg {
+    color: $theme-red;
+    font-weight: bold;
+  }
```
smoyth: Perfect.

app/helpers/sms_helper.rb
smoyth: Hmm, and I just learned something new. Rubocop prefers this:

```
out = []
out << content_tag(:li, "one")
out << content_tag(:li, "two")
safe_join(out)
```

So perhaps let's switch to that. Then we don't need `html_safe` at all!

arisah-ijeoma: do you think it would make sense to have a default value of "" in the db? @smoyth

app/controllers/sms_controller.rb
```diff
@@ -75,9 +75,14 @@ def deliver_reply(reply)
       if incoming_adapter.reply_style == :via_adapter
         outgoing_adapter.deliver(reply)
       else # reply via response
-        incoming_adapter.prepare_message_for_delivery(reply)
+        begin
+          raise ENV["STUB_REPLY_ERROR"].constantize if Rails.env.test? && ENV["STUB_REPLY_ERROR"].present?
```
smoyth: `ENV["STUB_REPLY_ERROR"]` contains an error message string. Therefore `ENV["STUB_REPLY_ERROR"].constantize` will raise an error since there is no class named `I am the reply error`. I know that `constantize` was used in the example code I showed you but it's not needed here.

app/controllers/sms_controller.rb
```diff
@@ -75,9 +75,14 @@ def deliver_reply(reply)
...
+          incoming_adapter.prepare_message_for_delivery(reply)
+        rescue StandardError
```
smoyth: It's generally not good to rescue `StandardError`. Can you be more specific about the type of error you want to catch? We know that the adapter may be returning `Sms::Error`. Maybe catch that one?

app/controllers/sms_controller.rb
```diff
@@ -75,9 +75,14 @@ def deliver_reply(reply)
       if incoming_adapter.reply_style == :via_adapter
         outgoing_adapter.deliver(reply)
       else # reply via response
-        incoming_adapter.prepare_message_for_delivery(reply)
+        begin
```
smoyth: Looks like this block is wrapping the wrong code. We had discussed earlier that replying via adapter is the part that might raise the error. So it should wrap the other half of this `if` block. You may need to adjust the test method call to use a different incoming adapter (e.g. Twilio) that has reply_style == `via_adapter`.

app/controllers/sms_controller.rb
```diff
@@ -75,9 +75,14 @@ def deliver_reply(reply)
...
+          raise ENV["STUB_REPLY_ERROR"].constantize if Rails.env.test? && ENV["STUB_REPLY_ERROR"].present?
+          incoming_adapter.prepare_message_for_delivery(reply)
+        rescue StandardError
+          reply.error_message = ENV["STUB_REPLY_ERROR"]
```
smoyth: `ENV["STUB_REPLY_ERROR"]` is just a testing tool. So it doesn't really make sense to be setting it as the error message. This code will be running even in production. I encourage you to think through what is really going on here. If you don't understand let me know.

app/controllers/sms_controller.rb
arisah-ijeoma: left you a message on slack @smoyth

app/controllers/sms_controller.rb
```diff
@@ -73,11 +73,16 @@ def incoming_msg
   def deliver_reply(reply)
     if reply
       if incoming_adapter.reply_style == :via_adapter
-        outgoing_adapter.deliver(reply)
+        begin
+          raise Sms::Error, ENV["STUB_REPLY_ERROR"] if Rails.env.test? && ENV["STUB_REPLY_ERROR"].present?
+          outgoing_adapter.deliver(reply)
+        rescue Sms::Error => e
+          reply.error_message = e
+          reply.save
```
smoyth: Remember: never call `save` by itself unless you're prepared to deal with validation errors. Use `save!` if there shouldn't be any or `save(validate: false)` if you want to skip validation. `save!` should work here.

app/controllers/sms_controller.rb
```diff
@@ -73,11 +73,16 @@ def incoming_msg
   def deliver_reply(reply)
     if reply
       if incoming_adapter.reply_style == :via_adapter
-        outgoing_adapter.deliver(reply)
+        begin
+          raise Sms::Error, ENV["STUB_REPLY_ERROR"] if Rails.env.test? && ENV["STUB_REPLY_ERROR"].present?
+          outgoing_adapter.deliver(reply)
+        rescue Sms::Error => e
+          reply.error_message = e
```
smoyth: This looks much better!

spec/requests/sms/incoming_sms_spec.rb
```diff
@@ -106,6 +106,17 @@
...
   end
 
+  it "should save error message if there is an error in reply" do
+    with_env("STUB_REPLY_ERROR" => "I am the reply error") do
+      assert_sms_response(
+        incoming: {body: "#{form_code} 1.15 2.20", adapter: "FrontlineCloud"},
+        outgoing: {body: "Your response to form '#{form_code}' was received. Thank you!"}
+      )
+      expect(Sms::Reply.count).to eq(1)
+      expect(Sms::Reply.first.error_message).to eq("I am the reply error")
```
smoyth: Very nice!

app/models/sms/adapters/twilio_adapter.rb
```diff
@@ -26,8 +26,8 @@ def deliver(message)
       client = Twilio::REST::Client.new(config.twilio_account_sid, config.twilio_auth_token)
 
       send_message_for_each_recipient(message, client)
-    rescue Twilio::REST::RequestError => e
-      raise Sms::Error.new(e)
+    rescue Twilio::REST::RequestError
+      raise Sms::Error.new("can not send to invalid number")
```
smoyth: This was fine the way it was I believe. Please revert.

PR #456, opened 2018-06-14
8397 key question answers missing in responses index

sapraaman:
Bug fixed in response decorator, added missing spec for response decorator, added spec for response listing


sapraaman:
Got it @smoyth , ya I missed the "key" part completely.

PR #458, opened 2018-06-21
6729 add mission coordinators in reply to

sapraaman:
Add mission coordinators (of the user) in reply-to header when sending the email. If the user does not have any mission coordinator, don’t send reply-to header


app/mailers/notifier.rb
```diff
@@ -1,19 +1,27 @@
...
     @user = user
     build_reset_url(user)
-    mail(to: user.email, subject: t("notifier.welcome", site: Settings.site_name))
+    mail(to: user.email, reply_to: reply_to(user), subject: t("notifier.welcome", site: Settings.site_name))
   end
 
   private
 
+  def reply_to(user)
+    role = "coordinator"
```
smoyth: Please use a symbol here.

sapraaman:
@smoyth have added specs, but instead of feature spec, I have added mailer notifier_spec, since thats what we are testing in this ticket. 

Also lint is passing but facing some rubocop errors (not violations). Not sure what to do about them. Any idea ?

spec/mailers/notifier_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      it "should have all mission coordinator's email in reply-to" do
+        expect(mail.reply_to.length).to eq 3
+        expect(mail.reply_to).to include mission_1_coordinator_1.email
+        expect(mail.reply_to).to include mission_1_coordinator_2.email
+        expect(mail.reply_to).to include mission_2_coordinator.email
+      end
+
+      it "should not have other mission coordinator's email in reply-to" do
+        expect(mail.reply_to).to_not include mission_3_coordinator.email
+      end
```
smoyth: Very nice job on the spec. Thanks!

Note in future if you want to test an array has exact elements in no particular order you should please use `contains_exactly` or `match_array`. This is fine for now.

PR #460, opened 2018-06-25
8562 correct action items positions RTL and LTR

ziedAb:


app/assets/stylesheets/all/components/_form_items.scss
```diff
@@ -99,15 +100,9 @@
     }
 
     .actions {
-      text-align: right;
+      text-align: end;
```
smoyth: This is cool but not sure how well supported:

https://caniuse.com/#feat=css-logical-props

You can use `$right` though. That will be `right` for LTR and `left` for RTL. Try it?

smoyth:
Please request my review again when you've updated these things. Thanks!

ziedAb:
RTL
image


LTR
image

PR #461, opened 2018-06-26
7446 add audio prompt column to question

arisah-ijeoma:


arisah-ijeoma:
screen shot 2018-06-27 at 1 26 35 pm 1

arisah-ijeoma:
By default in the view question page, we have this

screen shot 2018-06-27 at 1 45 29 pm

app/helpers/elmo_form_builder.rb
```diff
@@ -1,7 +1,7 @@
 class ElmoFormBuilder < ActionView::Helpers::FormBuilder
 
   # options[:type] - The type of field to display
-  #   (:text (default), :check_box, :radio_butons, :textarea, :password, :select, :timezone)
+  #   (:text (default), :check_box, :radio_butons, :textarea, :password, :select, :timezone, :file)
```
smoyth: Nice touch.

app/models/question.rb
```diff
@@ -21,6 +21,10 @@ class Question < ApplicationRecord
   has_many :taggings, dependent: :destroy
   has_many :tags, through: :taggings
 
+  has_attached_file :audio_prompt
+  validates_attachment_content_type :audio_prompt, content_type: [%r{\Aaudio/.*\Z}, "video/ogg", "video/webm"]
+  validates_attachment_file_name :audio_prompt, matches: /\.(mp3|ogg|webm|wav)\Z/i
```
smoyth: I just realized we need to check what formats Android supports, and they don't seem to support `webm`, so let's remove that one from here and from the hint please.

app/views/questions/_fields.html.erb
```diff
@@ -28,6 +28,10 @@
     <% end %>
   </div>
 
+  <div>
+    <%= f.field(:audio_prompt, type: :file, hint: t("activerecord.hints.question.audio_prompt")) %>
```
smoyth: You shouldn't need to specify the hint location explicitly. See `option_set_id` above for instance. Remember, [convention over configuration](https://en.wikipedia.org/wiki/Convention_over_configuration)

db/migrate/20180626162616_add_audio_prompt_to_questions.rb
```diff
@@ -0,0 +1,12 @@
...
+# Add audio prompt to question table for paperclip
+class AddAudioPromptToQuestions < ActiveRecord::Migration
+  def up
+    add_attachment :questions, :audio_prompt
+  end
+
+  def down
+    remove_attachment :questions, :audio_prompt
+  end
+end
```
smoyth: Nice and clean.

spec/features/question_form_spec.rb
```diff
@@ -51,4 +51,22 @@
...
     end
   end
+
+  scenario "audio upload works", js: true do
+    visit new_question_path(locale: "en", mode: "m", mission_name: mission.compact_name)
+    fill_in "Code", with: "AQuestion"
+    fill_in "Title", with: "Jay's"
+    select "Text", from: "Type"
+
+    attach_file("Audio Prompt", File.absolute_path("spec/fixtures/media/audio/powerup.mp3"))
```
smoyth: Using `absolute_path` is an interesting technique, but it depends on the specs being run from the correct working directory. In general it's less brittle to use `Rails.root` or better `Rails.root.join` to construct paths like this.

But _even better_ would be a function for retrieving fixture files, which we have I think. See if you can find it.

spec/features/question_form_spec.rb
```diff
@@ -51,4 +51,22 @@
...
+    click_on "Save"
+
+    visit edit_question_path(locale: "en", mode: "m",
+      mission_name: mission.compact_name, id: Question.last.id)
+
+    # on edit, paperclip does not show the audio file on the page even if it exists
+    # but we are certain that it gets saved
+    # perhaps this should be a model spec?
```
smoyth: For now I would say use the `show` page to verify the upload.

But you point out a usability problem -- if there is an existing value, it seems we should be showing it in the edit case. Can you please put that on your TODO list for the issue. We can address it later.

arisah-ijeoma:
screen shot 2018-06-27 at 5 59 22 pm

app/helpers/elmo_form_builder.rb
```diff
@@ -132,7 +132,13 @@ def elmo_field(field_name, options)
...
 
         # get field value
-        val = @object.send(field_name)
+        # audio_prompt is the audio file being set by paperclip
+        # to access the file name, we have to target the original_filename method
+        val = if field_name == :audio_prompt && @object.audio_prompt
+                @object.audio_prompt.original_filename
+              else
+                @object.send(field_name)
+              end
```
smoyth: BTW Jay there is a big problem with this approach. Can you think about it a bit before we meet and see if you can identify the issue?

app/views/questions/_fields.html.erb
```diff
@@ -28,6 +28,10 @@
     <% end %>
   </div>
 
+  <div>
+    <%= f.field(:audio_prompt, type: :file, read_only: cant_update_core) %>
```
smoyth: Another thing to reflect on -- why are you using `cant_update_core` here? I'm not saying it's wrong or right, I just want you to think about the reasons.

spec/features/question_form_spec.rb
```diff
@@ -51,4 +51,25 @@
...
+      mission_name: mission.compact_name, id: Question.last.id)
+
+    expect(page).to have_content("powerup.mp3")
+  end
+
+  private
+
+  def fixture(name)
+    File.expand_path("../../fixtures/media/audio/#{name}", __FILE__)
+  end
```
smoyth: Another question -- where did this code come from? Did you write it or copy it from elsewhere? If the latter, nice find! Does copying code raise any flags for you? Are there other approaches you might have taken?

app/views/questions/_fields.html.erb
arisah-ijeoma: I assumed it to be a field we wouldn't want to change without some form of permission. I looked into the ability file and saw that if the question has been published, it can't be edited

spec/features/question_form_spec.rb
arisah-ijeoma: After going through the codebase, I discovered that this was the code style for the app on expanding fixture files plus it required less code

There are different ways I would have gone about this: using Rails.root and appending it to the path or using join to build the path

app/helpers/elmo_form_builder.rb
arisah-ijeoma: One thing that comes to mind is that this helper is being used in multiple places
Another thing I can think about is that the code is not flexible

smoyth: Both are correct. It's important to think about it in terms of "knowledge" of the class. Metz talks about this a lot. This class shouldn't have to "know" anything about a field called "audio_prompt". That's a "bad coupling". The class is a generic form builder that should work with any form, not just the specific one we're dealing with.

So as a rule, to have something as specific as this in a general utility class is a big red flag.

Can you rewrite it so that it works for _any_ file input?

You know it's safe to do that (won't break any existing functionality) because _you_ added the file field type in this issue. So that is nice!

app/views/questions/_fields.html.erb
smoyth: Nice job checking the ability file! Slight correction though -- if it's published, you can't do something called `update_core` right? Plain `update` is still allowed?

You'll note, for instance, that you can still edit question titles and hints even if the question is published, right?

For this case, since the audio_prompt is kind of like the question title, I think it's ok to allow editing when published.

spec/features/question_form_spec.rb
smoyth: Gotcha. What I'd like you to consider is if you can refactor this into a helper file (maybe general_spec_helpers.rb) so that it doesn't need to be copied into each file that needs to load a fixture.

If you do so, you'll need to change it to the `Rails.root.join` version so that it works no matter what subdirectory the current spec file is in.

arisah-ijeoma: Wouldn't it be better if we just refactored the fixture methods themselves in the tests instead of moving it to a helper. I'm tilting towards this direction because the different files are expanding different directories and I'm thinking how it'd be passing in various arguments

smoyth: You can write a general helper method that takes a sub-path under the fixtures directory and returns the file. There is already a media_fixture method just for media fixtures. That method could call your new method.

app/helpers/elmo_form_builder.rb
```diff
@@ -147,6 +147,9 @@ def elmo_field(field_name, options)
             ""
           end
 
+        when :file
+          options[:value].original_filename
```
smoyth: Getting closer. Try `val` instead of `options[:value]`.

Think about how say, a regular text field works. You don't have to specify a `:value` option. Note that `val` above gets the existing value of the attribute using `send`. Try to look for parallels like this when you're refactoring something.

app/views/questions/_fields.html.erb
```diff
@@ -28,6 +28,10 @@
     <% end %>
   </div>
 
+  <div>
+    <%= f.field(:audio_prompt, type: :file, value: @question.audio_prompt) %>
```
smoyth: Per the above, the `value` part shouldn't be needed.

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -13,10 +13,22 @@ def parse_json(body)
...
+
   def media_fixture(name)
-    path = Rails.root.join("spec/fixtures/media/#{name}")
-    File.open(path)
+    fixture("media", name)
+  end
+
+  def audio_fixture(name)
+    fixture("media/audio", name)
+    # Rails.root.join("spec/fixtures/media/audio/#{name}")
```
smoyth: Remove commented code.

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -13,10 +13,22 @@ def parse_json(body)
...
+  end
+
+  def option_set_fixture(name)
+    fixture("option_set_imports", name)
+  end
+
+  def user_batch_fixture(name)
+    fixture("user_batches", name)
```
smoyth: These helpers look great!!

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -50,4 +62,13 @@ def with_env(vars)
   ensure
     vars.each_pair { |k, _| ENV.delete(k) }
   end
+
+  private
+
+  def fixture(dir, name)
+    path = Rails.root.join("spec/fixtures/#{dir}/#{name}")
```
smoyth: Nice! Note the proper way to use join is `join("spec", "fixtures", dir, name)`. That way you're not dependent on the `/`. But it's not a big deal.

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -50,4 +62,13 @@ def with_env(vars)
...
+
+    # attaching a media file does not require opening it
+    dir == "media/audio" ? path : File.open(path)
```
smoyth: Can you see the parallel here with the "knowledge" thing we discussed earlier? `fixture` is a generic method. It shouldn't "know" about specifics like the `"media/audio"` path. 

I understand you don't want to open the file in that case. Can you think of other ways of solving that problem? Feel free to Slack me some ideas instead of doing a whole new PR>

app/helpers/elmo_form_builder.rb
```diff
@@ -147,6 +147,9 @@ def elmo_field(field_name, options)
             ""
           end
 
+        when :file
+          val.original_filename
```
smoyth: Great. Just change `.` to `&.` in case `val` is nil.

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -13,10 +13,21 @@ def parse_json(body)
...
+  def audio_fixture(name)
+    fixture("media", "audio", name)
+  end
+
+  def option_set_fixture(name)
+    File.open(fixture("option_set_imports", name))
+  end
+
+  def user_batch_fixture(name)
+    File.open(fixture("user_batches", name))
```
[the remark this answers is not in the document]
smoyth: I agree, this is not optimal since you're repeating a lot of logic. How about putting it back the way it was and using `audio_fixture(...).path` to retrieve the file's path in the test?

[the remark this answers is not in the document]
arisah-ijeoma: cool
didn't know I could do that

spec/support/helpers/general_spec_helpers.rb
```diff
@@ -50,4 +61,18 @@ def with_env(vars)
   ensure
     vars.each_pair { |k, _| ENV.delete(k) }
   end
+
+  # decorate objects
+  def decorate(obj)
+    Odk::DecoratorFactory.decorate(obj)
+  end
```
smoyth: I love the initiative you've taken here, to move this into the general helpers file!

There is an issue here though. In fact two issues. 

1. There is an issue with naming specificity/scope. What is it?
2. There is an issue with cohesion. See https://en.wikipedia.org/wiki/Cohesion_(computer_science). In general, files like 'general ____' are a bad idea. We should probably get rid of this one even, because it tends to attract the kitchen sink (i.e. a barely cohesive set of methods that do all kinds of things). This is related to the fat models thing. Fat nothing, skinny everything! "Fatness" generally means lower cohesion, too much responsibility, etc. A better way to solve the problem of "where to put supporting methods for specs" is to use contexts. In this case, perhaps a context called "odk rendering". Are there other methods in the general file we could move in there? (I know there is an odk submission context but I don't want to mess with that one because Devney's branch is heavily modifying it, and also I think the rendering stuff is separate from the submission stuff). Remember that once you define the context, you need to include it in each place it's used. See existing code for how to do that.

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+  describe FormDecorator, :odk, :reset_factory_sequences do
+    describe "needs_manifest?", :odk do
+      context "for form with single level option sets only" do
+        before { @form = create(:form, question_types: %w(select_one)) }
+        it "should return false" do
+          decorated_form = decorate(@form)
+          expect(decorated_form.needs_manifest?).to be false
+        end
+      end
+      context "for form with multi level option set" do
```
smoyth: Please add a blank line between context blocks.

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+# Form decorator spec
+module Odk
+  describe FormDecorator, :odk, :reset_factory_sequences do
+    describe "needs_manifest?", :odk do
```
smoyth: Since the outer block has `:odk` on it already, the inner one doesn't need it.

Also, why did you add `:reset_factory_sequences`?

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      context "for form with single level option sets only" do
+        before { @form = create(:form, question_types: %w(select_one)) }
```
smoyth: Can you please refactor this to use `let` instead of `before`/ivars?

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+        it "should return false" do
+          decorated_form = decorate(@form)
```
smoyth: Notice the duplication here and below. That is a big sign to use a `let` block in the parent context. Please refactor.

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+# Form decorator spec
+module Odk
```
smoyth: I don't think we use modules like this in specs. You can just do `describe Odk::FormDecorator`

spec/decorators/odk/form_decorator_spec.rb
arisah-ijeoma: `reset_factory_sequences` reloads the factories

spec/decorators/odk/condition_group_decorator_spec.rb
```diff
@@ -2,7 +2,9 @@
 
 module Odk
   describe ConditionGroupDecorator, :odk, :reset_factory_sequences, database_cleaner: :truncate do
-    let(:result) { Odk::DecoratorFactory.decorate(condition_group).to_odk }
+    include_context "odk rendering"
```
smoyth: Woop woop!

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+# Form decorator spec
```
smoyth: This is not necessary since it's obvious from the next line. Only comment things that aren't obvious. Prefer descriptive code where possible, and that's what you have below. :)

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+describe Odk::FormDecorator, :odk, :reset_factory_sequences do
+  include_context "odk rendering"
+
+  describe "needs_manifest?" do
+    let(:decorated_form) { decorate(form) }
```
smoyth: Perfect!

smoyth: Now here's another fun thing!

This is a great place to use `subject` and the one-liner syntax for expect.

Up here, put `subject(:needs_manifest) { decorated_form.needs_manifest? }`

Then below, instead of the current `it` block, try `it { is_expected.to be false }`

Sweet!

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      let(:form) { create(:form, question_types: %w[select_one]) }
+      it "should return false" do
+        expect(decorated_form.needs_manifest?).to be_falsey
+      end
+    end
+
+    context "for form with multi level option set" do
+      let(:form) { create(:form, question_types: %w[select_one multilevel_select_one]) }
+      it "should return true" do
+        expect(decorated_form.needs_manifest?).to be_truthy
```
smoyth: Question: what is the difference between `be true` and `be_truthy`? Given that, which is more appropriate here? Why?

spec/support/contexts/odk_rendering.rb
```diff
@@ -0,0 +1,8 @@
+# odk rendering helper
```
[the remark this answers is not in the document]
smoyth: This is another comment you don't need. See my above tip.

spec/support/contexts/odk_rendering.rb
```diff
@@ -0,0 +1,8 @@
...
+shared_context "odk rendering" do
+
+  # decorate objects
```
smoyth: Same here. `decorate(obj)` is already a good cue that this method decorates things.

spec/decorators/odk/form_decorator_spec.rb
arisah-ijeoma: alrighty
Was trying to avoid rubocop lint warning

spec/decorators/odk/form_decorator_spec.rb
arisah-ijeoma: if the expectation expression is returning a boolean, you can append an underscore to `be` and have things like be_truthy, be_falsey, be_valid, be_any_boolean_return_value

after doing some googling, it's said that be_truthy passes when the expectation is neither nil nor false while be true uses a `==` to do its comparison

Is there a preference for the project?

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+            form.c[0].update!(hidden: true)
+          end
+
+          it { is_expected.to be false }
+        end
+      end
+
+      context "without audio prompt question" do
+        it { is_expected.to be false }
+      end
```
smoyth: Perfect! Bravo!

smoyth: One small thing -- all the inner `describe`s should be `context`s. See https://lmws.net/describe-vs-context-in-rspec

"The purpose of “describe” is to wrap a set of tests against one functionality"
"a “context” is to setup the state of your object"

Usually I just do one `describe` per method I'm testing, and the rest from there on in are `context`.

app/decorators/odk/form_decorator.rb
```diff
@@ -45,5 +45,17 @@ def ir_code_binding
...
+    # Whether this form needs an accompanying manifest for odk.
+    def needs_manifest?
+      # For now this is IFF there are any multilevel option sets or questions with audio prompts
+      @needs_manifest ||= option_sets.any?(&:multilevel?) || audio_prompt_content?
+    end
+
+    private
+
+    def audio_prompt_content?
+      smsable_questionings.values.any?(&:audio_prompt_content_type)
```
smoyth: Good attempt! 

I'm curious why you used `smsable_questionings` here. I'm guessing b/c it turns out there is no `visible_questionings` method and this was the closest thing? In cases like this you should please ask me rather than guess. `smsable_questionings` won't do here because it would exclude questions we don't want to exclude.

Check out the `questions` method in `Form`. Then you can also get rid of the delegator.

Also, there is a cleaner way to test if a Paperclip attachment is present:

https://makandracards.com/makandra/1419-check-whether-a-paperclip-attachment-exists

Using`audio_prompt_content_type` is not a bad idea but it's purpose is not as understandable as the above.

app/views/forms/odk_manifest.xml.erb
```diff
@@ -7,5 +7,13 @@
         <downloadUrl><%= url_for_path(@ifa.path) %></downloadUrl>
       </mediaFile>
     <% end %>
+
+    <% @questions.each do |q| %>
+      <mediaFile>
+        <filename><%= q.scramble_audio_prompt_filename %></filename>
+        <hash><%= Digest::MD5.file(q.audio_prompt.path).hexdigest %></hash>
```
smoyth: Nice! This should also go in an ODK decorator. Perhaps `q.audio_prompt_md5`.

app/models/question.rb
```diff
@@ -244,6 +248,12 @@ def sorted_tags
     tags.order(:name)
   end
 
+  # Make audio prompt file name unique to curb collisions and duplications
+  def scramble_audio_prompt_filename
+    names = audio_prompt_file_name.split(".")
+    names.first + "_#{id.first(5)}.#{names.last}"
+  end
```
smoyth: This is pretty neat! 

I think you can simplify it more though. What about just 

```
"#{id}_audio_prompt#{File.extname(audio_prompt_file_name)}"
```

It's better to use `extname` since it's designed for this purpose. Also this method should go in a decorator in the ODK namespace since it's view logic.

Also as a rule, try to name your methods as nouns instead of verbs. So `scrambled_audio_prompt_filename` or `unique_audio_prompt_filename` might be better. (When I think of scrambling I think of mixing up the letters of something)

spec/requests/odk_form_download_spec.rb
```diff
@@ -85,6 +87,33 @@
...
+          expect(response).to be_success
+          assert_select("mediaFile", count: 2)
+
+          assert_select "manifest" do |elements|
+            elements.each do |element|
+              assert_select(element, "filename", text: "powerup_#{form.c[0].question.id.first(5)}.mp3")
+              assert_select(element, "filename", text: "powerup_#{form.c[1].question.id.first(5)}.wav")
+
+              assert_select(element, "hash", text: "e7fe3aa406b8b67209b9d89c0cd50aa8")
+              assert_select(element, "hash", text: "ff2fd1e209465c5bffa784b5c57d84c4")
```
smoyth: This is close but not quite. Good job finding the alternative `assert_select` signature (with `element` first). But this setup does not ensure that the right md5 is with the right filename inside the mediaFile tags.

Can you set it up a bit differently so that it does? Hint: you may not need to iterate. Hint 2: There is only one `manifest` element.

Nice job getting the md5s working!

app/models/question.rb
arisah-ijeoma: 👍

arisah-ijeoma: I initially didn't put it in the decorator because I didn't want to initialize it with the form in the view

app/views/forms/odk_manifest.xml.erb
arisah-ijeoma: 👍

app/decorators/odk/form_decorator.rb
```diff
@@ -45,5 +45,20 @@ def ir_code_binding
...
     end
+
+    # Whether this form needs an accompanying manifest for odk.
+    def needs_manifest?
+      # For now this is IFF there are any multilevel option sets or questions with audio prompts
+      visible_questionings.any? { |q| q.multilevel? || q.audio_prompt.exists? }
+    end
+
+    # Make audio prompt file name unique to curb collisions and duplications
+    def unique_audio_prompt_filename(q)
```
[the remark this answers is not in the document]
smoyth: I think this method belongs in a _Question_ decorator, not the Form decorator. I don't think we have a question decorator yet but you can make one.

If you have a method like this that takes an argument and everything in the method depends on that argument, it's useful to ask if the method could just be a member method on that argument itself (or a decorator of it).

app/decorators/odk/form_decorator.rb
```diff
@@ -45,5 +45,20 @@ def ir_code_binding
...
+      "#{q.id}_audio_prompt#{File.extname(q.audio_prompt_file_name)}" if q.audio_prompt_file_name
+    end
+
+    def audio_prompt_md5(q)
+      Digest::MD5.file(q.audio_prompt.path).hexdigest if q.audio_prompt_file_name
+    end
```
smoyth: Same here.

app/views/forms/odk_manifest.xml.erb
```diff
@@ -7,5 +7,15 @@
         <downloadUrl><%= url_for_path(@ifa.path) %></downloadUrl>
       </mediaFile>
     <% end %>
+
+    <% unless @questions.empty? %>
+      <% @questions.each do |q| %>
```
smoyth: Note you can decorate all of these at once with `DecoratorFactory.decorate_collection`

arisah-ijeoma: whao!

app/controllers/forms_controller.rb
```diff
@@ -105,6 +105,8 @@ def odk_manifest
     authorize!(:download, @form)
     @cache_key = @form.odk_download_cache_key
     unless fragment_exist?(@cache_key)
+      questions = @form.visible_questionings.collect(&:question).select(&:audio_prompt_file_name)
```
smoyth: Always use `map` instead of `collect` please.

Also note that this is only working because `audio_prompt_file_name` (a string) is 'truthy'. But it doesn't read very cleanly. On first inspection it looks like you're assembling an array of strings but really it's an array of questions. Can you refactor for clarity?

app/controllers/questions_controller.rb
```diff
@@ -77,6 +77,17 @@ def bulk_destroy
     redirect_to(index_url_with_context)
   end
 
+  def audio_prompt
+    decorated_question = QuestionDecorator.decorate(@question)
+
+    authorize! :show, decorated_question
```
smoyth: Better to use the undecorated object here since you have it available.

app/controllers/questions_controller.rb
```diff
@@ -77,6 +77,17 @@ def bulk_destroy
...
+
+    send_file decorated_question.audio_prompt.path,
+      type: decorated_question.audio_prompt_content_type,
+      disposition: "attachment",
+      filename: decorated_question.unique_audio_prompt_filename
```
smoyth: Nice work here!

app/decorators/question_decorator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true.
+
+class QuestionDecorator < ApplicationDecorator
```
smoyth: I didn't catch this earlier but this should be namespaced to `Odk` since it's only related to ODK stuff. You can/should then also use the ODK::DecoratorFactory thing to instantiate.

app/views/forms/odk_manifest.xml.erb
```diff
@@ -7,5 +7,15 @@
         <downloadUrl><%= url_for_path(@ifa.path) %></downloadUrl>
       </mediaFile>
     <% end %>
+
+    <% unless @decorated_questions.empty? %>
+      <% @decorated_questions.each do |question| %>
+        <mediaFile>
+          <filename><%= question.unique_audio_prompt_filename %></filename>
+          <hash><%= question.audio_prompt_md5 %></hash>
+          <downloadUrl><%= url_for_path(audio_prompt_question_path(question)) %></downloadUrl>
```
smoyth: I think you can just use `audio_prompt_question_url(question)`.

Remember, when pasting code, always scrutinize every piece -- is this still relevant? Do I need this? Does it work in this context?

spec/decorators/odk/condition_decorator_spec.rb
```diff
@@ -73,11 +75,11 @@
...
       let(:form) { form = create(:form, question_types: %w(integer text date time datetime text)) }
-      let(:int_q) { Odk::DecoratorFactory.decorate(form.questionings[0]) }
-      let(:text_q) { Odk::DecoratorFactory.decorate(form.questionings[1]) }
-      let(:date_q) { Odk::DecoratorFactory.decorate(form.questionings[2]) }
-      let(:time_q) { Odk::DecoratorFactory.decorate( form.questionings[3]) }
-      let(:datetime_q) { Odk::DecoratorFactory.decorate(form.questionings[4]) }
+      let(:int_q) { decorate(form.questionings[0]) }
+      let(:text_q) { decorate(form.questionings[1]) }
+      let(:date_q) { decorate(form.questionings[2]) }
+      let(:time_q) { decorate( form.questionings[3]) }
```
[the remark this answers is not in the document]
smoyth: Please fix this.

config/initializers/paperclip.rb
```diff
@@ -1,6 +1,6 @@
 Paperclip::Attachment.default_options.merge!(
   path: ":rails_root/uploads/:class/:attachment/:id_partition/:style/:filename",
-  url: "/:locale/m/:mission/:class/:id/:style"
+  url: "/:locale/m/:mission/:class/:attachment/:id/:style"
```
smoyth: This is actually not being used now, right? It can't be because it's incorrect!

This only gets used if we do `question.audio_prompt.url` but we're not doing that. So I'd say delete this line.

spec/requests/odk_form_download_spec.rb
```diff
@@ -85,6 +87,54 @@
...
+            get("/#{audio_prompt_question_path(form.c[0].question)}")
+            expect(response).to be_success
+            expect(response.header["Content-Disposition"]).to include(form.c[0].question.id)
+          end
+
+          it do
+            get("/#{audio_prompt_question_path(form.c[1].question)}")
+            expect(response).to be_success
+            expect(response.header["Content-Disposition"]).to include(form.c[1].question.id)
+          end
```
smoyth: Very nice work here! I think this is good to go!

For checking the body, did you try `expect(response.body).to eq audio_fixture("powerup.mp3").read`?

app/decorators/question_decorator.rb
arisah-ijeoma: thanks

app/views/forms/odk_manifest.xml.erb
arisah-ijeoma: I didn't try it without the url_for_path helper but didn't just add it. I checked the method and noticed it worked with the http(s) protocols. That's why I used it but using the `_url` helper is...nicer

spec/requests/odk_form_download_spec.rb
arisah-ijeoma: I did. Not in the same syntax tho and I got similar content but the special characters where different. One was using `u0003` and the other was using `x88`

app/views/questions/_fields.html.erb
```diff
@@ -28,10 +28,14 @@
...
 
+  <div>
+    <%= f.field(:audio_prompt, type: :file) %>
+  </div>
+
   <div class="minmax">
     <%= f.field(:casted_minimum, :partial => 'questions/min_max_condition') %>
     <%= f.field(:casted_maximum, :partial => 'questions/min_max_condition') %>
-  </div>
+  </div >
```
smoyth: Please fix this.

app/views/forms/odk/_translation.xml.erb
```diff
@@ -11,6 +11,11 @@
       <% qing.subqings.each do |subq| %>
         <text id="<%= subq.odk_code %>:label">
           <value><%= subq.name(lang, :strict => false) %></value>
+          <% if subq.audio_prompt.exists? %>
+            <value form="audio">
+              jr://audio/<%= Odk::DecoratorFactory.decorate(subq.question).unique_audio_prompt_filename %>
```
smoyth: :heart:

config/initializers/local_config.rb.example
```diff
@@ -9,6 +9,9 @@ configatron.offline_mode = false
 # And copy and paste into this file in place of `secret-token`.
 ELMO::Application.config.secret_key_base = "secret-token"
 
+# raise error if host or protocol is not set
+raise "URL protocol or host is not set" if (%i[protocol host] - configatron.url.keys).any?
```
smoyth: Let's drop this for now. Been at this long enough! :)

spec/requests/odk/form_download_spec.rb
```diff
@@ -32,28 +31,28 @@
...
       it "should succeed" do
         # XML rendering details are tested elsewhere.
-        get("/en/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml})
+        get("/en/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml}, headers: basic_auth)
         expect(response).to be_success
       end
 
       it "should succeed with no locale" do
-        get("/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml})
+        get("/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml}, headers: basic_auth)
```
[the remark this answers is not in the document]
arisah-ijeoma: fixed

spec/requests/odk/form_download_spec.rb
```diff
@@ -32,28 +31,28 @@
...
+        get("/m/#{mission.compact_name}/formList", params: {format: :xml}, headers: basic_auth)
         expect(response).to be_success
       end
     end
 
     describe "showing forms" do
       it "should succeed" do
         # XML rendering details are tested elsewhere.
-        get("/en/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml})
+        get("/en/m/#{mission.compact_name}/forms/#{form_multiselect.id}", params: {format: :xml}, headers: basic_auth)
```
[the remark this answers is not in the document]
arisah-ijeoma: fixed

PR #465, opened 2018-07-09
8538 response form tree

scttnlsn:


scttnlsn:
@smoyth Should I address all the linting issues in the duplicated controller?

smoyth:
No that's ok @scttnlsn

app/controllers/hierarchical_responses_controller.rb
```diff
@@ -0,0 +1,350 @@
...
+          files = params.select { |k, v| v.is_a? ActionDispatch::Http::UploadedFile }
+          files.each { |k, v| files[k] = v.tempfile }
+
+          unless upfile
+            render_xml_submission_failure("No XML file attached.", 422)
+            return false
+          end
+
+          @response.awaiting_media = true if params["*isIncomplete*"] == "yes"
+          @submission = XMLSubmission.new response: @response, files: files, source: "odk"
```
DevneyHamilton: Was this branch rebased on develop after 6441 was merged? This method has changed a lot.

scttnlsn: I don't think so.  This controller is a duplicate of the `ResponsesController` but with a few things changed to keep it working with a `Response` model (as opposed to a `HierarchicalResponse` model).  Perhaps I'll rebase against `develop` and copy this action over again?

smoyth: @scttnlsn That sounds like a good idea. Has it changed at all?

smoyth: Might be a good idea to re-copy any other files that changed in Devney's issue also...

scttnlsn: Did the rebase and then recreated my changes to a duplicated responses controller.  What other files are you referring to @smoyth?

app/models/answer_set.rb
```diff
@@ -2,4 +2,21 @@
 
 # Corresponds with a multilevel questioning. An AnswerSet's parent is an AnswerGroup. Its children are Answers
 class AnswerSet < ResponseNode
+  belongs_to :questioning
+
+  alias answers c
```
smoyth: Alias this to `children` instead of `c`? They both exist and it's a bit more descriptive...

app/models/answer_group.rb
```diff
@@ -4,4 +4,5 @@
 # An AnswerGroups's parent is an AnswerGroupSet or an AnswerGroup.
 # Its children can be Answers, AnswerSets, AnswerGroups, or AnswerGroupSets.
 class AnswerGroup < ResponseNode
+  belongs_to :questioning, class_name: "QingGroup"
```
smoyth: Name this association `qing_group` instead? We eventually need to refactor the db so that the col name is `form_item_id` instead of `questioning_id` anyway...

app/models/results/blank_response_tree_builder.rb
```diff
@@ -3,37 +3,41 @@
...
-      root = AnswerGroup.create!(form_item: form.root_group, response: response)
+      root = AnswerGroup.new(form_item: form.root_group, response: response)
+      root.save! if options[:save]
       add_level(form.root_group.sorted_children, root)
       response.associate_tree(root)
 
       # TODO: We can remove the `validate: false` once various validations are
       # removed from the response model
-      response.save(validate: false)
+      response.save(validate: false) if options[:save]
```
smoyth: I think you can get rid of save altogether since we gave up on that approach.

You just need to change one thing to keep specs from breaking:

```
-      root = AnswerGroup.create!(form_item: form.root_group, response: response)
+      root = AnswerGroup.create!(form_item: form.root_group, response: response, new_rank: 0)
```

spec/models/results/blank_response_tree_builder_spec.rb
```diff
@@ -6,7 +6,7 @@
   include_context "response tree"
 
   let(:response) { create(:response, form: form, answer_values: nil) }
-  let(:response_tree) { Results::BlankResponseTreeBuilder.new(response).build }
+  let(:response_tree) { Results::BlankResponseTreeBuilder.new(response, save: true).build }
```
smoyth: Can get rid of this with the change I suggested above.

spec/features/results/response_form_tree_handling_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+          ".answer-group", ".answer-group-set",
+          ".answer-group", ".answer-group-set", ".answer-group", ".answer .ro-val"
+        ], visible: false)
+      end
+    end
+  end
+
+  def expect_path(path, options = {})
+    selector = path.join(" .children ")
+    expect(page).to have_selector(selector, options)
```
smoyth: Ooh cute!

PR #472, opened 2018-07-31
8680 export option set bug

arisah-ijeoma:


spec/features/option_set_flow_spec.rb
```diff
@@ -52,6 +52,11 @@
     add_options(%w[Pear])
     click_button("Save")
     expect(page).to have_selector("td.options_col div", text: "Banana, Apple, Pear")
+
+    # Test export page does not error out
```
DevneyHamilton: Would be good to add 'export' to the description of the scenario as well, e.g. "creating, showing, editing, and exporting"

spec/requests/option_set_spec.rb
```diff
@@ -3,15 +3,15 @@
 # Using request spec b/c Authlogic won't work with controller spec
 describe "option_sets", type: :request do
   let(:user) { create(:user, role_name: "coordinator") }
+  let(:option_set) { create(:option_set, super_multilevel: multilevel) }
```
DevneyHamilton: nice :)

spec/requests/option_set_spec.rb
```diff
@@ -183,6 +183,11 @@
     end
   end
 
+  it "exports successfully" do
```
DevneyHamilton: Thanks for adding this spec too

Gemfile.lock
```diff
@@ -124,9 +124,9 @@ GEM
     autoprefixer-rails (8.6.3)
       execjs
     awesome_print (1.6.1)
-    axlsx_rails (0.3.0)
+    axlsx_rails (0.5.2)
```
DevneyHamilton: cool fix :)

spec/requests/option_set_spec.rb
arisah-ijeoma: ooh thanks for the catch

spec/features/option_set_flow_spec.rb
arisah-ijeoma: oooh thanks for the catcg

spec/features/option_set_flow_spec.rb
```diff
@@ -52,6 +52,11 @@
...
+    click_on "Foo"
+    click_link("Edit Option Set")
+    click_on "Export Option Set"
```
smoyth: Great to have this added!

spec/requests/option_set_spec.rb
smoyth: It is nice to be thorough but I don't think we actually need this as it simply duplicates the feature spec unless I'm missing something...

I think our goal is to use feature specs where possible and request specs only where there is no browser-accessible UI to exercise. So unless I'm off base here I think we should remove this spec and just leave the feature spec for the sake of simplicity.

PR #473, opened 2018-07-31
8707 response factory builds answer tree

DevneyHamilton:


app/models/results/web_response_parser.rb
```diff
@@ -4,6 +4,7 @@ module Results
   # Builds (does not save) an answer tree based on answer data in a web response.
   class WebResponseParser
     TOP_LEVEL_PARAMS = %i[
+      type
```
DevneyHamilton: Changes to the web response parser are wrapped up in 8646; they are for updating web responses (that work prompted 8707 refactor of response factory).

spec/support/helpers/model_spec_helpers.rb
```diff
@@ -1,8 +1,10 @@
 module ModelSpecHelpers
-  def expect_location_answer(answer, params)
+  def expect_location_answer(response, answer, params)
+    response.save
     answer.save
+
```
DevneyHamilton: I'm not certain about this response.save and answer.save, and what it means for how we're handling saving responses and the response tree. If I remove answer.save, then spec/models/results/geopoint_processing_spec.rb:223 fails (and no other specs fail). It updates a location with less data.

spec/support/helpers/model_spec_helpers.rb
```diff
@@ -1,8 +1,10 @@
...
     # If the answer is invalid there is no need to check if values were copied correctly.
-    return if answer.invalid?
+    return if response.invalid?
```
DevneyHamilton: Also would like to go over what this means for validation flow.

app/models/form_item.rb
```diff
@@ -208,6 +208,19 @@ def group?
...
+    chunks = []
+    chunks << " " * (indent * 2)
+    chunks << rank.to_s.rjust(2)
+    chunks << " "
+    chunks << self.class.name.ljust(15)
+    chunks << " Type: #{qtype_name}," if qtype_name.present?
+    chunks << " Code: #{code}"
+    chunks << " Repeatable" if repeatable?
+    "\n#{chunks.join}#{child_tree}"
+  end
```
smoyth: Woop woop for debug methods!

app/models/response_node.rb
```diff
@@ -8,6 +8,10 @@ class ResponseNode < ApplicationRecord
   belongs_to :response
   has_closure_tree order: "new_rank", numeric_order: true, dont_order_roots: true, dependent: :destroy
 
+  before_save { children.each { |c| c.response_id = response_id } }
```
smoyth: Can we move this to a method? I prefer methods since it's a chance to give a nice juicy name to what you're doing.

app/models/response_node.rb
```diff
@@ -28,4 +32,15 @@ def debug_tree(indent: 0)
   def casted_value
     nil
   end
+
+  #_destroy defaults to false unless set otherwise
```
smoyth: Space after #

spec/factories/responses.rb
```diff
@@ -1,65 +1,169 @@
...
+# Answer                      10
+#                           ],
+# Inner rpt grp instance2   [
+# Answer                      "b",
+# Answer                      11
+#                           ]
+#                        ]}
+#                       ]
+#                     ]}
+#                    ]
```
smoyth: Nice big juicy comment!

spec/decorators/answer_arranger_spec.rb
```diff
@@ -153,10 +151,10 @@
     let(:response) do
       create(:response, form: form, answer_values: [
         123,
-        [:repeating,
+        {repeating: [
```
smoyth: Nice to harmonize this!

spec/factories/responses.rb
```diff
@@ -1,65 +1,169 @@
...
+  # form_group may or may not be repeating
+  # answer_values is an array of answers or groups in this group
+  # parent must be an AnswerGroup (if form item not repeating) or AnswerGroupSet (if form item is repeating)
+  def self.add_group(form_group, answer_values, parent)
+    answer_group = parent.children.build({type: "AnswerGroup", form_item: form_group}
+      .merge(rank_attributes("AnswerGroup", parent)))
+    add_level(form_group.c, answer_values, answer_group)
+  end
+
+  # value may be integer, string, object, or (for select one or multilevel only) an array of strings
```
smoyth: Should this be `select multiple or multilevel only`

spec/factories/responses.rb
smoyth: Should we put something about blank string vs. nil? Technically nil is a valid answer value and just means no object gets created? So nil implies irrelevant while "" implies relevant but blank?

spec/models/report/summary_collection_single_spec.rb
```diff
@@ -258,8 +261,8 @@ def prepare_form_and_collection(*args)
   end
 
   def prepare_form(qtype, answers, options = {})
-    @form = create(:form, {:question_types => [qtype], :option_names => %w(Yes No)}.merge(options))
-    @responses = answers.map { |a| create(:response, :form => @form, :answer_values => [a]) }
+    @form = create(:form, {question_types: [qtype], option_names: %w[Yes No]}.merge(options))
```
smoyth: Thanks for all these cleanups!

spec/models/response_tree_construction_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  context "tree includes invalid node" do
+    let(:int_answer) { "invalid" }
+    it "errors with validation error on invalid node" do
+      expect { response.save! }.to raise_error(ActiveRecord::RecordInvalid)
+      expect(group.c[0].errors[:value].join).to match("Please enter a valid number")
+    end
+  end
+end
```
smoyth: I ❤️ this spec

spec/models/results/answer_location_spec.rb
```diff
@@ -152,8 +153,8 @@
       it "copies the coordinates from the option to the answer lat/long columns" do
         questioning.option_set.update!(geographic: true, allow_coordinates: true)
         option.update!(latitude: 12.34, longitude: -56.78)
-        expect(response.answers[0].latitude).to eq(12.34)
-        expect(response.answers[0].longitude).to eq(-56.78)
+        expect(answer.latitude).to eq(12.34)
+        expect(answer.longitude).to eq(-56.78)
```
smoyth: Much cleaner!

spec/models/results/web_response_parser_spec.rb
```diff
@@ -7,237 +7,527 @@
...
+        end
+      end
     end
-    let(:instance_two_answers) do
-      {
-        "0" => web_answer_hash(form.c[1].c[0].id, value: "D"),
-        "1" => web_answer_hash(form.c[1].c[1].id, value: "E")
-      }
+
+    context "repeat groups" do
```
smoyth: Do we still need this one if we are covering the nested repeat case? This file is huge and (unavoidably) kind of hard to parse so if we can shrink it a bit that might be nice.

spec/support/helpers/model_spec_helpers.rb
smoyth: Yeah we should check into this.

app/models/response_node.rb
```diff
@@ -28,4 +32,15 @@ def debug_tree(indent: 0)
...
+  def _destroy
+    @destroy.nil? ? false : @destroy
+  end
+
+  # A flag indicating whether this node should be destroyed before save.
+  # convert string 'true'/'false' to boolean
+  def _destroy=(d)
+    @destroy = d.is_a?(String) ? d == "true" : d
```
smoyth: Where are we taking care of deleting the node if relevant is false?

app/models/response.rb
```diff
@@ -26,8 +28,11 @@ class Response < ApplicationRecord
   friendly_id :shortcode
 
   before_save :normalize_answers
+  after_save { root_node.save if root_node.present? }
```
DevneyHamilton: I deleted autosave because it wasn't having any effect visible in specs.

DevneyHamilton: Autosave is replaced by this and an after_save in response_node.rb

spec/models/results/web_response_parser_spec.rb
DevneyHamilton: I think it'll be nice to have separate specs if there's an issue with nested in particular. I agree it's an unfortunately large spec.

app/models/response_node.rb
DevneyHamilton: That comes in 8663, integrating the controller and actually saving the response.

spec/models/results/web_response_parser_spec.rb
DevneyHamilton: For parsing's sake, we could move the update specs to another file. That would make it less daunting and easier to parse the lets and context blocs.

PR #477, opened 2018-08-10
7460 duplicate mission user in db

arisah-ijeoma:


db/migrate/20180810162637_delete_assignments_with_duplicate_mission_and_user.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class DeleteAssignmentsWithDuplicateMissionAndUser < ActiveRecord::Migration[5.1]
+  def up
+    duplicates = []
+    dup_attrs = Assignment.select(:mission_id, :user_id).group(:mission_id, :user_id).having("count(*) > 1")
+
+    dup_attrs.each do |dup|
+      dup_assignments = Assignment.where(mission_id: dup.mission_id, user_id: dup.user_id)
+      duplicates << dup_assignments.first
```
smoyth: This is close! But it assumes there are only two duplicates. What if there are 50?

You can grab _all but the first_ elements in an array with `array[1..-1]`. Note that you'll then have an array, not a single object.

db/migrate/20180810162637_delete_assignments_with_duplicate_mission_and_user.rb
```diff
@@ -0,0 +1,18 @@
...
+    end
+
+    duplicates.map(&:destroy)
```
smoyth: Two things here:

1. Why not do this in the loop? Then you don't need an extra instance variable and the code is simpler.
2. Only use `map` if you intend to do something with the result. Use `each` otherwise. If you're not familiar with the difference, please read up and practice. Really important methods, both!

db/migrate/20180810170823_add_composite_index_to_assignment_columns.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddCompositeIndexToAssignmentColumns < ActiveRecord::Migration[5.1]
+  def change
+    add_index :assignments, %i[deleted_at mission_id user_id], unique: true
```
smoyth: ✅

db/migrate/20180810162637_delete_assignments_with_duplicate_mission_and_user.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class DeleteAssignmentsWithDuplicateMissionAndUser < ActiveRecord::Migration[5.1]
+  def up
+    duplicates = []
+    dup_attrs = Assignment.select(:mission_id, :user_id).group(:mission_id, :user_id).having("count(*) > 1")
```
smoyth: Nice use of `group` and `having`!

db/migrate/20180810162637_delete_assignments_with_duplicate_mission_and_user.rb
arisah-ijeoma: thanks

db/migrate/20180810162637_delete_assignments_with_duplicate_mission_and_user.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    dup_attrs.each do |dup|
+      dup_assignments = Assignment.where(mission_id: dup.mission_id, user_id: dup.user_id)
+      duplicates << dup_assignments[1..-1]
+      duplicates.flatten.each(&:destroy)
```
smoyth: The way this is written it will destroy the first element over and over again.

You can simply do `Assignment.where(mission_id: dup.mission_id, user_id: dup.user_id)[1..-1].each(&:destroy)` or some such.

PR #480, opened 2018-08-15
 8664 conditional logic

scttnlsn:


app/views/hierarchical_responses/answers/_nodes.html.erb
```diff
@@ -19,4 +26,20 @@
       <%= render("hierarchical_responses/answers/group_set", group_set: node, context: new_context) %>
     <% end %>
   </div>
+
+  <%# Setup condition js. %>
+  <%# Note that we need to do this in any form mode, even show. %>
```
smoyth: This is no longer true. Can you set a condition here only set this up on `new`? `edit` will come later, and `show` will never come :D

app/assets/javascripts/views/hierarchical_response_condition_checker.js.coffee
```diff
@@ -0,0 +1,133 @@
...
+      # If you aren't checking for "not null"
+      if refValue == null && expected != null
+        false
+      else
+        refValue != expected
+
+  # Gets the actual answer for the referred question.
+  actual: ->
+    # If readonly, use the data-val attrib
+    if @manager.readOnly
```
smoyth: We don't need readOnly any more, or any of the .ro-val stuff. Can you strip all that out? Conditions are now only useable when you're creating/editing a response.

spec/features/hierarchical_response_display_logic_spec.rb
```diff
@@ -0,0 +1,386 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "hierarhcical response form display logic", js: true do
+  def visit_new_hierarchical_response_page
```
smoyth: Can helper methods go at the bottom please?

spec/features/hierarchical_response_display_logic_spec.rb
```diff
@@ -0,0 +1,386 @@
...
+          display_conditions_attributes: [
+            {ref_qing_id: qings[:rpt1].id, op: "eq", value: "qux"} # References same group Q
+          ])
+
+        qings[:rpt3] = create_questioning("text", form,
+          parent: rpt_group,
+          display_if: "all_met",
+          display_conditions_attributes: [
+            {ref_qing_id: qings[:grp1].id, op: "eq", value: "nix"} # References Q from sibling group
+          ])
```
smoyth: Perhaps we could add a nested group in here without too much trouble? And add in the bit that tests that we search first in our own neighborhood before globally?

app/views/hierarchical_responses/answers/_nodes.html.erb
scttnlsn: `setup_condition_computer` is only called in the `new` controller action so the condition below should limit it to `new`-only as is

spec/features/hierarchical_response_display_logic_spec.rb
scttnlsn: Will move into context

smoyth:
I just made one commit to simplify the refQingElement method. Not sure if you're done with this yet. Are you still going to attempt the nested group spec?

scttnlsn:
@smoyth Great.  Yeah, I'll work on that spec now.

PR #481, opened 2018-08-15
8662 web response saving

DevneyHamilton:


app/models/response_node.rb
```diff
@@ -4,15 +4,35 @@
...
-    children.each { |c| c.response_id = response_id }
+    children.each do |c|
+      if !c.relevant? || c._destroy
+        c.destroy
+      else
+        c.response_id = response_id
+      end
+    end
+    # update inst_nums. inst_num and this block will go away with answer_arranger
+    children.reject(&:destroyed?).sort_by(&:new_rank).each_with_index do |c, i|
```
smoyth: Then let's put this in a third method and mark it with a TODO so we can easily search and destroy it.

spec/models/response_spec.rb
```diff
@@ -103,4 +103,43 @@
...
+            "B",
+            {repeating: [["C", %w[Dog], create(:media_image)]]}
+          ],
+          [
+            "D",
+            {repeating: [["E", %w[Cat Dog], create(:media_image)], ["F", %w[Cat], create(:media_image)]]}
+          ]
+        ]}
+      ]
+    end
```
smoyth: Nice example!

spec/models/response_spec.rb
```diff
@@ -103,4 +103,43 @@
...
+    let(:user) { create(:user) }
+    let(:form) { create(:form, question_types: question_types) }
+    let(:response) { create(:response, user: user, form: form, answer_values: answer_values) }
+
+    it "destroys nested response tree nodes, media, choices" do
+      expect(Choice.all.count).to eq 0
+      expect(Media::Image.all.count).to eq 0
+      response # create response
```
smoyth: Nice comment

spec/models/response_spec.rb
```diff
@@ -103,4 +103,43 @@
...
+      expect(Choice.all.count).to be > 0
+      expect(Media::Image.all.count).to be > 0
+      expect(Answer.where(response_id: response.id, type: "Answer").count).to eq 12
```
smoyth: `type: Answer` is redundant here b/c you're doing where on `Answer`

smoyth: But for that matter why not instead do a check of how many ResponseNodes there are before and after? Then you cover AnswerGroups, etc...

smoyth: Oh, also since the DB should be empty before each example, you can just do `ResponseNode.count` should be zero. No need to filter at all I think.

smoyth:
PS. It's a testament to the quality of your work that most of this "just worked" :-)

app/models/response_node.rb
```diff
@@ -4,15 +4,35 @@
...
+  attr_accessor :relevant
+
   belongs_to :form_item, inverse_of: :answers, foreign_key: "questioning_id"
   belongs_to :response
   has_closure_tree order: "new_rank", numeric_order: true, dont_order_roots: true, dependent: :destroy
 
   before_save do
-    children.each { |c| c.response_id = response_id }
+    children.each do |c|
+      if !c.relevant? || c._destroy
```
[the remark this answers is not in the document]
DevneyHamilton: I don't think the destroy completely takes effect until save, so I do still have to filter out the destroyed nodes in `propogate_response_id` and `update_inst_nums` but other than that this is all great!

app/models/response_node.rb
```diff
@@ -47,4 +59,33 @@ def _destroy
...
+          c.parent.inst_num
+        else
+          1
+        end
+      c.inst_num = new_inst_num
+    end
+  end
+
+  def irrelevant_or_marked_destroy?
+    !relevant? || _destroy
```
DevneyHamilton: For some reason `destroy?` didn't work here. I need to move on to other project work so I'd prefer not to figure it out today, if that's ok?

PR #483, opened 2018-08-16
8633 enumerator login instruction bug

arisah-ijeoma:


spec/features/user_form_password_spec.rb
```diff
@@ -218,6 +218,13 @@ def fill_out_form(role: "Enumerator", email: true, admin: false)
           include_examples("generating new password")
           include_examples("entering new password")
         end
+
+        context "enumerator can see login instructions after he resets his password" do
+          let(:actor) { user }
+
+          include_examples("generating new password")
+          include_examples("entering new password with login instructions")
+        end
```
smoyth: Well done Jay! This is a great use of existing shared examples to test the thing you need to test. Did you confirm this failed before you made the change to make it pass?

Also, I'd like you to take this a step further. Note that the first example in this file is about the user editing their profile. So there is a bit of duplication here now. Can you find it? Can you clean it up? Happy to pair if you need.

app/models/ability.rb
```diff
@@ -85,7 +84,7 @@ def initialize(params)
       end
 
       # anybody can access missions to which assigned (but don't need this permission if admin)
-      if !user.admin?
+      unless user.admin?
```
smoyth: Thanks for cleanup!

app/models/ability.rb
```diff
@@ -56,13 +56,12 @@ def initialize(params)
 
           # standard objects, missions, settings, and all users are available in no-mission (admin) mode
           [Form, Questioning, FormItem, SkipRule, QingGroup, Condition, Question, OptionSet,
-            OptionNode, Option, OptionSetImport, Tag, Tagging].each do |k|
+            OptionNode, Option, OptionSetImport, Setting, Tag, Tagging].each do |k|
```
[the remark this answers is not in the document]
smoyth: Nice cleanup here too!

app/models/ability.rb
```diff
@@ -110,6 +109,9 @@ def initialize(params)
             u == user
           end
 
+          # Enumerators can edit themselves in mission mode even if they're not currently assigned.
+          can :login_instructions, User, id: user.id
```
smoyth: I think this code makes sense but the comment doesn't...

spec/features/user_form_password_spec.rb
smoyth: I am also now wondering if this would work for a staffer. We should probably have specs for all of them...

PR #492, opened 2018-09-10
6423: spec that creates nested groups on edit, not saving right (pass…

DevneyHamilton:
…es before save).


spec/features/qing_group_form_spec.rb
```diff
@@ -53,4 +53,50 @@
...
+
+    click_button("Save") # save form
+
+    # find new parent group's li
+    # append an <ol class="item-list ui-sortable"> to parent
+    # get child li
+    # append child li to new ol
+    # trigger drop_happened on child li with null event
+    execute_script("
+      var new_parent = $('li:contains(\"#{outer_name}\")')[0];
```
DevneyHamilton: I tried making a function for this so I could just pass in the parent and child groups names, but I couldn't get it to work - the inner group kept disappearing.

DevneyHamilton: Tested a couple times in console, couldn't figure out why it was working there and not in the spec.

spec/features/qing_group_form_spec.rb
```diff
@@ -53,4 +53,49 @@
...
+        var newParent = itemLi(newParentName);
+        var newOl = $('<ol class=\"item-list ui-sortable\"></ol>');
+        newParent.append(newOl);
+        var item = itemLi(itemName);
+        item.detach().appendTo(newOl);
+        ELMO.formItemsView.draggable.drop_happened(null, {item: item});
+      }
+      moveItem('#{middle_name}', '#{outer_name}');
+      moveItem('#{inner_name}', '#{middle_name}');
+    SCRIPT
```
smoyth: I think this works. I think the key was adjusting the way the `li` tags are grabbed. Note that `:contains` will match any `li` with the text anywhere inside it, even if it's nested a few layers deep. If we use the `.inner` `div`, that fixes it.

smoyth: Note also the change to camel case. Obviously the app is inconsistent but camelCase for JS is the prevailing standard so I want to move us in that direction.

spec/features/qing_group_form_spec.rb
```diff
@@ -53,4 +53,49 @@
...
+    within(".form-items") { expect(page).to have_css(middle_css, text: middle_name) }
+    within(".form-items") { expect(page).to have_css(inner_css, text: inner_name) }
+  end
+
+  def create_group(name)
+    click_link("Add Group")
+    fill_in("Name (English)", with: name)
+    check("qing_group_repeatable")
+    within(".modal") { click_button("Save") }
+  end
```
smoyth: Great job on this spec! I know it was tedious but I think it's worth it!

spec/features/qing_group_form_spec.rb
```diff
@@ -53,4 +53,49 @@
...
+        newParent.append(newOl);
+        var item = itemLi(itemName);
+        item.detach().appendTo(newOl);
+        ELMO.formItemsView.draggable.drop_happened(null, {item: item});
+      }
+      moveItem('#{middle_name}', '#{outer_name}');
+      moveItem('#{inner_name}', '#{middle_name}');
+    SCRIPT
+    execute_script(move_items_js)
+
```
smoyth: I think we should actually have a 'Save' in here and then return to the page to ensure persistence happened.

Refreshing the page would work too I think since it saves via AJAX but might as well do 'Save' since that's what the user would do? 

Sorry if I wasn't clear about this earlier.

app/assets/javascripts/views/form_items_draggable_list_view.js.coffee
```diff
@@ -72,7 +72,7 @@ class ELMO.Views.FormItemsDraggableListView extends ELMO.Views.ApplicationView
 
   check_depth: (placeholder, item) ->
     depth = this.get_depth(placeholder)
-    depth == 1 || item.hasClass('form-item-question') && depth <= 2
+    depth == 1 || item.hasClass('form-item-question') && depth <= 2 || item.hasClass('form-item-group')
```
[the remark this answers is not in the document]
smoyth: Actually, there is an issue here. I don't think adding a question to a nested group works. Can you please add that to your feature spec?

This code can really change meaning now. The only constraint is that the parent must be root or a group. Can't be a question.

Also I noticed the error messages in the placeholders are not translated. Are you getting that? Might be an older bug but can you fix please?

DevneyHamilton:
The new spec is failing on CI often. If I adjust the timeout value up or down, it fails more often on local. I am pretty sure it depends on the timing of the JS in CI.

PR #494, opened 2018-09-13
8672: check that form item mission matches response mission in web re…

DevneyHamilton:
…sponse parser

### DIFF
--- a/app/models/results/web_response_parser.rb
+++ b/app/models/results/web_response_parser.rb
@@ -111,7 +111,8 @@ module Results
     end
 
     def form_items_in_mission
-      @form_items_in_mission || FormItem.where(mission_id: @response.mission_id).pluck(:id).index_by(&:itself)
+      @form_items_in_mission ||=
+        FormItem.where(mission_id: @response.mission_id).pluck(:id).index_by(&:itself)
     end
   end
 end
--- a/app/models/results/web_response_parser.rb
+++ b/app/models/results/web_response_parser.rb
@@ -61,7 +61,7 @@ module Results
 
     def update_or_add_node(web_hash_node, tree_parent)
       id = web_hash_node[:id]
-      raise SubmissionError, "Form item id invalid." if invalid?(web_hash_node[:questioning_id])
+      raise SubmissionError, "Form item id invalid." unless item_in_mission?(web_hash_node[:questioning_id])
       # add
       if id.blank?
         tree_parent.children.build(new_tree_node_attrs(web_hash_node, tree_parent))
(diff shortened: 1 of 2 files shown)


spec/models/results/web_response_parser_spec.rb
```diff
@@ -61,6 +61,22 @@
           expect_children(tree, %w[Answer Answer], [form.c[0].id, form.c[2].id], %w[A C])
         end
       end
+
+      context "with form item that is not in this mission" do
+        let(:other_mission) { create(:mission) }
+        let(:other_form) { create(:form, mission: other_mission, question_types: %w[text]) }
+        let(:answers) do
+          {
+            "0" => web_answer_hash(other_form.c[0].id, value: "A"),
```
smoyth: So nice to be able to reuse helpers like this!

app/models/results/web_response_parser.rb
```diff
@@ -104,5 +105,9 @@ def inst_num(type, tree_parent)
         1
       end
     end
+
+    def invalid?(questioning_id)
+      FormItem.find(questioning_id).mission_id != @response.mission_id
```
smoyth: Nice! One issue with this though is that it will cause a DB query for every single ResponseNode, which is unnecessary.

As an alternative, you could have a method that returns a hash where the keys are IDs of all FormItems in the mission and the values are the same thing. (Try `FormItem.where(mission_id: @response.mission_id).pluck(:id).index_by(&:itself)`). This should be a memoized method.

Then you can simply do an _O(1)_ hash lookup to see if an incoming qing_id is in the mission.

```

File: app/models/results/web_response_parser.rb [hunk 3]
```diff
@@ -104,5 +105,9 @@ def inst_num(type, tree_parent)
         1
       end
     end
+
+    def invalid?(questioning_id)
```
smoyth: Maybe make the method name more specific since this could be confused with validation.

spec/models/results/web_response_parser_spec.rb
```diff
@@ -61,6 +61,22 @@
...
+            "1" => web_answer_hash(form.c[1].id, {value: "B"}, destroy: "true"),
+            "2" => web_answer_hash(form.c[2].id, value: "C")
+          }
+        end
+
+        it "errors" do
+          expect { described_class.new(response).parse(input) }.to raise_error(SubmissionError)
+        end
```
smoyth: Lovely spec!

app/models/results/web_response_parser.rb
```diff
@@ -104,5 +105,13 @@ def inst_num(type, tree_parent)
         1
       end
     end
+
+    def item_in_mission?(questioning_id)
+      form_items_in_mission[questioning_id].present?
+    end
+
+    def form_items_in_mission
+      @form_items_in_mission || FormItem.where(mission_id: @response.mission_id).pluck(:id).index_by(&:itself)
```
smoyth: You're missing a `=` after the `||`. Other than that you nailed it.

PR #495, opened 2018-09-14
8793 response read only

scttnlsn:


spec/features/results/read_only_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "response form tree handling", js: true do
```
smoyth: `"response form read only mode"`?

spec/features/results/read_only_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+        }
+      ]
+    )
+  end
+
+  let(:params) { {locale: "en", mode: "m", mission_name: get_mission.compact_name} }
+
+  before { login(user) }
+
+  describe "response rendering" do
```
smoyth: I think `describe` is not needed?

spec/features/results/reviewer_notes_spec.rb
```diff
@@ -8,25 +8,60 @@
...
+    scenario "should be visible" do
+      visit(hierarchical_response_path(response, params))
+      expect(page).to have_content(notes)
+    end
+
+    scenario "can submit notes" do
+      visit(hierarchical_response_path(response, params))
+      fill_in("Notes", with: "testing")
+      click_button("Save")
+      expect(page)
```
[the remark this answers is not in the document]
scttnlsn: Whoops.  I'll remove this.

smoyth:
This looks real purdy!

Couple things:

[ ] Shouldn't be a save button
[ ] None of the metadata should be editable
[ ] For the 'reviewed' line, if is reviewed, just include the same label text ("This response has been reviewed") but no checkbox.
[ ] If not reviewed, show "This response has _not_ been reviewed", with "not" in italic.
[ ] If not reviewed, don't show reviewer or reviewer notes
[ ] Add a normalization to Response model to set reviewed = true if reviewer_id or reviewer_notes is true. See FormItem for the standard way of doing normalization. Include a simple model spec, see form_item_spec.rb:118 for a good example of this.
[ ] Add a DB migration to set reviewed = true for all records where reviewer_id or reviewer_notes is true

Then we should be golden. Thanks!

app/views/hierarchical_responses/_metadata.html.erb
```diff
@@ -2,21 +2,32 @@
...
+      <% end %>
     </div>
     <% if @response.persisted? && can?(:review, @response) %>
       <div class="col-sm-6">
         <%= f.field(:reviewed, type: :check_box) %>
         <%= f.field(:reviewer_id, type: :select, options: sel_opts_from_objs(@response.reviewer)) %>
         <%= f.field(:reviewer_notes, type: :textarea) %>
+
+        <% if context.read_only? %>
+          <%= f.submit(class: "btn btn-primary pull-right") %>
```
smoyth: Can you change this to be wrapped in a `<div class="submit-buttons">`? `pull-right` is not RTL-friendly. This may add too much margin to the top, in which case you could add a second `submit-buttons-tight` class or something like that that would cancel or reduce the margin in the `app/assets/stylesheets/local/html_forms/_all_forms.scss` file.

PR #496, opened 2018-09-14
6440 nested groups on printable form

arisah-ijeoma:


arisah-ijeoma:
screen shot 2018-09-18 at 3 06 13 pm

app/views/form_items/_printable.html.erb
```diff
@@ -1,10 +1,10 @@
-<% items.each do |item, subitems| %>
-  <% if item.is_a?(QingGroup) %>
-    <% if !subitems.empty? %>
-      <%= render "qing_groups/printable", qing_group: item %>
```
smoyth: Why did this move out of the `if` block? I think if there are no children, we don't want to render the group at all.

app/views/form_items/_printable.html.erb
```diff
@@ -1,10 +1,10 @@
...
-      <%= render "form_items/printable", items: subitems %>
+<% items.each do |item| %>
+  <% if item.is_a?(Questioning) %>
+    <%= render "questionings/printable", qing: item.decorate %>
+  <% elsif item.is_a?(QingGroup) %>
+    <%= render "qing_groups/printable", qing_group: item %>
+    <% if item.children? %>
+      <%= render "form_items/printable", items: item.visible_children %>
     <% end %>
-  <% elsif item.printable? %>
```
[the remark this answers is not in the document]
arisah-ijeoma: I did not see a need for it as I am doing a check for the item is a Questioning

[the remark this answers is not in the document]
smoyth: `printable?` is checking the question type to make sure it is printable. No sense in printing an image capture question, for instance.

arisah-ijeoma:
WIP - over 15 failing specs

DevneyHamilton:
Hey Ijeoma - I was able to get the new spec passing with number and qtype_names (but haven't checked other specs) by:

1) removing line 56 `  #form.questionings << questioning` from the factories/forms.rb (this is what was causing the 'questionings' method to run a bunch of times, and it's not necessary). 
and
2) removing line `self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }` from the `preordered_descendants` method. It should not be necessary because we already order by rank in the first line. 

Try these two things and let me know what happens.

app/models/form.rb
```diff
@@ -35,9 +35,10 @@ def self.receivable_association
   scope(:by_name, -> { order("forms.name") })
   scope(:default_order, -> { by_name })
 
-  delegate :arrange_descendants,
-    :children,
+  delegate :children,
+    :preordered_descendants,
```
smoyth: If this is not being used elsewhere in the app then I don't think we should add it to the class' public interface like this. You always want to keep the public interface as small as possible.

app/models/form_item.rb
```diff
@@ -77,46 +77,14 @@ def base_item
...
-  end
-
-  def preordered_descendants(eager_load: nil)
-    items = descendants.order(:rank)
-    items = items.includes(eager_load) if eager_load
-    self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }
+  def preordered_descendants(eager_load: nil, type: nil)
+    items = eager_load ? descendants.includes(eager_load) : descendants
+    sorted_items = self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }
+    type ? sorted_items.select { |i| i.type == type } : sorted_items
```
smoyth: Looks terrific! I think we should add a comment here as to why the items are filtered by type _after_ the sort. The passing observer might be tempted to do the type filtering before the sort, but we know that that breaks things because the QingGroups are needed in the array for the sort to work properly!

spec/factories/forms.rb
```diff
@@ -52,9 +52,7 @@ def create_questioning(qtype_name_or_question, form, attribs = {})
   attribs[:mission] = form.mission
   attribs[:form] = form
   attribs[:question] = question
-  questioning = create(:questioning, attribs)
-  form.questionings << questioning
-  questioning
+  create(:questioning, attribs)
```
smoyth: Glad this could be simplified without breaking anything!

spec/support/helpers/request_spec_helpers.rb
```diff
@@ -50,4 +50,35 @@ def post_s(path, opts = {})
...
+    sort = "(case when ancestry is null then 0 else 1 end), ancestry, rank"
+    # We eager load questions and option sets since they are likely to be needed.
+    nodes = form.descendants.includes(question: {option_set: :root_node}).order(sort).to_a
+    with_self = Questioning.arrange_nodes(nodes)
+    with_self.keys
+  end
+
+  # Arrange array of nodes into a nested hash of the form
+  # {node => children}, where children = {} if the node has no children
+  def arrange_nodes(nodes)
```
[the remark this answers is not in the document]
smoyth: Nice work!

spec/support/helpers/request_spec_helpers.rb
```diff
@@ -50,4 +50,35 @@ def post_s(path, opts = {})
...
+      depth = node.depth
+      if depth < min_depth
+        min_depth = depth
+        arranged.clear
+      end
+      arranged[node] = children if depth == min_depth
+    end
+
+    arranged
+  end
```
smoyth: Can these helper methods please be moved to the `odk_submissions` context since that's the only place they're needed? And perhaps marked with comments to the effect that these methods should be removed when the spec is refactored to use the newer method of constructing XML fixtures.

spec/support/contexts/odk_submissions.rb
```diff
@@ -82,4 +82,40 @@ def submission_path(mission = nil)
   def allow_forgery_protection(allow)
     ActionController::Base.allow_forgery_protection = allow
   end
+
+  ###############################################################################
+  # These last two methods should be removed when the spec is refactored to use
+  # the newer method of constructing XML fixtures. See ODK parser specs.
+  ###############################################################################
```
smoyth: Nice!

app/models/form_item.rb
```diff
@@ -77,46 +77,17 @@ def base_item
...
+  def preordered_descendants(eager_load: nil, type: nil)
+    items = eager_load ? descendants.includes(eager_load) : descendants
+    sorted_items = self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }
 
-  def preordered_descendants(eager_load: nil)
-    items = descendants.order(:rank)
-    items = items.includes(eager_load) if eager_load
-    self.class.sort_by_ancestry(items) { |a, b| a.rank <=> b.rank }
+    # the items are filtered after the sort because we need to have all
+    # FormItem types (Questioning and QingGroup) to be available for the sort to work properly
```
smoyth: Perfect.

PR #501, opened 2018-09-25
8911 set default locale

scttnlsn:


smoyth:
After looking at this more closely I think `configatron.default_locale` was a mistake and I think it can be removed. I have done so. Take a look and see what you think.

smoyth:
There is feature spec coverage on this and they are still passing so I think we're good.

PR #504, opened 2018-09-27
8902 dynamic pattern validator error

arisah-ijeoma:


spec/features/questionings_form_spec.rb
```diff
@@ -87,6 +87,15 @@
...
       end
+
+      context "errors" do
+        it "should render error properly when there is no type" do
+          visit("/en/m/#{get_mission.compact_name}/forms/#{form.id}/choose-questions")
+          fill_in "Code", with: "foodie"
+          click_on("Save")
+          expect(page).to have_content("Question is invalid")
+        end
+      end
```
smoyth: Nice job on this spec. It is correct and covers the issue appropriately.

However I think it should rather be a model spec. Remember the testing pyramid. We try to save feature specs for integrations that model specs can't cover. But you can cover this case in the DynamicPatternValidator model spec, no?

app/models/question.rb
```diff
@@ -69,7 +69,7 @@ class Question < ApplicationRecord
            :odk_tag,
            :odk_name,
            :select_multiple?,
-           to: :qtype
+           to: :qtype, allow_nil: true
```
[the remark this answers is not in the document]
smoyth: This approach works but it has the unfortunate side effect of changing the method contract of 9 different methods on this class—specifically it means that they all may return nil, where that wasn't possible before. This is a big problem because it may violate the assumptions of all kinds of code throughout the app. Code that didn't need to check for a nil value when calling e.g. `printable?` now does, or should, or otherwise it is vulnerable to hard-to-debug errors. This is why nils need to be used quite carefully and are actually considered a code smell in some languages!

I think a better way to fix this is to change this line in `Questioning`:

```
  validates_with Forms::DynamicPatternValidator, field_name: :default, force_calc_if: :numeric?
```

Instead of having it provide the symbol `:numeric?` for `force_calc_if`, have it provide a Proc and then in that proc you can do e.g.  `->(qing) { qing.qtype&.numeric? }`. Then you will have to update the DynamicPatternValidator to call the proc instead of using `send` to call the named method. Since Questioning is the only class using `force_calc_if`, it's fine to swap out this behavior.

Or if you have another suggestion of a way to avoid changing the contract for all these methods, feel free to propose it!

Do you see the difference and why it's important?

smoyth: This approach works but it has the unfortunate side effect of changing the method contract of 9 different methods on this class—specifically it means that they all may return nil, where that wasn't possible before. This is a big problem because it may violate the assumptions of all kinds of code throughout the app. Code that didn't need to check for a nil value when calling e.g. `printable?` now does, or should, or otherwise it is vulnerable to hard-to-debug errors. This is why nils need to be used quite carefully and are actually considered a code smell in some languages!

I think a better way to fix this is to change this line in `Questioning`:

```
  validates_with Forms::DynamicPatternValidator, field_name: :default, force_calc_if: :numeric?
```

Instead of having it provide the symbol `:numeric?` for `force_calc_if`, have it provide a Proc and then in that proc you can do e.g.  `->(qing) { qing.qtype&.numeric? }`. Then you will have to update the DynamicPatternValidator to call the proc instead of using `send` to call the named method. Since Questioning is the only class using `force_calc_if`, it's fine to swap out this behavior.

Or if you have another suggestion of a way to avoid changing the contract for all these methods, feel free to propose it!

Do you see the difference and why it's important?

arisah-ijeoma:
I didn't get rid of the force_calc_if because it ties to another method the validator uses to validate
This method was written two months ago @smoyth

app/validators/forms/dynamic_pattern_validator.rb
```diff
@@ -11,7 +11,7 @@ def initialize(options)
 
     def validate(record)
       calc_must_wrap_all_of_default_response_name(record)
-      force_calc_for_dollar_refs(record) if force_calc_if.present? && record.send(force_calc_if)
+      force_calc_for_dollar_refs(record) if force_calc_if&.call(record)
```
smoyth: Nice.

PR #505, opened 2018-09-28
8916 add reviewed column to response export

arisah-ijeoma:


spec/models/results/csv/generator_spec.rb
```diff
@@ -231,12 +241,12 @@
...
       # Simulate some conditions like we'd get from a search.
       Response.for_mission(get_mission).joins(:form)
-        .where("(((responses.reviewed = 'n')) AND ((forms.name ILIKE '%Sample%')))")
+        .where("(((responses.reviewed = 'n')) OR ((forms.name ILIKE '%Sample%')))")
     end
 
     before do
       Timecop.freeze(Time.zone.parse("2015-11-20 12:30 UTC")) do
-        create_response(form: form1, mission: missions[0], answer_values: ["foo"])
+        create_response(form: form1, mission: missions[0], answer_values: ["foo"], reviewed: true)
```
smoyth: I appreciate that you're trying to exercise the reviewed code throughout this spec. It's a very thorough energy you're bringing! But you really only need to do it in perhaps one place so that you've proven that the values coming out in the CSV really correspond to the ones you've set (e.g. they're not just 'true' everytime).

This is generally something to keep in mind in your work— ask yourself: "what is the least number (and type) of tests I need to write here to properly cover the code I'm adding". Tests are great but they are also code that needs to be written AND maintained, so keeping them to a minimum (without sacrificing coverage) is important.

So with that in mind I don't think you need to disturb as much of this file or e.g. change the SQL fragment above. I think you could go with the factory default of `false` except in one of the above where you have a mixture. That would increase the simplicity of your changes here, which is generally a good thing!

spec/requests/responses_csv_spec.rb
```diff
@@ -15,8 +15,8 @@
     expect(response.headers["Content-Disposition"]).to match(
       /attachment; filename="elmo-#{get_mission.compact_name}-responses-\d{4}-\d\d-\d\d-\d{4}.csv"/
     )
-    expect(result.size).to eq 3 # 2 response rows, 1 header row
-    expect(result[1][8]).to eq "Animal"
-    expect(result[2][8]).to eq "Plant"
+    expect(result.size).to eq(3) # 2 response rows, 1 header row
+    expect(result[1][9]).to eq("Animal")
+    expect(result[2][9]).to eq("Plant")
```
smoyth: Thx for cleanup!

PR #508, opened 2018-10-01
7479 option set clone

scttnlsn:


scttnlsn:
```ruby
expect(value).to eq expected
```
is pretty standard RSpec syntax.  Should I actually add the parentheses that Hound wants?

smoyth:
Next step here is that Tom needs to test the migration and make sure adequate backups are built into the upgrade procedure if necessary.

PR #510, opened 2018-10-04
8102 operations migration

DevneyHamilton:
For merging with attachment-as-operation work, need to rename operations attribute "description" to "details"

### DIFF
--- a/app/models/operation.rb
+++ b/app/models/operation.rb
@@ -1,10 +1,6 @@
 class Operation < ApplicationRecord
   belongs_to :creator, class_name: 'User'
 
-  validates :job_class, presence: true
-  validates :creator, presence: true
-  validates :details, presence: true
-
   def name
     "##{id}"
   end
--- a/app/models/operation.rb
+++ b/app/models/operation.rb
@@ -3,7 +3,7 @@ class Operation < ApplicationRecord
 
   validates :job_class, presence: true
   validates :creator, presence: true
-  validates :description, presence: true
+  validates :details, presence: true
 
   def name
     "##{id}"
--- a/db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
+++ b/db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
@@ -2,7 +2,6 @@
 
 class RefactorOperationsToHaveMissionAndNewAttributes < ActiveRecord::Migration[5.1]
   def up
-    # soft delete? Operation.all.each{|o| o.delete}
     execute("DELETE FROM operations")
     remove_column(:operations, :creator_old_id)
     remove_column(:operations, :old_id)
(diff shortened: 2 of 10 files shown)


app/models/operation.rb
```diff
@@ -3,7 +3,7 @@ class Operation < ApplicationRecord
 
   validates :job_class, presence: true
   validates :creator, presence: true
-  validates :description, presence: true
+  validates :details, presence: true
```
smoyth: Let's please remove _all_ these validations (not just the new one you added).

Validations only make sense if there is a form involved, and there is not in this case.

db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+class RefactorOperationsToHaveMissionAndNewAttributes < ActiveRecord::Migration[5.1]
+  def up
+    execute("DELETE FROM operations")
+    remove_column(:operations, :creator_old_id)
+    remove_column(:operations, :old_id)
+    rename_column(:operations, :description, :details)
+    add_column(:operations, :mission_id, :uuid)
```
smoyth: Whenever you add a foreign key you should add a foreign key constraint and an index. This checklist might be helpful: https://redmine.sassafras.coop/projects/sassafras-internal/wiki/Rails_Best_Practices

Search the project for add_foreign_key and add_index for the appropriate syntax.

Don't need to worry about them in the `down` migration, they will disappear when you remove the column.

db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
```diff
@@ -0,0 +1,21 @@
...
+    add_column(:operations, :unread, :boolean, default: true, null: false)
+  end
+
+  # up and down instead of change because can't un-delete
+  def down
+    add_column(:operations, :creator_old_id, :integer)
+    add_column(:operations, :creator_old_id, :integer)
```
smoyth: Duplicate line? Should be `old_id`?

db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
```diff
@@ -0,0 +1,21 @@
...
+    rename_column(:operations, :details, :description)
+    remove_column(:operations, :mission_id)
+    remove_column(:operations, :unread)
```
smoyth: Looks great otherwise!

db/migrate/20181004181337_refactor_operations_to_have_mission_and_new_attributes.rb
DevneyHamilton: @smoyth A lot of the foreign keys in the db have `on_update: restrict`, and `on_delete: restrict`. Is this what we want for this foreign key as well? Is this assuming that the `has_many :operations, inverse_of: :mission, dependent: :destroy` will delete an operation when its mission is deleted?

DevneyHamilton: answered :)

PR #512, opened 2018-10-08
8103 mission based operations

DevneyHamilton:

PR #516, opened 2018-10-09
8930 add operations panel link permanently on navbar

arisah-ijeoma:
We need to test this with #512 before merging into develop


DevneyHamilton:
Note for merging: we will need to manually test that the link correctly is admin or mission-based after this 8103 is merged in.

PR #522, opened 2018-10-16
8979 generalize media uploader

DevneyHamilton:


app/assets/javascripts/views/file_uploader_view.js.coffee
```diff
@@ -26,6 +27,9 @@ class ELMO.Views.MediaUploaderView extends ELMO.Views.ApplicationView
   events:
     'click .existing a.delete': 'delete_existing'
 
+  set_listener: (listener) ->
```
DevneyHamilton: Would prefer to just pass in the listener. Made this method because in media_editable.html, the params are passed in in ruby, and referencing the JS object in ruby isn't working. Not sure if this is just a syntax thing, or a js/ruby issue.

config/locales/en/main.yml
```diff
@@ -144,7 +144,7 @@ en:
   errors:
     # our messages are all complete sentences so we use the colon
     format: "%{attribute}: %{message}"
-
+    file_upload: "There was an error uploading the file."
```
DevneyHamilton: I moved the error out of active record bc it will be used in generic file uploading. Is there a better place to put it?

app/assets/javascripts/views/file_uploader_view.js.coffee
```diff
@@ -1,18 +1,20 @@
-class ELMO.Views.MediaUploaderView extends ELMO.Views.ApplicationView
+class ELMO.Views.FileUploaderView extends ELMO.Views.ApplicationView
```
smoyth: Can you refactor this class to use `camelCase` variables throughout? I want us to move us in that direction—right now there is a mixture. The serializer lib we use can do this so that incoming data is in camelCase also (but that's not used here, just for future reference). You'll have to refactor the instantiation call in the template slightly as well.

smoyth: (Remember to use Cmd+D to select multiple occurrences. It will go quickly that way.)

app/views/responses/answers/_media_editable.html.erb
```diff
@@ -16,11 +16,15 @@
...
     zone_id: "#dropzone-#{context.path_str}",
     generic_thumb_path: image_path("media/#{answer.qtype.media_type}.png"),
     post_path: media_create_path(type: media_type),
-    delete_path: answer.media_object ? media_delete_path(answer.media_object, type: media_type) : nil
-  )%>);
+    delete_path: answer.media_object ? media_delete_path(answer.media_object, type: media_type) : nil,
+    preview_template: render("layouts/dropzone_preview"),
+    param_name: "upload"
+  )%>
+  options.listener = ELMO.mediaUploaderManager
```
smoyth: Very nice and clean. Semicolon tho plz.

app/views/responses/form.html.erb
```diff
@@ -1,11 +1,10 @@
 <% content_for(:per_page_js, javascript_google_maps) %>
 <% @title_args = {id: @response.decorate.shortcode} %>
 
-<%# This needs to be up here so it runs before the MediaUploaderViews are initialized. %>
+<%# This needs to be up here so it runs before the FileUploaderViews are initialized. %>
 <%= javascript_doc_ready do %>
   ELMO.mediaUploaderManager = new ELMO.Views.MediaUploaderManager(<%=json(
-    el: ".response_form",
-    preview_template: render("responses/answers/dropzone_preview")
+    el: ".response_form"
```
smoyth: We can get rid of the ERB and `json` call now right?

smoyth: Can also change this please to `response-form` (dash). Both are present on the form (it's complicated, but underscores in CSS classes are deprecated).

app/assets/javascripts/views/file_uploader_view.js.coffee
smoyth: Please also add a class comment.

config/locales/en/main.yml
DevneyHamilton: @smoyth how would I figure out if it's a validation error? I see "validates_attachment_presence" in /models/media/object.rb but I dont' know how to see what the text of that error would be?

smoyth: If it were being used that way, the symbol `file_upload` would need to be passed to `error.add`. So if that's not showing up anywhere I think you're golden.

DevneyHamilton: the old one was "generic" but that's not getting added to errors anywhere either so I think it's good to go.

app/assets/javascripts/views/media_uploader_manager.js.coffee
```diff
@@ -1,28 +1,27 @@
...
+    @updateButtons()
 
-  update_buttons: ->
-    can_submit = @uploads_in_progress == 0
-    @$(".submit-buttons .btn-primary").css('display', if can_submit then 'inline-block' else 'none')
-    @$("#upload-progress-notice").css('display', if can_submit then 'none' else 'inline-block')
+  updateButtons: ->
+    canSubmit = @uploadsInProgress == 0
+    @$(".submit-buttons .btn-primary").css('display', if canSubmit then 'inline-block' else 'none')
+    @$("#upload-progress-notice").css('display', if canSubmit then 'none' else 'inline-block')
```
smoyth: Nice work changing this one over too!

PR #523, opened 2018-10-16
8814 admin react form

jpdimond:


spec/features/user_form_spec.rb
```diff
@@ -2,34 +2,75 @@
...
+      click_link("Ella Baker")
+      expect(page).to have_content("Staffer")
+
+      # Edit user and change assignment, verify in show
+      click_link("Edit User")
+      select("reviewer", from: "user[assignments_attributes][0][role]")
+      click_on("Save")
+      expect(page).to have_content("Success: User updated successfully")
+      click_on("View User")
+      expect(page).to have_content("Reviewer")
```
smoyth: Also nice and clear, well done. You need to be careful with doing `have_content` on the whole page. You may get a false pass, if for instance 'Reviewer' appears somewhere else on the page, which is not totally unrealistic. You could perhaps try `"Role Reviewer"`, which would narrow it down a lot. Even better would be `have_css('.user_role', content: "Reviewer")`. Note this should change in several places in this file.

spec/features/user_form_spec.rb
```diff
@@ -2,34 +2,75 @@
...
+  end
+
+  context "admin user in admin mode" do
+    scenario "create and edit user should work" do
+      visit("/en/admin/users/new")
+      fill_in("Full Name", with: "Ella Baker")
+      fill_in("Username", with: "ellab")
+      fill_in("Email", with: "ella@baker.com")
+      find("a", text: "Add Assignment").click
+      select("staffer", from: "user[assignments_attributes][0][role]")
```
smoyth: Capital 'S'?

spec/features/user_form_spec.rb
```diff
@@ -2,34 +2,75 @@
...
+      click_on("Save")
+      expect(page).to have_content("Success: User created successfully")
+
+      # Go to show user
+      click_link("Ella Baker")
+      expect(page).to have_content("Staffer")
+
+      # Edit user and change assignment, verify in show
+      click_link("Edit User")
+      select("reviewer", from: "user[assignments_attributes][0][role]")
```
smoyth: Capital 'R'?

app/views/users/form.html.erb
```diff
@@ -57,17 +57,7 @@
...
-      <%=json(
-        form_mode: form_mode,
-        assignments: @assignments,
-        assignment_permissions: @assignment_permissions,
-        assignable_missions: @assignable_missions,
-        assignable_roles: @assignable_roles
-      )%>
-    <% end %>
-  );
+  new ELMO.Views.UserForm();
```
smoyth: This can go once you move the JS.

app/helpers/users_helper.rb
```diff
@@ -74,4 +74,10 @@ def country_options
     end
     country_translations
   end
+
+  def assignment_data
+    { missions: Mission.all.map { |m| MissionSerializer.new(m) },
+      roles: User::ROLES.map { |r| r },
```
smoyth: The block here accomplishes nothing I think?

app/javascript/components/UserAssignmentForm.js
```diff
@@ -0,0 +1,54 @@
...
+class UserAssignmentForm extends React.Component {
+  constructor(props) {
+    super();
+    this.state = props;
+    this.handleAddClick = this.handleAddClick.bind(this);
+    this.handleDeleteClick = this.handleDeleteClick.bind(this);
+  }
+
+  randomUUID() {
+    return "new-mission-"+ Math.floor(Math.random() * Math.floor(9000));
```
smoyth: Space before `+`. Bummer that linter didn't catch :(

smoyth: `Math.floor(9000)` is the same as just `9000`. If you copied this from somewhere please fix there too.

app/javascript/components/UserAssignmentForm.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  handleAddClick() {
+    let assignments = this.state.assignments;
+    assignments.push({role: "", mission: this.randomUUID(), new_assignment: true});
+    this.setState({assignments: assignments});
```
smoyth: See https://stackoverflow.com/questions/23966438/what-is-the-preferred-way-to-mutate-a-react-state for notes on how to properly add to an array in state. You're not supposed to mutate the array directly I think.

app/javascript/components/UserAssignmentForm.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  handleDeleteClick(mission, e) {
+    let assignments = this.state.assignments;
+    let idx = assignments.findIndex((a) => { return a.mission_id == mission; });
+    assignments[idx].new ? assignments.splice(idx, 1) : assignments[idx]["destroy"] = true
```
smoyth: Note that `obj["foo"]` is same as `obj.foo`. Please simplify accordingly.

Also please don't do assignment inside a ternary operator. Please use if/else.

Also same goes here for how to properly mutate an array in state. Can you please lookup how to safely remove an item?

app/javascript/components/UserAssignmentForm.js
```diff
@@ -0,0 +1,54 @@
...
+    this.handleDeleteClick = this.handleDeleteClick.bind(this);
+  }
+
+  randomUUID() {
+    return "new-mission-"+ Math.floor(Math.random() * Math.floor(9000));
+  }
+
+  handleAddClick() {
+    let assignments = this.state.assignments;
+    assignments.push({role: "", mission: this.randomUUID(), new_assignment: true});
```
smoyth: A comment here on what the `randomUUID` for would be nice since this is not the actual ID of any mission or whatever. Kind of confusing. I think you're using it as a key for React only? But not sure.

app/javascript/components/UserAssignmentFormField.js
```diff
@@ -0,0 +1,87 @@
+import React from "react";
+import PropTypes from "prop-types";
+
+class UserAssignmentFormField extends React.Component {
```
smoyth: Please include class comments for all React components. What is this for?

app/serializers/assignment_serializer.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AssignmentSerializer < ActiveModel::Serializer
+  attributes :id, :mission_id, :role, :new_assignment, :name, :destroy
+
+  def new_assignment
+    false
+  end
```
smoyth: This and `destroy` seem a bit hacky. Can you design the component code to work without them, just defaulting to false if they're not there?

app/assets/stylesheets/local/users/_users.scss
```diff
@@ -35,44 +35,27 @@ form.user_form {
...
+      display: inline-block;
+      margin-right: 10px;
+      width: 150px;
     }
 
-    div.assignment, div.assignment_header {
-      display: inline-block;
-      min-height: 20px;
+    .assignment-row {
+      height: 50px;
```
smoyth: Please don't use hard coded heights! Does not work well with people changing font sizes, etc.

app/assets/stylesheets/local/users/_users.scss
```diff
@@ -35,44 +35,27 @@ form.user_form {
...
 
   div.form-field#assignments {
-
-    div.item {
-      padding: 2px 0;
+    .mission,
+    .role {
+      display: inline-block;
+      margin-right: 10px;
+      width: 150px;
```
smoyth: I think it would look nicer to do these as 50% columns. Let's discuss.

jpdimond:
[x]It seems you're not showing the translated role names, just the machine readable ones. That will need to be fixed.
[x]Also note that the select boxes don't look like the others on the page. You need to add .form-control I think.
[x]The 'Add Assignment' and trash links don't have the pointer cursor when you mouseover them. This was added using CSS for the skip logic form I believe.
[x]Also let's add a plus icon as with the skip logic one:

app/assets/javascripts/legacy/views/user_form.js
```diff
@@ -33,43 +16,4 @@
     togglePasswordFields.call(passwordSelect);
   }
```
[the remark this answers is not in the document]
jpdimond: done!

app/helpers/users_helper.rb
```diff
@@ -74,4 +74,10 @@ def country_options
...
+      assignments: @user.assignments.map { |a| AssignmentSerializer.new(a) } }
```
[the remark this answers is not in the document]
jpdimond: done

app/javascript/components/UserAssignmentFormField.js
```diff
@@ -0,0 +1,90 @@
...
+      <div className="assignment-row">
+        <div className="mission">
+          {this.missionField()}
+        </div>
+        <div className="role">
+          <select className="form-control" {...roleSelectProps}>
+            {this.roleOptionTags()}
+          </select>
+        </div>
+        <a className="trash" onClick={(e) => this.props.deleteClick(this.props.index)}>
```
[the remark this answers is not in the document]
jpdimond: not sure why not use arrow functions?

PR #526, opened 2018-10-24
8558 form drag invalid position

scttnlsn:


spec/features/forms_flow_spec.rb
```diff
@@ -80,6 +81,64 @@
...
+    last = form_items.last
+
+    # drag first item to the bottom
+    first.drag_to(last)
+
+    form_items = all("li.form-item")
+    last_id = form_items.last["data-id"]
+
+    expect(last_id).to eq(first_id)
+  end
```
smoyth: This is great! It shows that `drag_to` can work! I was curious about how it's working and why our previous assumption was wrong, so I went digging.

It turns out if you do:

```
    page.driver.browser.action.click_and_hold(first.native).perform
    page.driver.browser.action.move_to(last.native).perform
    page.driver.browser.action.move_by(0, 10).perform
```

That will end up with the element `first` hanging out above `last` with the placeholder showing. If you change the `0` in the last line to a `40` (40 should probably be a constant), then you will get the 'parent must be a group' error in the placeholder! This is really exciting. It means we can test this stuff properly.

Can we do the following tests:

* Simple, normal move (basically as you have it above, but I'd like to clean up the specs with some supporting methods, see below).
* Another crack at the big commented spec (basically create three groups, nest them each one under the other, move the question inside the innermost group, save, check for 'success' flash, and see that everything is in the same place. Let's also add this same save check to the simple/normal move.
* The condition drag prohibition thing but with a test of the error placeholder (maybe split this into two tests, one for dragging the q with the condition, one for dragging a different question past it).
* A spec testing that you can't drag a question to be a child of a question, and that the right placeholder error appears.

I'd like to beef up `have_nested_item` a bit to be able to specify the rank path to the expected item. Could maybe do something like `ol > li:first-child + li + li + li` to get the exactly fourth child of a given level, etc. This will ensure the placeholders and items are ending up where we expect.

Lets move all these draggy things to a separate spec `forms/item_edit_spec.rb` and rename the existing one to `forms/printable_spec.rb` since I think that's all thats in there.

Can we also clean up the specs so that moving an item is as simple as `drag_item(from: [0], to: [3], release: false)` (to drag the first item to position 3 without releasing) and `drag_item(from: [2,1], to: [3,0,0])` (for the nested version), maybe `drag_item(from: [0], to: [3], release: false, indent: true)` to move it to position 3 but over by 40px as I mentioned above. And `release_item` to release the item if `release: false` was passed. (To do this you can do `page.driver.browser.action.release.perform` I think).

This combined with the new/improved have_nested_item should allow a clear but powerful set of specs. What do you think?

scttnlsn: Sounds good.  I've been implementing the `drag_item` method you suggested and re-working the commented-out spec but I'm having trouble getting the indent drags to work consistently.  Ends up working for the first level but not the second.  I'll keep playing around with it (been making heavy use of `screenshot_and_open_image` BTW- quite handy).

spec/features/forms/item_edit_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+  scenario "dragging form elements" do
+    visit(edit_form_path(form, locale: "en", mode: "m", mission_name: get_mission.compact_name))
+
+    question_name = form.questions.first.name
+
+    # drag first item to the bottom
+    drag_item(from: [0], to: [3])
+
+    drag_item(from: [0], to: [1], indent: true, release: false)
+    expect(page).to have_content("The parent must be a group")
```
smoyth: 😍

spec/features/forms/item_edit_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+      expect(page).to have_content("This position would invalidate a condition")
+      release_item
+
+      # try dragging conditional item further down the list
+      drag_item(from: [1], to: [2], release: false)
+      expect(page).not_to have_content("This position would invalidate a condition")
+      release_item
+
+      click_button("Save")
+      expect(page).to have_content("Form updated successfully.")
```
smoyth: 😍 😍

spec/features/forms/item_edit_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    within(".modal") { click_button("Save") }
+  end
+
+  def form_item_selector(path)
+    selector = [".draggable-list-wrapper"]
+
+    path.each do |index|
+      selector << "> ol.item-list"
+      # nth-child is 1-indexed
+      selector << "> li.form-item:nth-child(#{index + 1})"
```
smoyth: Very slick!

spec/features/forms/item_edit_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+    page.driver.browser.action.move_by(indent ? INDENT : 0, 10).perform
+
+    release_item if release
+  end
+
+  def have_nested_item(path:, name:)
+    selector = form_item_selector(path)
+    have_css(selector, text: name)
+  end
+end
```
smoyth: This may be the coolest feature spec ever written :D

PR #527, opened 2018-10-31
9016 rescue from big decimal parse error

scttnlsn:


app/models/answer.rb
```diff
@@ -265,7 +265,7 @@ def replicate_location_values
     if location_type_with_value?
       tokens = self.value.split(" ")
       LOCATION_ATTRIBS.each_with_index do |a, i|
-        self[a] = tokens[i] ? BigDecimal.new(tokens[i]) : nil
+        self[a] = tokens[i] ? (BigDecimal(tokens[i]) rescue nil) : nil
```
smoyth: This form of `rescue` is discouraged I think. Can you move the decimal coercion to a supporting private method, use a named exception, and add a comment explaining why we need to do this?

spec/features/results/responses/location_picker_spec.rb
```diff
@@ -11,6 +11,17 @@
...
+  scenario "manually entering invalid location" do
+    visit new_response_path(params)
+
+    fill_in("Location Question Title 1", with: "invalid")
+    click_button("Save")
+
+    expect(page).to have_content("Response is invalid")
+    expect(page).to have_content("Latitude is out of range")
+    expect(page).to have_content("Longitude is out of range")
+  end
```
smoyth: 👍

PR #529, opened 2018-10-31
8928 v2 user batch progress bar

DevneyHamilton:


lib/tasks/option_set_reclone.rake
```diff
@@ -0,0 +1,6 @@
+require "./lib/task_helpers/option_set_reclone"
```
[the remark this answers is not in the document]
DevneyHamilton: this file has this

app/models/operation.rb
```diff
@@ -8,7 +8,12 @@
 # implement the `perfom(operation, *args)` method.
 class Operation < ApplicationRecord
   include MissionBased
-  belongs_to :creator, class_name: 'User'
+
+  # This is an ephemeral attribute that gets passed to the job when it is enqueued.
+  # Anything stored in here should be small and serializable.
+  attr_accessor :job_params
+
+  belongs_to :creator, class_name: "User"
```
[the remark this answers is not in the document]
DevneyHamilton: ?

app/controllers/user_batches_controller.rb
```diff
@@ -42,14 +51,14 @@ def do_import
     render("form")
   end
 
-  def operation
+  def operation(temp_file_path, original_filename)
```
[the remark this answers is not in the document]
DevneyHamilton: need to review this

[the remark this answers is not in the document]
DevneyHamilton: all fixed

app/controllers/user_batches_controller.rb
```diff
@@ -12,11 +13,21 @@ def new
     render("form")
   end
 
+  def upload
+    original_file_name = params[:userbatch].original_filename
+    temp_file_path = UploadSaver.new.save_file(params[:userbatch])
+    render(json: {tempFilePath: temp_file_path, originalFilename: original_file_name})
+  rescue StandardError => e
```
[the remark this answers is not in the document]
DevneyHamilton: used below

DevneyHamilton:
Ready for review. I have not been able to convince myself that the progress bar dropzone makes actually reflects upload progress. hoping to discuss ways to test this.

app/controllers/user_batches_controller.rb
```diff
@@ -14,11 +14,22 @@ def new
     render("form")
   end
 
+  def upload
```
scttnlsn: Should there be a manual authorization here?  i.e. something like `authorize!(:create, UserBatch)`

app/assets/javascripts/views/file_uploader_manager.js.coffee
```diff
@@ -0,0 +1,12 @@
+class ELMO.Views.FileUploaderManager extends ELMO.Views.ApplicationView
```
scttnlsn: Is this class needed?  It looks like the only use of it is as the `listener` in the uploader view.  Can we just log these messages from the uploader view?  I'm also not sure if that `events` hash does anything here.

app/models/user_batch.rb
```diff
@@ -9,7 +9,7 @@ class UserBatch
                         gender_custom nationality notes user_groups]
   EXPECTED_HEADERS =  %i[login name phone phone2 email birth_year gender nationality notes user_groups].freeze
 
-  attr_accessor :file, :mission_id, :name
+  attr_accessor :file, :mission_id, :name, :temp_file_path
```
scttnlsn: Is this used?  I noticed the form was just using `hidden_field_tag` and so might not need a corresponding model field?  Not sure.

app/assets/javascripts/views/file_uploader_view.js.coffee
```diff
@@ -3,25 +3,30 @@
...
       url: @postPath
       paramName: @paramName # The name that will be used to transfer the file
       maxFiles: 1
       uploadMultiple: false
       previewTemplate: @previewTemplate,
       thumbnailWidth: 100,
-      thumbnailHeight: 100
+      thumbnailHeight: 100,
+      acceptedFiles: @acceptedFileFormats,
+      maxFilesize: 80000 #kb, fits a maximum row xlsx file
```
smoyth: Curious about this. Strangely, the [docs](https://www.dropzonejs.com/#config-maxFilesize) say this is for limiting the _number_ of files uploaded but that doesn't make sense. 

Elsewhere in the docs there is:

```
// "myAwesomeDropzone" is the camelized version of the HTML element's ID
Dropzone.options.myAwesomeDropzone = {
  paramName: "file", // The name that will be used to transfer the file
  maxFilesize: 2, // MB
  accept: function(file, done) {
    if (file.name == "justinbieber.jpg") {
      done("Naha, you don't.");
    }
    else { done(); }
  }
};
```

which seems to say the unit is MB!?

Was this needed to overcome a failure? Do we know what the default value is?

The nginx server configuration also includes a max upload size restriction. Our default config file currently has it at 50mb.

So we should probably match that.

The thing about size checks is that there are often a crap ton of places where you can specify them and it's often tricky to keep them in sync.

Paperclip has `validates_attachment_size`, but we don't seem to be using that, which is good.

Thoughts?

app/views/user_batches/form.html.erb
```diff
@@ -1,7 +1,33 @@
...
+    el: "#file-uploader",
+    zoneId: "#dropzone-file-uploader",
+    genericThumbPath: nil,
+    postPath: "#{user_batches_path}/upload",
+    previewTemplate: render("layouts/dropzone_preview"),
+    paramName: "userbatch",
+    acceptedFileFormats: ".xlsx, .csv",
+    metaFields: {
+      "tempFilePath" => "input#temp_file_path",
+      "originalFilename" => "input#original_filename"
```
smoyth: I see what is happening here and this is a clever solution but it's is a bit unconventional. Let's talk about how to make this cleaner.

app/controllers/user_batches_controller.rb
```diff
@@ -14,11 +14,22 @@ def new
...
+    original_file_name = params[:userbatch].original_filename
+    temp_file_path = UploadSaver.new.save_file(params[:userbatch])
+    render(json: {tempFilePath: temp_file_path, originalFilename: original_file_name})
+  rescue StandardError => e
+    Rails.logger.error(e)
+    render(errors: [I18n.t("activerecord.errors.models.user_batch.internal")])
```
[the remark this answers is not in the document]
DevneyHamilton: should be json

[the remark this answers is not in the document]
DevneyHamilton: test again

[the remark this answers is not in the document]
DevneyHamilton: removed because rescuing error prevents dropzone from handling error

app/assets/javascripts/views/file_uploader_manager.js.coffee
DevneyHamilton: removed; and renamed old media uploader manager to file uploader manager

app/assets/javascripts/views/file_uploader_view.js.coffee
```diff
@@ -38,30 +44,39 @@ class ELMO.Views.FileUploaderView extends ELMO.Views.ApplicationView
...
 
   uploadStarting: ->
     if @listener?
       @listener.uploadStarting()
-    @$('img')[0].src = @genericThumbPath
+    if @genericThumbPath?
+      @$('img')[0].src = @genericThumbPath
+    else
+      @$('img').hide()
+      @$('.dz-details').css('margin', '0px')
```
DevneyHamilton: is there a cleaner alternative here besides having two separate dropzone_preview templates?

app/assets/stylesheets/local/forms/_file_upload.scss
```diff
@@ -0,0 +1,89 @@
+#upload-progress-notice {
```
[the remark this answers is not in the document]
smoyth: No.

app/controllers/user_batches_controller.rb
```diff
@@ -14,11 +14,22 @@ def new
     render("form")
   end
 
+  def upload
+    authorize!(:create, UserBatch)
+    original_file_name = params[:userbatch].original_filename
+    temp_file_path = UploadSaver.new.save_file(params[:userbatch])
+    # Json keys match hidden input names that contain the key in dropzone form.
```
smoyth: Nice

app/views/user_batches/form.html.erb
```diff
@@ -1,7 +1,35 @@
...
+    <div id="dropzone-file-uploader" class="dropzone">
+      <div class="dz-message">
+        <span class="message"><%= t("common.file_drop_message") %>&nbsp;</span>
+      </div>
+    </div>
+    <div class="dz-error-msg"></div>
+  </div>
   <div class="submit-buttons">
     <%= f.submit(t("common.import"), class: "btn btn-primary") %>
+    <%= render("file_upload/upload_progress") %>
```
smoyth: Perhaps could be more specific and call it `upload_in_progress_button`? The actual upload progress is handled by dropzone...

app/views/responses/answers/_media_editable.html.erb
```diff
@@ -21,9 +21,9 @@
     zoneId: "#dropzone-#{context.path_str}",
     genericThumbPath: image_path("media/#{answer.qtype.media_type}.png"),
     postPath: media_create_path(type: media_type),
-    previewTemplate: render("layouts/dropzone_preview"),
-    paramName: "upload"
+    previewTemplate: render("file_upload/dropzone_preview"),
+    paramName: "upload",
```
smoyth: Extra comma

app/assets/javascripts/views/file_uploader_view.js.coffee
smoyth: Yes. You can just add a class to one of the outer elements like "no-thumb" and then in the appropriate CSS file hide the image and remove the margin. Since the HTML is different in each case you can add it literally in the template instead of adding from JS using `addClass` or passing it as an option to a partial or whatever.

app/controllers/media/objects_controller.rb
```diff
@@ -29,7 +29,9 @@ def create
     media.answer = Answer.find(params[:answer_id]) if params[:answer_id]
 
     if media.save
-      render(json: {id: media.id}, status: :created)
+      # Json keys match hidden input names that contain the key in dropzone form.
+      # See ELMO.Views.FileUploaderView for more info.
+      render(json: {media_object_id: media.id}, status: :created)
```
smoyth: Dandy.

PR #533, opened 2018-11-11
9048 s3 storage

scttnlsn:


scttnlsn:
Rebased against `develop` and fixed the merge conflict

app/controllers/media/objects_controller.rb
```diff
@@ -1,79 +1,79 @@
...
-      type: @media_object.item_content_type,
-      disposition: disposition,
-      filename: media_filename)
-  end
+      authorize!(:show, @response) if @response
+
+      send_attachment(
+        @media_object.item,
+        style: style,
+        type: @media_object.item_content_type,
```
smoyth: Don't think we are using this param anymore?

app/controllers/media/objects_controller.rb
```diff
@@ -1,79 +1,79 @@
...
+        disposition: disposition,
+        filename: media_filename
+      )
```
smoyth: Can you condense this into two lines?

app/controllers/operations_controller.rb
```diff
@@ -22,7 +24,10 @@ def show
 
   def download
     if @operation.attachment.present?
-      send_file(@operation.attachment.path, filename: @operation.attachment_download_name)
+      send_attachment(
+        @operation.attachment,
+        filename: @operation.attachment_download_name
+      )
```
smoyth: Condense please

app/controllers/questions_controller.rb
```diff
@@ -78,10 +79,11 @@ def audio_prompt
 
     decorated_question = Odk::QuestionDecorator.decorate(@question)
 
-    send_file decorated_question.audio_prompt.path,
-      type: decorated_question.audio_prompt_content_type,
+    send_attachment(
+      decorated_question.audio_prompt,
       disposition: "attachment",
```
smoyth: Maybe make disposition attachment the default to simplify things further? Inline is only rarely used I think?

app/helpers/answers_helper.rb
```diff
@@ -15,16 +15,29 @@ def excerpt_if_exists(response, answer)
...
+  def media_path(object, params = {})
+    media_object_path(object, params.merge(type: object.kind.pluralize))
+  end
+
+  def thumb_path(object, params = {})
+    if object.dynamic_thumb?
+      media_path(object, params.merge(style: :thumb))
+    else
+      object.static_thumb_path
+    end
```
smoyth: This turned out nice!

app/jobs/tabular_import_operation_job.rb
```diff
@@ -19,4 +24,18 @@ def format_error_report(errors)
...
+    if file.options[:storage] == "fog"
+      # Preserve the file extension (Roo, for example, requires this in order to parse)
+      pathname = Pathname.new(file.path)
+      tmp = Tempfile.new(["tabular_import", pathname.basename.to_s])
+      tmp.binmode
+      URI.parse(file.expiring_url).open { |io| tmp.write(io.read) }
+      tmp.rewind
+      tmp
+    else
+      file
```
smoyth: Cool!

app/controllers/media/objects_controller.rb
scttnlsn: I added the following to the storage concern: `params[:type] ||= attachment.content_type` which is essentially the same as what's happening here.  I'll remove this line from the objects controller.

PR #534, opened 2018-11-14
9020 search query bug

scttnlsn:


scttnlsn:
@smoyth This just casts the value based on the database column type.  We could alternatively check that the casted value equals the original value and throw an error otherwise.  What do you think?

app/models/search/token.rb
```diff
@@ -138,6 +137,21 @@ def comparison_fragment(qualifier, column, op, value_token)
...
+          input_value = value_sql.downcase
+          bool_value = if truthy.include?(input_value)
+                         "t"
+                       elsif falsy.include?(input_value)
+                         "f"
+                       else
+                         raise_error_with_qualifier("boolean_error", qualifier.name)
+                       end
+
+          sanitize("#{column} = ?", bool_value)
```
smoyth: Wow, lovely!

PR #535, opened 2018-11-15
8933 progress bar option set import

DevneyHamilton:
Screenshots:
screen shot 2018-11-15 at 8 30 40 am
screen shot 2018-11-15 at 8 30 23 am
screen shot 2018-11-15 at 8 30 02 am

### DIFF
--- a/app/controllers/option_set_imports_controller.rb
+++ b/app/controllers/option_set_imports_controller.rb
@@ -10,11 +10,22 @@ class OptionSetImportsController < ApplicationController
     render("form")
   end
 
+  def upload
+    authorize!(:create, OptionSetImport)
+    original_file_name = params[:userbatch].original_filename
+    temp_file_path = UploadSaver.new.save_file(params[:file_import])
+    # Json keys match hidden input names that contain the key in dropzone form.
+    # See ELMO.Views.FileUploaderView for more info.
+    render(json: {temp_file_path: temp_file_path, original_filename: original_file_name})
+  end
+
   def create
-    if @option_set_import.valid?
-      do_import
+    temp_file_path = params[:temp_file_path]
+    original_filename = params[:original_filename]
+    if temp_file_path.present? && original_filename.present?
+      do_import(temp_file_path, original_filename)
     else
-      flash.now[:error] = I18n.t("activerecord.errors.models.option_set_import.general")
+      flash.now[:error] = I18n.t("errors.file_upload.file_missing")
       render("form")
     end
   end
@@ -26,8 +37,8 @@ class OptionSetImportsController < ApplicationController
 
   protected
 
-  def do_import
-    @stored_path = UploadSaver.new.save_file(@option_set_import.file)
+  def do_import(temp_file_path, original_filename)
+    operation(temp_file_path, original_filename).enqueue
     operation.enqueue
     prep_operation_queued_flash(:option_set_import)
     redirect_to(option_sets_url)
@@ -40,8 +51,8 @@ class OptionSetImportsController < ApplicationController
   def operation
     Operation.new(
       creator: current_user,
-      job_class: TabularImportOperationJob,
       mission: current_mission,
+      job_class: TabularImportOperationJob,
       details: t("operation.details.option_set_import", name: @option_set_import.name),
       job_params: {
         name: @option_set_import.name,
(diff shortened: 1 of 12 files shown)


spec/features/results/responses/file_upload_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+    expect_preview(image_node)
+
+    # delete video
+    delete_file(video_node)
+    expect_no_preview(video_node)
+
+    # save w/ user
+    select2(user.name, from: "response_user_id")
+    click_button("Save")
+    expect(page).to_not have_content("Response is invalid")
```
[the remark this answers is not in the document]
DevneyHamilton: I am confused about these instructions.

spec/support/contexts/dropzone.rb
```diff
@@ -25,6 +25,20 @@ def drop_in_dropzone(file_path, index = 0)
     JS
   end
 
+  def expect_preview(node)
+    expect(node).to have_selector(".dz-preview")
+    expect(node).to_not have_content("The uploaded file was not an accepted format.")
+  end
+
+  def expect_no_preview(node)
+    expect(node).to_not have_selector(".dz-preview")
```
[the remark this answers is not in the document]
DevneyHamilton: I am confused about these instructions. I'm used to rspec having this style without parens.

app/controllers/option_set_imports_controller.rb
```diff
@@ -1,59 +1,24 @@
...
-  include OperationQueueable
+class OptionSetImportsController < TabularImportsController
 
   load_and_authorize_resource
 
-  def new
-    render("form")
+  def tabular_class
+    OptionSetImport
   end
```
DevneyHamilton: Is there a more Rails-y way to do this, and the symbol, and the url?

app/views/user_imports/form.html.erb
```diff
@@ -1,34 +1,5 @@
 <%= elmo_form_for(@user_import, html: {multipart: true}) do |f| %>
   <p class="instructions"><%= formatted_instructions %></p>
```
smoyth: `formatted_instructions` is a bad global method name when it applies only to user imports. Can you please move the code back into the template directly as it is with option set imports? We can maybe abstract out the instructions stuff later but let's get rid of that helper method.

Then I notice the only other method in that helper is not being used so let's nix it altogether. One less helper in the world! Yay!

app/views/user_imports/form.html.erb
```diff
@@ -1,34 +1,5 @@
...
-    genericThumbPath: nil,
-    postPath: "#{user_imports_path}/upload",
-    previewTemplate: render("file_upload/dropzone_preview"),
-    paramName: "file",
-    acceptedFileFormats: ".xlsx, .csv",
-  )%>
-  options.listener = ELMO.fileUploaderManager;
-  new ELMO.Views.FileUploaderView(options);
+  <%= render("file_upload/operation_file_uploader", form: f, operation_type_prefix: "user_imports",
+                                                    post_path: user_imports_path) %>
```
smoyth: You could just pass `import_type: :user` as the symbol and then do `send("#{import_type}_imports_path")` to get the post path. And a similar thing with the form CSS selector.

config/routes.rb
```diff
@@ -169,6 +169,7 @@
 
     resource :option_set_imports, path: "option-set-imports", only: [:new, :create] do
       collection do
+        post :upload
         get "option-sets-template", as: "template", action: "template", defaults: { format: "xlsx" }
```
smoyth: This is a weird old way to define these routes. Could you please refactor to just `get :template` and quickly manually test the user batch case? I think that's all we need. There is no need for a default format.

spec/features/operations/file_upload_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  let(:admin) { create(:user, role_name: "coordinator", admin: true) }
+  let(:user_import) { Rails.root.join("spec", "fixtures", "user_imports", "batch_of_3.csv") }
+  let(:option_set_import) { Rails.root.join("spec", "fixtures", "option_set_imports", "simple.csv") }
+  let(:invalid_file) { Rails.root.join("spec", "fixtures", "media", "images", "the_swing.jpg") }
+  let(:params) { {locale: "en", mode: "m", mission_name: get_mission.compact_name} }
+
+  before { login(admin) }
+
+  scenario "uploading user batch" do
+    visit new_user_import_path(params)
```
smoyth: Parens please

DevneyHamilton:
also try taking out skip_authorization_check declarations.

PR #546, opened 2018-11-27
8947 load test 2018

scttnlsn:


app/models/utils/load_testing/dsl.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Utils
+  module LoadTesting
+    # Extends the JMeter DSL with helper methods
+    class Dsl < RubyJmeter::ExtendedDSL
+      def login(username, password)
```
scttnlsn: I implemented this before realizing I needed basic auth instead.  It's not currently used but I figured I'd leave it around since we'll likely use it in the future.

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,100 @@
...
+        when "date"
+          n = Faker::Number.number(3).to_i
+          n.days.ago.to_date.to_s
+        when "time"
+          n = Faker::Number.number(3).to_i
+          n.days.ago.to_s.split(" ")[1]
+        when "select_one"
+          question.options.rand.id
+        when "select_multiple"
+          # FIXME
```
scttnlsn: @smoyth Do you have an example of an ODK submission with a `select_multiple` value?  Not sure what that's supposed to look like.

app/models/utils/load_testing/load_test.rb
```diff
@@ -0,0 +1,63 @@
...
+
+          threads(count: options[:thread_count], duration: options[:duration]) do
+            instance_eval(&block)
+          end
+        end
+      end
+
+      # This outputs the test plan (and supporting files) to tmp/test_plans/<name>_<timestamp>/
+      # The test plan can be run by copying those files to the test machine and running
+      #   jmeter -n -t testplan.jmx
```
smoyth: Nice comment!

app/models/utils/load_testing/load_test.rb
```diff
@@ -0,0 +1,63 @@
...
+      def generate
```
smoyth: Can we call this `generate_plan` to disambiguate from `generate_test_data`?

app/models/utils/load_testing/load_test.rb
```diff
@@ -0,0 +1,63 @@
...
+        @timestamp = Time.zone.now
+        FileUtils.mkdir_p(path)
+
+        generate_test_data
+        plan.jmx(file: path.join("testplan.jmx"))
+        path
+      end
+    end
+  end
+end
```
smoyth: This is a super nice abstract parent class! Great job!

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+module Utils
+  module LoadTesting
+    # Builds a JMeter test plan for submitting a lot of ODK form submissions.
+    #
+    # This test building process should be run on the server that is being targeted
+    # for the test because it will query the database while building the sample submission
+    # data (i.e. the XML for the ODK submission) for the test.
+    #
+    # The actual JMeter test will be run on a different server that does not have NEMO installed.
+    # It wil just know how to make a bunch of HTTP requests based on the options and data we give it.
```
smoyth: s/wil/will/ (My fault I think)

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      #   form_id: The ID of the form being submitted to.
+      #   mission_name: The name of the mission in which the form is being submitted
+      #
+      # The test currently knows how to handle the following question types:
+      #   text, long_text, barcode, integer, counter, decimal, location, select_one,
+      #   select_multiple, datetime, date, time
+
+      SUBMISSION_FILENAME = "submission.xml"
+
+      # Generates a bunch of test data to be used to make the test requests and store it in a CSV file.
```
smoyth: s/store/stores/

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+    # data (i.e. the XML for the ODK submission) for the test.
+    #
+    # The actual JMeter test will be run on a different server that does not have NEMO installed.
+    # It wil just know how to make a bunch of HTTP requests based on the options and data we give it.
+    class OdkSubmissionLoadTest < LoadTest
+      # Required `options`:
+      #   username: User doing the submitting
+      #   password: User's password
+      #   form_id: The ID of the form being submitted to.
+      #   mission_name: The name of the mission in which the form is being submitted
```
smoyth: I think we can get the mission from the form. Can you do so and remove this requirement?

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      # The test currently knows how to handle the following question types:
+      #   text, long_text, barcode, integer, counter, decimal, location, select_one,
+      #   select_multiple, datetime, date, time
+
+      SUBMISSION_FILENAME = "submission.xml"
+
+      # Generates a bunch of test data to be used to make the test requests and store it in a CSV file.
+      def generate_test_data
+        form = Form.find(options[:form_id])
+        data = odk_submission(form)
```
smoyth: Are we just generating one submission? Can we do a few hundred perhaps? I'd rather not have them all be completely identical...

I also note we're not doing a CSV file. I guess that makes sense given the way we have to attach files to the submission. But I think to allow us to do multiple different submissions and keep it tractable we could use CSV and construct our own multipart request manually. See: https://stackoverflow.com/a/39020448/2066866

So to recap:

* CSV file full of lines of XML, each line being a full submission
* In the `test` block, we construct a multipart POST request setting `xml_submission_file` to the next line from the CSV.
* JMeter automatically wraps up to the top of the CSV file when it reaches the end.

Doable?

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+          n.days.ago.to_s.split(" ")[1]
+        when "select_one", "select_multiple"
+          "on#{question.options.rand.option_nodes.first.id}"
+        when "barcode"
+          Faker::Code.ean
+        when "location"
+          lat = Faker::Address.latitude
+          lng = Faker::Address.longitude
+          "#{lat} #{lng}"
+        end
```
smoyth: Very nice!

app/models/utils/load_testing/odk_submission_load_test.rb
```diff
@@ -0,0 +1,96 @@
...
+
+      def odk_submission(form)
+        items = form.preordered_items.map { |i| Odk::DecoratorFactory.decorate(i) }
+
+        data = items.map do |item|
+          value = test_value(item.question)
+          "<#{item.odk_code}>#{value}</#{item.odk_code}>"
+        end.join("\n")
+
+        "<?xml version='1.0' ?><data id='#{form.id}' version='#{form.code}'>\n#{data}\n</data>"
```
smoyth: Have you tested the output of this a little by POSTing it manually? Just curious...

lib/tasks/load_test.rake
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+namespace :load_test do
+  desc "Generate load test plan for ODK submission"
+  task odk: :environment do
+    generate_test(Utils::LoadTesting::OdkSubmissionLoadTest,
+      username: ENV["USERNAME"],
+      password: ENV["PASSWORD"],
+      mission_name: ENV["MISSION_NAME"],
```
smoyth: Remove.

lib/tasks/load_test.rake
```diff
@@ -0,0 +1,22 @@
...
+      form_id: ENV["FORM_ID"])
+  end
```
smoyth: This looks perfect! Terrific!

spec/models/utils/load_testing/odk_submission_load_test_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  before { form.publish! }
+
+  it "generates expected test plan" do
+    test = described_class.new(
+      threads: threads,
+      duration: duration,
+      username: username,
+      password: password,
+      mission_name: mission.compact_name,
```
smoyth: Remove.

spec/models/utils/load_testing/odk_submission_load_test_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      form_id: form.id
+    )
+
+    path = test.generate
+    actual_content = File.read(path.join("testplan.jmx"))
+    expected_content = fixture_file("test_plans/odk_submission.jmx")
+
+    expect(without_timestamps(actual_content)).to eq(without_timestamps(expected_content))
```
smoyth: This is super cool! If my request to use CSV with XML causes this method to break somewhat, I wouldn't stress out too much and it doesn't have to test every last char of the test plan. The main thing is that we are just running the generating the test plan in the spec suite and not getting any errors.

spec/models/utils/load_testing/odk_submission_load_test_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      counter
+      decimal
+      location
+      select_one
+      select_multiple
+      datetime
+      date
+      time
+      barcode
+    ])
```
smoyth: Can you shorten this to just 2-3 lines please.

spec/fixtures/test_plans/odk_submission.jmx
```diff
@@ -0,0 +1,98 @@
...
+                </elementProp>
+              </collectionProp>
+            </elementProp>
+          </HTTPSamplerProxy>
+          <hashTree/>
+        </hashTree>
+      </hashTree>
+    </hashTree>
+  </hashTree>
+</jmeterTestPlan>
```
smoyth: Wow!

app/models/utils/load_testing/odk_submission_load_test.rb
scttnlsn: I have only tested it by POSTing it via JMeter but it does work as expected.  I see the submissions when I browse my locally running ELMO instance.

scttnlsn:
I'm trying to get the multipart submission working but running into an issue with the line separators in the multipart request body.  I realized that the spec for this requires CRLF characters: https://tools.ietf.org/html/rfc1867#section-5.9

I can successfully submit a multipart ODK submission with `curl` on my local machine:

```
curl -X POST http://localhost:3000/en/m/fakemission640/submission \
  -H "Authorization: Basic YWRtaW46VGVzdGluZzEyMw==" \
  -H "Content-Type: multipart/form-data; boundary=BOUNDARY" \
  --data-binary $'--BOUNDARY\r\nContent-Disposition: form-data; name="xml_submission_file"; filename="submission.xml"\r\nContent-Type: text/xml\r\n\r\nXMLHERE\r\n--BOUNDARY--'
```
(I omitted the actual XML content for readability)

The problem is that JMeter writes the request body to the XML test plan file (which does not include the carriage returns on my machine) and the request does not work when the line separators are just a LF character.  I've tried various ways of escaping the CRLF- `\r\n`, `\\r\\n`, `\\\r\\\n` and `&#13;&#10;` (XML char entities), but no luck with any of those.  Rails still complains about a "Bad request content body".

@smoyth Any suggestions?

scttnlsn:
If we can't get the manual multipart submission working then we could do the following:

* generate 100 unique XML submission files (each with a unique file name)
* generate a CSV that includes all 100 file names
* use the filenames from the CSV data to submit the appropriate file in the test plan

Might be annoying to have to transfer so many files to a test machine.

PR #547, opened 2018-12-03
9120 sms submission load test

scttnlsn:


app/models/utils/load_testing/sms_submission_load_test.rb
```diff
@@ -0,0 +1,108 @@
...
+      # * The Twilio adapter must be configured appropriately on the target
+
+      CSV_FILENAME = "submissions.csv"
+
+      # Generates a bunch of test data to be used to make the test requests and stores it in a CSV file.
+      def generate_test_data
+        CSV.open(path.join(CSV_FILENAME), "wb") do |csv|
+          csv << ["message_body"]
+
+          100.times do
```
smoyth: We will need a lot more than 100 here because duplicate SMSes are rejected within a time window of 12 hours. So basically if we want to send 200k SMSes we really need 200k lines in this file! Can you perhaps make the number of sample responses an argument to both of these tasks that defaults to 100?

app/models/utils/load_testing/sms_submission_load_test.rb
```diff
@@ -0,0 +1,108 @@
...
+      # Required `options`:
+      #   user_id: User doing the submitting
+      #   form_id: The ID of the form being submitted to.
+      #
+      # The test currently knows how to handle the following question types:
+      #   text, long_text, integer, counter, decimal, select_one,
+      #   select_multiple, datetime, date, time
+      #
+      # * The submitting user must have a phone number and SMS auth code
+      # * The Twilio adapter must be configured appropriately on the target
```
smoyth: Not sure if we can use Twilio because we would have to sign each message correctly for it to work. We could do something like:

```
validator = Twilio::Util::RequestValidator.new("xyz")
validator.build_signature_for("#{configatron.url.protocol}://#{configatron.url.host}/foo", {"From" => "+123456776", "To" => "+1234567890", "Body" => "abc aldflat"}) 
```

for each message and store that in the CSV file. It seems fraught though. Another option would be to use the frontline cloud adapter which doesn't do any inbound validation.

app/models/utils/load_testing/sms_submission_load_test.rb
```diff
@@ -0,0 +1,108 @@
...
+        mission_name = form.mission.compact_name
+        url_token = form.mission.setting.incoming_sms_token
+        params = {"From" => user.phone, "To" => "+1234567890", "Body" => "${message_body}"}
+
+        test do
+          csv_data_set_config(filename: CSV_FILENAME, variableNames: "message_body")
+          header([{name: "X-Twilio-Signature", value: "1"}])
+
+          transaction("post_sms_response") do
+            submit("/m/#{mission_name}/sms/submit/#{url_token}",
```
smoyth: If you make this one line it will silence the hound I think.

app/models/utils/load_testing/sms_submission_load_test.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      private
+
+      def form
+        Form.find(options[:form_id])
+      end
+
+      def user
+        User.find(options[:user_id])
+      end
```
smoyth: Let's memoize these two methods to avoid lookups over and over again?

app/models/utils/load_testing/sms_submission_load_test.rb
```diff
@@ -0,0 +1,108 @@
...
+          n = question.options.length
+          options = ("a".."z").to_a[0...n]
+          options.rand
+        else
+          raise ArgumentError, "unsupported question type: #{question.qtype_name}"
+        end
+      end
+
+      def sms_submission(form)
+        items = form.preordered_items.map { |i| Odk::DecoratorFactory.decorate(i) }
```
smoyth: Don't think we need the ODK decorator?

spec/models/utils/load_testing/sms_submission_load_test_spec.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe Utils::LoadTesting::SmsSubmissionLoadTest do
+  let(:setting) { FactoryGirl.build(:setting, incoming_sms_token: "token") }
```
smoyth: Don't need `FactoryGirl` literal here or in the other spec. Please remove from both.

spec/models/utils/load_testing/sms_submission_load_test_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    )
+
+    path = test.generate_plan
+    actual_content = File.read(path.join("testplan.jmx"))
+    expected_content = fixture_file("test_plans/sms_submission.jmx")
+
+    expect(without_timestamps(actual_content)).to eq(without_timestamps(expected_content))
+  end
+
+  def without_timestamps(content)
```
smoyth: Can we move this into a shared "load testing" context please.

PR #549, opened 2018-12-06
8866 only send password reset email when requesting new password reset

scttnlsn:


scttnlsn:
@smoyth Specs fixed

scttnlsn:
@smoyth The specs are failing w/o those load test changes.  Should I include them in a separate PR?

smoyth:
OK I merged. In future yes a fix to a broken build should be in its own PR and should get the highest priority to minimize effects on other devs.

PR #554, opened 2018-12-12
8566 clone form item hidden attribute

scttnlsn:


scttnlsn:
I couldn't reproduce the issue w/ names not being cloned.  Wrote some tests to demonstrate that it is working as expected.

PR #556, opened 2018-12-12
9150: reworked the map position in a geographic response

jpdimond:
It used to be a float, but thought it would be better just inline, rather than to the right. See before and after.
Before:
image

After:
image

### DIFF
--- a/app/assets/stylesheets/local/responses/hierarchical_form/_question_types.scss
+++ b/app/assets/stylesheets/local/responses/hierarchical_form/_question_types.scss
@@ -1,10 +1,10 @@
 .response-form {
   .answer-map {
     display: none;
-    float: $right;
-    height: 80px;
+    height: 90px;
     margin-bottom: 10px;
-    width: 160px;
+    margin-top: 10px;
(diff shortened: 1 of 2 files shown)


smoyth:
Looks great. Thanks for the screenshot! Couple small things:

1. What is with the giant +/- buttons? Can we make them smaller or perhaps make the map a touch bigger?
2. No need for bottom-margin on map since the box has padding. See how the left and bottom margins are not the same:

image

I think it looks more polished and nice when they match.

jpdimond:
> Looks great. Thanks for the screenshot! Couple small things:
> 
> 1. What is with the giant +/- buttons? Can we make them smaller or perhaps make the map a touch bigger. It's quite small ATM.
> 2. No need for bottom-margin on map since the box has padding. See how the left and bottom margins are not the same:
> 
> image
> 
> I think it looks more polished and nice when they match.

The option to make the zoom controls smaller was removed from the google maps api, but I did make the map a little bigger. 

image

2. removed bottom margin, good catch.

PR #557, opened 2018-12-13
9088 upgrade to Rails 5.2.2

scttnlsn:


scttnlsn:
Lots of warnings coming from `acts_as_list`:

> DEPRECATION WARNING: Dangerous query method (method whose arguments are used as raw SQL) called with non-attribute argument(s): "\"form_items\".\"rank\" DESC". Non-attribute arguments will be disallowed in Rails 6.0. This method should not be called with user-provided values, such as request parameters or model attributes. Known-safe values can be passed by wrapping them in Arel.sql(). (called from bottom_item at /home/scott/.rbenv/versions/2.4.3/lib/ruby/gems/2.4.0/bundler/gems/acts_as_list-20fa19384f53/lib/acts_as_list/active_record/acts/list.rb:283)

config/storage.yml
```diff
@@ -0,0 +1,34 @@
+test:
```
smoyth: ooooo!

app/models/user.rb
```diff
@@ -60,14 +60,18 @@ class User < ApplicationRecord
...
   # which can be an issue if they will be being re-assigned
   # validate(:must_have_assignments_if_not_admin)
-  validates :password, format: { with: PASSWORD_FORMAT,
-                                 if: :require_password?,
-                                 message: :invalid_password }
+  validates(:password,
+    confirmation: {if: :require_password?},
+    format: {with: PASSWORD_FORMAT, if: :require_password?, message: :invalid_password},
+    length: {minimum: 8, if: :require_password?})
+  validates(:password_confirmation, length: {minimum: 8}, if: :require_password?)
```
smoyth: I think we need specs for these validations. Could borrow some from Gather.

spec/features/users/user_form_spec.rb
```diff
@@ -30,7 +30,7 @@
 
       # Make sure password email url is correct and no missing translations
       email = ActionMailer::Base.deliveries.last
-      expect(email.body.to_s).to match(%r{^https?://.+/en/password-resets/\w+/edit$})
+      expect(email.body.to_s).to match(%r{^https?://.+/en/password-resets/[\w_-]+/edit$})
```
smoyth: Nice fix!

PR #559, opened 2018-12-17
9054 added summary report support for all qtypes

scttnlsn:


scttnlsn:
Should I include similar specs for all question types?

app/models/report/summary_collection_builder.rb
```diff
@@ -3,16 +3,24 @@ class Report::SummaryCollectionBuilder
...
+    time: "stat",
+    datetime: "stat",
+    select_one: "select",
+    select_multiple: "select",
+    date: "date",
+    text: "raw",
+    long_text: "raw",
+    barcode: "raw",
+    location: "raw",
+    image: "raw",
```
smoyth: What will image/video/audio summaries look like? Seems this would need more work. Thoughts?

scttnlsn: @smoyth It looks like it's not even possible to report on image/video/audio columns (those qtypes are not available in the report builder popup).

scttnlsn: Should I just remove them here?  Or do we want to include them in the report somehow?

PR #563, opened 2019-01-07
9226: remove multi select to span all objects

jpdimond:


smoyth:
I think we definitely need feature spec coverage on this b/c it has caused so much heartache. Did you check with Mel to see what she has done on this? I also want to consider the other batch actions across the app and ensure they're all properly tested and working. There is also the question of changing the select-all interaction to be less surprise-y. Can we discuss?

PR #565, opened 2019-01-09
9254: Change form.responses_count from counter cache to query

smoyth:


spec/models/forms/responses_count_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+require "rails_helper"
+
+describe "Form#responses_count" do
+  let(:user) { create(:user) }
+  let!(:form1) { create(:form, question_types: %w[integer]) }
+  let!(:form2) { create(:form, question_types: %w[integer]) }
+  let!(:form3) { create(:form, question_types: %w[integer]) }
+  let(:forms_by_id) { scope.index_by(&:id) }
+
+  shared_examples_for "correct counts" do |responses_counts, query_count|
```
jpdimond: should this be in a context file?

smoyth: Since it's only used in one place, I don't think so. If/when it gets further abstracted and reused it could be moved.

PR #566, opened 2019-01-14
9314: update scrypt gem, update development setup instructions to fix…

DevneyHamilton:
… issues encountered setting up dev env on Mojave Mac


lib/themeing/scss_preprocessor.rb
```diff
@@ -37,7 +37,7 @@ def original_scss
 
     def replace_top_comment(scss)
       lines = scss.split("\n")
-      loop { lines[0].match?(%r{\A(//|\w*\z)}) ? lines.shift : break }
+      loop { lines[0].match(%r{\A(//|\w*\z)}) ? lines.shift : break }
```
DevneyHamilton: I am mystified by this. `match` was updated to `match?` between ruby 2.3.x and ruby 2.4.x. But this errored when running with `bundle exec` with ruby at 2.4. Is there any explanation for that?

DevneyHamilton: turns out there is - bundler was using 2.3.x

docs/development-setup.md
```diff
@@ -126,6 +126,7 @@ git checkout develop
 #### Start the server
 
 For a development setup, run `nvm use && rails s`.
+You may want to run `bundle exec rake db:create_fake_data` to create a sample mission.
```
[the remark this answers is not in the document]
DevneyHamilton: yes it does :)

PR #577, opened 2019-02-14
9395 Set up basic Jest/Enzyme tests for React

cooperka:
Fairly simple change overall, but you can read commit-by-commit if you want to see the process.

- [x] Jest framework
- [x] Enzyme toolkit
- [x] Basic working snapshot test with component interaction
- [ ] Integrate with CI (@smoyth wanted to handle this)

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,6 @@
     [
       "env",
       {
-        "modules": false,
         "targets": {
           "browsers": "> 1%",
           "uglify": true
--- a/.babelrc
+++ b/.babelrc
@@ -1,25 +1,81 @@
 {
-  "presets": [
-    [
-      "env",
-      {
-        "targets": {
-          "browsers": "> 1%",
-          "uglify": true
-        },
-        "useBuiltIns": true
-      }
-    ],
-    "react"
-  ],
-  "plugins": [
-    "syntax-dynamic-import",
-    "transform-object-rest-spread",
-    [
-      "transform-class-properties",
-      {
-        "spec": true
-      }
-    ]
-  ]
+  "env": {
+    "production": {
+      "presets": [
+        [
+          "env",
+          {
+            "modules": false,
+            "targets": {
+              "browsers": "> 1%",
+              "uglify": true
+            },
+            "useBuiltIns": true
+          }
+        ],
+        "react"
+      ],
+      "plugins": [
+        "syntax-dynamic-import",
+        "transform-object-rest-spread",
+        [
+          "transform-class-properties",
+          {
+            "spec": true
+          }
+        ]
+      ]
+    },
+    "development": {
+      "presets": [
+        [
+          "env",
+          {
+            "modules": false,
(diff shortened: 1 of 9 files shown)


.babelrc
```diff
@@ -3,7 +3,6 @@
     [
       "env",
       {
-        "modules": false,
```
cooperka: Explanation: Jest expects `import` statements to be transpiled (see [jest issue](https://github.com/facebook/jest/issues/3202#issuecomment-290967270) for more context). I doubt this affects much but I wanted to point it out.

Instead of changing all envs, we could instead add a separate `test` env if needed.

smoyth: Does this mean if I try to use eg `import defaultExport from "module-name";` in some application code it won't work?

I worry this will lead to a really confusing bug down the road. Could you go ahead with the testing environment plan? Or if I've misunderstood this, help me out...

.eslintrc.yml
```diff
@@ -1,5 +1,6 @@
 env:
   es6: true
+  jest: true
```
smoyth: Yay linter!

spec/javascript/components/__snapshots__/FormSelect.test.js.snap
```diff
@@ -0,0 +1,73 @@
...
+  >
+    First
+  </option>
+  <option
+    key="item2"
+    value="item2"
+  >
+    Second
+  </option>
+</select>
```
smoyth: Neato

.babelrc
cooperka: ES6 imports should work with ES5 exports; this should behave as expected in this case. I'll add a `test` env just in case; we may as well have it for the future anyway.

cooperka: New envs are working.

I couldn't get either Jest or Webpacker to recognize `.babelrc.js` or `babel.config.js` (see relevent [babel docs](https://babeljs.io/docs/en/config-files#configuration-file-types) and [jest docs](https://jestjs.io/docs/en/23.x/getting-started.html#using-babel)).

That would have saved some duplication since it would support variables. I'm not sure why it isn't working because it's a common convention, but all it gives us is a more DRY babel config so it's probably not worth banging my head any longer.

cooperka:
@smoyth I think tests are failing now because some test suite isn't setting a NODE_ENV (or it's something other than `test` or `production`). Tests worked both before and after removing `modules: false` from the config, so it must be because of the env change.

Can you point me to the command that runs the test that gave this output on CI:

```console
------------ BROWSER LOGS -------------
SEVERE 2019-02-15 17:48:45 +0000: http://127.0.0.1:44746/packs-test/application-d44347d2ad8823ff9d55.js 23362:16 Error: Module build failed: SyntaxError: Unexpected token (20:6)
  18 |     const displayIfOptions = ["always", "all_met", "any_met"];
  19 |     return displayIfOptions.map((option) => (
> 20 |       <option
     |       ^
```

PR #584, opened 2019-03-11
9396 form filter

cooperka:
This adds a new `Filters` component containing a `FormFilter` button.

⚠️ Currently the search bar is not contained within `Filters` so they don't "play nice" together. That should be done before release as part of [9402 (Kevin)](https://redmine.sassafras.coop/issues/9402).

⚠️ This will also require the "advanced search" change from [9393 (Tom)](https://redmine.sassafras.coop/issues/9393) in order to remove the now-obsolete `form:(foo|bar)` text from the advanced searchbox. We should probably prioritize this separately from the rest of that issue.


.eslintrc.yml
```diff
@@ -34,7 +36,6 @@ rules:
   eqeqeq: error
   func-call-spacing: error
   func-name-matching: error
-  func-style: error
```
cooperka: How intentional are these rules; is it fine to modify as I see fit or do you feel strongly about them?

I've had success with the [Airbnb linter config](https://github.com/airbnb/javascript) which has become a bit of a standard within React; we could discuss if you're interested.

app/views/searches/_form.html.erb
```diff
@@ -1,4 +1,21 @@
...
+        id: "2",
+        displayName: "Form B"
+      },
+      {
+        id: "3",
+        displayName: "Form C"
+      }
+    ],
+    selectedFormIds: ["2"]
+  }) %>
```
cooperka: This is only an example; eventually this state needs to be provided dynamically. I assume we'll wait to merge until that's ready too?

package.json
```diff
@@ -1,8 +1,12 @@
 {
   "scripts": {
+    "i18nExport": "bundle exec rake i18n:js:export",
     "test": "jest"
   },
   "jest": {
+    "roots": [
+      "<rootDir>/spec/javascript"
```
cooperka: This lets Jest check for a `__mocks__` directory under `spec/javascript` in order to mock node_modules outside of `setupTests.js`. This is quite common ([docs](https://jestjs.io/docs/en/manual-mocks)).

package.json
```diff
@@ -12,15 +16,21 @@
   "dependencies": {
     "@rails/webpacker": "3.5",
     "babel-preset-react": "^6.24.1",
+    "jquery": "3.3.1",
+    "lodash": "4.17.11",
     "prop-types": "^15.6.1",
     "react": "^16.4.1",
+    "react-bootstrap": "0.32.4",
     "react-dom": "^16.4.1",
+    "react-select2-wrapper": "cooperka/react-select2-wrapper#2019-03-08-select2Full",
```
cooperka: I forked it to allow using `select2.full.js` which is needed for using config such as `dropdownCssClass`; if you want the fork to be under TCC let me know.

So far no reply ([PR here](https://github.com/rkit/react-select2-wrapper/pull/89)).

package.json
```diff
@@ -12,15 +16,21 @@
...
     "react_ujs": "^2.4.4"
   },
   "devDependencies": {
+    "babel-eslint": "10.0.1",
```
cooperka: Needed for ESLint to understand certain code.

spec/javascript/components/search/Filters.test.js
```diff
@@ -0,0 +1,32 @@
...
+
+it("renders as expected", () => {
+  const wrapper = shallow(<Component {...defaultProps} />);
+  expect(wrapper).toMatchSnapshot();
+});
+
+describe("integration", () => {
+  suppressErrors(STUB_COMPONENT_WARNINGS);
+  const wrapper = mount(<Component {...defaultProps} />);
+  unsuppressAllErrors();
```
cooperka: See explanation [down here](https://github.com/thecartercenter/nemo/pull/584/files#diff-a236e2044e9c892d4b1ad43a22b38c18R6).

app/javascript/components/search/FormFilter.js
```diff
@@ -0,0 +1,77 @@
...
+import OverlayTrigger from "react-bootstrap/lib/OverlayTrigger";
+import Select2 from "react-select2-wrapper/lib/components/Select2.full";
+
+import "react-select2-wrapper/css/select2.css";
+
+/**
+ * Converts a list of forms from the backend into something Select2 understands.
+ */
+const parseFormsForSelect2 = (allForms) => allForms
+  .map((form) => mapKeys(form, (value, key) => key === "displayName" ? "text" : key));
```
smoyth: Why is this outside of the class? Seems like something that would be a private method. Is this how private methods are accomplished in React?

app/javascript/components/search/FormFilter.js
```diff
@@ -0,0 +1,77 @@
...
+      <Popover
+        className="filters-popover"
+        id="form-filter">
+        <Select2
+          data={parseFormsForSelect2(allForms)}
+          onSelect={onSelectForm}
+          options={{
+            placeholder: I18n.t("filter.chooseForm"),
+            dropdownCssClass: "filters-select2-dropdown",
+            width: "100%",
```
smoyth: Trailing comma on purpose?

app/javascript/components/search/FormFilter.js
```diff
@@ -0,0 +1,77 @@
...
+          </Button>
+        </div>
+      </Popover>
+    );
+  }
+
+  render() {
+    return (
+      <OverlayTrigger
+        containerPadding={25}
```
smoyth: Seems wrong to have style details like this in here. Or am I misunderstanding what this is...

app/views/searches/_form.html.erb
```diff
@@ -1,4 +1,21 @@
 <form class="search_form form-inline">
+  <%= react_component("search/Filters", {
+    allForms: [
+      {
+        id: "1",
+        displayName: "Form A"
```
smoyth: I think this should just be `name`.

config/locales/en/main.yml
```diff
@@ -9,6 +9,7 @@ en:
     add: "Add"
     and: "and"
     AND: "AND"
+    apply: "Apply"
```
smoyth: Good spot for this.

docs/development-setup.md
```diff
@@ -136,6 +136,16 @@ You may want to run `bundle exec rake db:create_fake_data` to create a sample mi
...
 
+### Dealing with JavaScript
+
+NEMO includes several React components that are provided via [react-rails](https://github.com/reactjs/react-rails).
+
+#### Tests
+
+You can run JS-only tests with `yarn test`.
+
+These rely on separately compiled i18n translations which can be generated via `rails i18n:js:export`.
```
smoyth: Thanks for adding this stuff. I think it should go above with the main 'testing' section though. Perhaps the rspec tests can have one set of nested bullets and yarn another. So:

```
### Testing

1. nvm use
1. rspec
    1. blah
    1. blah
1. yarn
    1. blah
    1. blah
```

package.json
smoyth: The other forks are under sassafras so let's put it there.

I forked to https://github.com/sassafrastech/react-select2-wrapper and gave you access.

spec/javascript/setupTests.js
```diff
@@ -1,6 +1,38 @@
...
+
+// Provide translations.
+// eslint-disable-next-line no-undef
+window.I18n = I18n;
+
+try {
+  require("../../app/assets/javascripts/i18n/translations");
+} catch (error) {
+  // eslint-disable-next-line no-console
+  console.error("Failed to find translations. Did you run `rails i18n:js:export` yet?");
```
smoyth: Nice touch!

app/javascript/components/search/FormFilter.js
cooperka: This is effectively a private method (though JavaScript doesn't have that concept), but more importantly, it's static.

I could have instead declared it a `static` class method but that's messier to factor out. I like putting utils outside of the class for ease of future refactoring. As soon as some other component wants to use this logic, it can move into a shared `utils.js` or something.

app/javascript/components/search/FormFilter.js
cooperka: Yes for better `git diff`. The linter will soon enforce this too.

app/javascript/components/search/FormFilter.js
cooperka: I agree 😢 Bootstrap requires this. This adds a pseudo-margin to the overlay itself so it doesn't hit the side of the screen.

cooperka:
@smoyth all feedback is addressed

PR #585, opened 2019-03-15
9402 advanced search filter

cooperka:
Integrates the existing search box into the new Filters component.


spec/javascript/components/search/__snapshots__/AdvancedSearchFilter.test.js.snap
```diff
@@ -0,0 +1,77 @@
...
+<div>
+  <input
+    autoComplete="off"
+    className="form-control"
+    id="search_str"
+    name="search"
+    onChange={[MockFunction]}
+    placeholder="Advanced search"
+    type="text"
+    value="query"
```
smoyth: Shouldn't this be `foo`? I'm not seeing `foo` anywhere in this snapshot even though that's the query param...

app/javascript/components/search/Filters.js
```diff
@@ -38,6 +47,14 @@ class Filters extends React.Component {
     this.setState({selectedFormIds: []});
   }
 
+  handleChangeAdvancedSearch(event) {
+    this.setState({advancedSearchText: event.target.value});
+  }
+
+  handleClearFilters() {
+    submitSearch(null);
```
smoyth: When clearing search, can we remove the search query string param (and even the `?` if it's the only param) please? I'm sure there is an npm package that will do that for you!

spec/javascript/components/search/__snapshots__/AdvancedSearchFilter.test.js.snap
smoyth: OH RIGHT. I guess this snapshot will change once we integrate with the backend.

smoyth:
@cooperka Hmm, it seems there are a lot of CI failures. We will need to figure those out...

spec/javascript/components/search/__snapshots__/AdvancedSearchFilter.test.js.snap
cooperka: Actually this value is coming from the test mock in `spec/.../search/utils.js` which defines `advancedSearchProps`. The AdvancedSearchFilter component itself gets props directly, it doesn't care what the actual url querystring is.

cooperka:
@smoyth tests should finally be passing. There were a lot of weird inconsistencies with the way Ruby talks to React, and also with the way Capybara manipulates the DOM.

I squashed the hell out of the commits, hopefully the solution is fairly readable now. Ready for review.

PR #588, opened 2019-03-29
9538 react refactor

cooperka:
⚠️ Currently based on `9454_upgrade_bootstrap` (view [real diff](https://github.com/thecartercenter/nemo/compare/9454_upgrade_bootstrap...9538_react_refactor)).

Major changes to be aware of:
- Use [Airbnb's conventions](https://github.com/airbnb/javascript) for our ESLint config
    - Specifically, this uses the [eslint-config-cooperka](https://github.com/cooperka/eslint-config-cooperka) package which is slightly less strict about inane things
- Explicitly pick props to use everywhere (instead of setting `state = props`)
- Bind class methods inline (instead of in constructor)
- Upgrade Jest to v24 (with some minor tweaks in order to keep using old Babel 6)


cooperka:
Note: Hound bot requires manual whitelisting for every config. I'll submit an issue soon. We can also copy/paste the entire config into NEMO but I hope that's not necessary.

PR #596, opened 2019-04-23
9562 opt set update performance

smoyth:
Fixes an issue where updating an option in an option set that is used in many many responses was crashing the server due to poor performance.

### DIFF
--- a/spec/models/results/answer_search_vector_updater_spec.rb
+++ b/spec/models/results/answer_search_vector_updater_spec.rb
@@ -16,7 +16,7 @@ describe Results::AnswerSearchVectorUpdater do
     option.update!(name_en: "Kitty") # This calls our class in a callback.
     response.reload
     expect(response.c[0].tsv).to eq("'kitty':1")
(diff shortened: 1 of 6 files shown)


spec/models/results/answer_search_vector_updater_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+  end
+  let!(:response) { create(:response, form: form, answer_values: ["Cat", %w[Cat Dog], "Cat"]) }
+  let(:option) { form.c[0].option_set.c[0].option }
+
+  it "updates search index tsv on both answers but not decoy" do
+    option.update!(name_en: "Kitty") # This calls our class in a callback.
+    response.reload
+    expect(response.c[0].tsv).to eq("'kitty':1")
+    expect(response.c[1].tsv).to match(/\A'dog':\d 'kitty':\d\z/)
```
cooperka: I'd never seen `\A ... \z` before, is that a Ruby thing instead of `^ ... $`?

PR #599, opened 2019-05-08
 9546: Email mission coordinators when sms token changes

cooperka:
Example email:

Screen Shot 2019-05-08 at 14 29 59


app/mailers/notifier.rb
```diff
@@ -15,6 +15,13 @@ def intro(user, mission: nil)
          subject: t("notifier.welcome", site: Settings.site_name))
   end
 
+  def sms_token_change_alert(mission)
+    @mission = mission
+    @site_name = Settings.site_name
+    mail(to: reply_to(mission), reply_to: reply_to(mission),
+         subject: t("notifier.sms_token_change.subject", mission_name: mission.name))
```
cooperka: Currently this is sent to all mission coordinators as specified in the `reply_to` method. Is "admin" a different type of user entirely or is this okay?

Is there guaranteed to be at least one coordinator, assuming this token is being regenerated by a user (in other words, can someone besides a coordinator regenerate it)?

app/mailers/notifier.rb
```diff
@@ -15,6 +15,13 @@ def intro(user, mission: nil)
          subject: t("notifier.welcome", site: Settings.site_name))
   end
 
+  def sms_token_change_alert(mission)
+    @mission = mission
```
smoyth: Can we have a guard clause here requiring mission be non-nil since the recipients can't be determined otherwise.

config/locales/en/main.yml
```diff
@@ -1427,6 +1427,9 @@ en:
     password_reset_instructions: "Password Reset Instructions"
     request_made: "A request to reset your %{site} password has been made. If you did not make this request, simply ignore this email."
     welcome: "Welcome to %{site}!"
+    sms_token_change:
+      subject: "Incoming SMS Token changed for %{mission_name}"
```
smoyth: Nice choice to use mission name here.

config/locales/en/main.yml
```diff
@@ -1427,6 +1427,9 @@ en:
...
+      body: "Notice: The Incoming SMS Token for %{site_name} - %{mission_name} was regenerated by you or another coordinator under your mission settings. This means you will need to update the incoming SMS URL with your gateway provider before new SMS submissions can be accepted."
```
smoyth: Let's remove 'Notice: '? I don't think we use that elsewhere?

app/mailers/notifier.rb
smoyth: This makes sense but can we change the method name to e.g. `coordinator_emails`.

There is also a risk that an admin may not be a member of a mission that they are actively involved in because we don't make it obvious that they're not. It might be a good idea to add a kind of notification to the UI if this is the case... Let's ask this tomorrow at standup.

app/views/notifier/sms_token_change_alert.erb
```diff
@@ -0,0 +1 @@
+<%= t("notifier.sms_token_change.body", site_name: @site_name, mission_name: @mission.name) %>
```
smoyth: Filename should be `.text.erb` instead of `.erb`.

config/locales/en/main.yml
```diff
@@ -1427,6 +1427,9 @@ en:
     password_reset_instructions: "Password Reset Instructions"
     request_made: "A request to reset your %{site} password has been made. If you did not make this request, simply ignore this email."
     welcome: "Welcome to %{site}!"
+    sms_token_change:
```
smoyth: Nice job finding the right place to put the YML.

spec/mailers/notifier_spec.rb
```diff
@@ -39,4 +39,24 @@
...
+      let!(:coordinator1) { create(:user, role_name: :coordinator, mission: mission) }
+      let!(:coordinator2) { create(:user, role_name: :coordinator, mission: mission) }
+      let!(:staffer) { create(:user, role_name: :staffer, mission: mission) }
+
+      it "should have all coordinators in to" do
+        expect(mail.to).to contain_exactly(coordinator1.email, coordinator2.email)
+      end
+
+      it "should have all coordinators in reply-to" do
+        expect(mail.reply_to).to contain_exactly(coordinator1.email, coordinator2.email)
```
smoyth: I would put both these assertions in one `it` block for efficiency since the fixtures are destroyed and rebuilt for each `it` block.

spec/mailers/notifier_spec.rb
```diff
@@ -39,4 +39,24 @@
...
+      end
+    end
+  end
```
smoyth: Can we also please add a check that this email gets delivered to the feature spec in the other issue. I can show you how.

app/mailers/notifier.rb
```diff
@@ -4,29 +4,36 @@
...
 
-  def reply_to(mission)
+  def coordinator_emails(mission)
     return [] if mission.nil?
-    User.with_roles(mission, :coordinator).pluck(:email).uniq[0, 10] # Max of 10 reply to, should be rare.
+    User.with_roles(mission, :coordinator).pluck(:email).uniq[0, 10] # Max of 10, should be rare.
+  end
+
+  def admin_emails
+    User.where(admin: true).pluck(:email).uniq[0, 10] # Max of 10, should be rare.
```
smoyth: Hmm... let's not limit admin emails. Seems potentially dangerous and confusing. I'm on the fence about coordinator ones, the likelihood of there being many coordinators seems higher, and they're going into the reply-to header of a commonly sent email...

spec/mailers/notifier_spec.rb
```diff
@@ -47,15 +47,13 @@
...
 
-      it "should have all coordinators in to" do
-        expect(mail.to).to contain_exactly(coordinator1.email, coordinator2.email)
-      end
-
-      it "should have all coordinators in reply-to" do
-        expect(mail.reply_to).to contain_exactly(coordinator1.email, coordinator2.email)
+      it "should have all coordinators/admins in 'to' and 'reply-to'" do
+        expect(mail.to).to contain_exactly(coordinator1.email, coordinator2.email, staffer2.email)
+        expect(mail.reply_to).to contain_exactly(coordinator1.email, coordinator2.email, staffer2.email)
```
smoyth: Nice use of decoy here.

PR #600, opened 2019-05-08
9700: Don't allow filtering by the current question

cooperka:
Strange this fix was so easy, I repro'd it on v9.10 so I figured it would be something more than this.

Also includes a small cleanup of all React components to make sure the vars saved into `state` are actually changed by `setState`; otherwise keep them in `props` for clarity.

### DIFF
--- a/app/javascript/components/SkipRuleFormField.js
+++ b/app/javascript/components/SkipRuleFormField.js
@@ -37,7 +37,6 @@ class SkipRuleFormFieldRoot extends React.Component {
       namePrefix,
       remove,
       id,
-      laterItems,
       skipIf,
       conditions,
       refableQings,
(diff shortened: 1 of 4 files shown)


app/javascript/components/SkipRuleFormField.js
```diff
@@ -117,10 +114,8 @@ class SkipRuleFormFieldRoot extends React.Component {
   }
 
   render() {
-    const { namePrefix, ruleId } = this.props;
-    const {
-      id, destItemIdOrEnd, laterItems, skipIf, destination, destItemId,
-    } = this.state;
+    const { id, laterItems, namePrefix, ruleId } = this.props;
+    const { destItemIdOrEnd, skipIf, destination, destItemId } = this.state;
```
smoyth: I like these cleanups. Thanks!

PR #605, opened 2019-05-14
9391 filter data sources

cooperka:
Added:
- Provides the `all_forms` prop for the Form filter
- Provides endpoints for users/groups for the Submitter filter
- Adds a `natural_sort` utility for sorting data

Changed:
- Save `{ name, id }` instead of `id` for selected options. Required because the map of id to name is no longer easily available (it's in select2's cache instead of our own). Nicely simplifies things.

Risks:
- Related select2s (including on Form/Submitter filter, Broadcasts route, maybe others)

9736 "Parse **current selection** for Form filter" will come next.


app/controllers/responses_controller.rb
```diff
@@ -112,33 +115,34 @@ def destroy
...
+    users = User.with_roles(current_mission, %w[coordinator staffer reviewer]).by_name
+    render_possible_users(users)
   end
 
   private
 
+  def render_possible_users(possible_users)
+    @possible_users = apply_search_if_given(User, possible_users)
+      .natural_sort_by_key
+      .paginate(page: params[:page], per_page: 20)
```
cooperka: `natural_sort` is needed before `paginate`, but that wrecks the nice lazy-evaluation for ActiveRecord Relations. @smoyth do you think this will significantly impact performance? Any simple way to load-test this?

app/controllers/responses_controller.rb
```diff
@@ -112,33 +115,34 @@ def destroy
...
-      when "reviewers" then User.with_roles(current_mission, %w[coordinator staffer reviewer]).by_name
-      end
-    @possible_users = apply_search_if_given(User, @possible_users)
-    @possible_users = @possible_users.paginate(page: params[:page], per_page: 20)
-
-    render(json: {possible_users: ActiveModel::ArraySerializer.new(@possible_users),
-                  more: @possible_users.next_page.present?}, select2: true)
+  def possible_reviewers
+    users = User.with_roles(current_mission, %w[coordinator staffer reviewer]).by_name
+    render_possible_users(users)
```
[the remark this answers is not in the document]
smoyth: Oh I think `by_name` can move into the private method since it's common across all, right?

app/controllers/responses_controller.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require "will_paginate/array"
```
smoyth: I don't think we should need this.

app/controllers/user_groups_controller.rb
```diff
@@ -1,6 +1,9 @@
+# frozen_string_literal: true
+
+require "will_paginate/array"
```
smoyth: I don't think we should need this.

app/controllers/user_groups_controller.rb
```diff
@@ -1,6 +1,9 @@
...
 class UserGroupsController < ApplicationController
   before_action :load_user_groups
-  before_action :find_user_group, only: [:add_users, :remove_users]
+  before_action :find_user_group, only: %i[add_users remove_users]
```
smoyth: Thanks for cleanups.

app/controllers/user_groups_controller.rb
```diff
@@ -66,12 +69,28 @@ def remove_users
...
+    end
   end
 
   private
 
+  def possible_groups_select2
+    @user_groups = @user_groups.name_matching(params[:search]) if params[:search].present?
+    @user_groups = @user_groups
+      .natural_sort_by_key
+      .paginate(page: params[:page], per_page: 20)
```
smoyth: Style-wise here I think it's better to do like:

```
x = y.foo
  .bar
```

because it's one less line, meaning you're less likely to run afoul of the method line limit.

app/controllers/user_groups_controller.rb
```diff
@@ -66,12 +69,28 @@ def remove_users
...
 
   def possible_groups
-    @user_groups = @user_groups.name_matching(params[:q])
-    render json: @user_groups
+    if params[:select2].present?
+      possible_groups_select2
+    else
+      @user_groups = @user_groups.name_matching(params[:q])
+      render(json: @user_groups)
+    end
```
smoyth: Let's talk about the split here...

app/views/searches/_form.html.erb
```diff
@@ -2,43 +2,10 @@
...
-    allGroups: Settings.filters_beta.present? ? [
-      {
-        id: "A",
-        name: "Group A"
-      },
-      {
-        id: "B",
-        name: "Group B"
-      },
-    ] : [],
```
smoyth: Yay!

config/initializers/enumerable_natural_sort.rb
```diff
@@ -0,0 +1,32 @@
...
+      x_value = x[key]
+      x_value = x_value.to_s unless x_value.respond_to?(:to_sort_atoms)
+
+      y_value = y[key]
+      y_value = y_value.to_s unless y_value.respond_to?(:to_sort_atoms)
+
+      x_value.to_sort_atoms <=> y_value.to_sort_atoms
+    end
+  end
+end
```
smoyth: I like the natural sort thing but I think I need to be convinced that adding all this is worth it. We are doing all kinds of sorting all over the app and it's generally just the naive kind. If there were some kind of data that had a lot of numbers at the beginning I could see it being worth it. But right now we're just sorting one of the filter lists with it, right? And we can't really use it for paginated database results. So yeah let's discuss...

spec/javascript/components/search/utils.js
```diff
@@ -61,35 +52,23 @@ export const getFiltersStore = () => new FiltersModel({
...
     [submitterType.GROUP]: [],
   },
 
-  selectedSubmitterIdsForType: {
-    [submitterType.USER]: ['B'],
-    [submitterType.GROUP]: ['B'],
+  selectedSubmittersForType: {
+    [submitterType.USER]: [{ id: 'B', name: 'User B' }],
+    [submitterType.GROUP]: [{ id: 'B', name: 'Group B' }],
   },
```
smoyth: What are these for?

config/initializers/enumerable_natural_sort.rb
cooperka: Code source again for posterity: https://makandracards.com/makandra/9185-ruby-natural-sort-strings-with-umlauts-and-other-funny-characters

We can leave this stuff in commit history but I'll remove from the codebase.

app/controllers/responses_controller.rb
cooperka: It was needed for pagination after sorting in-place, but no longer since we're removing that logic 👍

app/controllers/responses_controller.rb
```diff
@@ -112,33 +115,34 @@ def destroy
...
-                  more: @possible_users.next_page.present?}, select2: true)
+  def possible_reviewers
+    users = User.with_roles(current_mission, %w[coordinator staffer reviewer]).by_name
+    render_possible_users(users)
   end
 
   private
 
+  def render_possible_users(possible_users)
+    @possible_users = apply_search_if_given(User, possible_users)
```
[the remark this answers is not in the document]
cooperka: Hmm it was like this before but I don't see any current usages -- will remove.

cooperka:
CI failure details:

```console
# Failing due to invalid Maps API key?
rspec ./spec/features/results/location_picker_spec.rb:25

# All pass locally, will diagnose soon
rspec ./spec/features/results/skip_logic_spec.rb:37
rspec ./spec/features/results/form_rendering_and_submission_spec.rb:125
rspec ./spec/features/results/form_rendering_and_submission_spec.rb:266
rspec ./spec/features/results/form_rendering_and_submission_spec.rb:244
rspec ./spec/features/results/file_upload_spec.rb:20
```

PR #606, opened 2019-05-15
9722 fix assets:precompile in prod

cooperka:
To allow deploy. See commit messages for details.

Changes to `url_builder` are from this revert point: https://github.com/thecartercenter/nemo/blob/af33ed863b2479471b5d9517b408f7e28c135227%5E/app/assets/javascripts/legacy/url_builder.js.

Deployed successfully on staging.


cooperka:
Merging to make v9.12 release easier

PR #608, opened 2019-05-21
9687 q as ref support cond model

smoyth:


db/schema.rb
```diff
@@ -593,7 +593,7 @@
   add_foreign_key "broadcasts", "missions", name: "broadcasts_mission_id_fkey", on_update: :restrict, on_delete: :restrict
   add_foreign_key "choices", "answers", name: "choices_answer_id_fkey", on_update: :restrict, on_delete: :restrict
   add_foreign_key "choices", "options", name: "choices_option_id_fkey", on_update: :restrict, on_delete: :restrict
-  add_foreign_key "conditions", "form_items", column: "ref_qing_id", name: "conditions_ref_qing_id_fkey", on_update: :restrict, on_delete: :restrict
+  add_foreign_key "conditions", "form_items", column: "left_qing_id"
```
cooperka: Why were the extra params removed here?

app/models/condition.rb
```diff
@@ -147,7 +146,6 @@ def clean_times
         self.value = nil
       end
     end
-    return true
```
cooperka: For "Remove extraneous return trues" can you explain why they're extraneous now?

spec/factories/conditions.rb
```diff
@@ -39,8 +39,8 @@
   factory :condition do
     op "eq"
     left_qing { build(:questioning) }
-    value { left_qing.has_options? ? nil : "1" }
-    option_node { left_qing.has_options? ? left_qing.option_set.c[0] : nil }
+    value { left_qing.present? && !left_qing.has_options? ? "1" : nil }
+    option_node { left_qing&.has_options? ? left_qing.option_set.c[0] : nil }
```
cooperka: In the two lines above there's `left_qing.has_options` and `left_qing&.has_options`, why the difference?

db/schema.rb
smoyth: Good question! I'll look into it.

app/models/condition.rb
smoyth: https://blog.bigbinary.com/2016/02/13/rails-5-does-not-halt-callback-chain-when-false-is-returned.html

spec/factories/conditions.rb
smoyth: `&.` is safe navigation, meaning if the object is `nil`, the method won't be called.

It's not needed on line 42 because there is an explicit check. `&.` is not used on that line because it's a negative condition. If you look at the truth table it doesn't work out right if you use `&.`.

db/schema.rb
smoyth: Looks like Rails changed the way these are generated at some point. Doesn't make a difference practically.

PR #610, opened 2019-05-22
9693 q as ref support odk

smoyth:
Updates the ODK XML rendering engine to support questioning references as the right side of a conditional expression.


smoyth:
> can I support this with manual testing or do you think that's better done in bulk after the rest of the changes?

Thanks for asking! I think we'll be fine with testing at the end. The edge cases of converting our model into XML are best covered by the model specs. If you think something is missing there, let me know.

As for whether the XML works on ODK, I have done some manual testing as part of this work and am feeling confident.

PR #613, opened 2019-06-03
9745 searcher parsing

cooperka:
Note: First commit does a lot of rubocop. [Better diff](https://github.com/thecartercenter/nemo/compare/d2bfda4...9745_searcher_parsing?w=1).


app/controllers/concerns/searchable.rb
```diff
@@ -4,26 +4,38 @@
 module Searchable
   extend ActiveSupport::Concern
 
-  # If params[:search] is present, runs a search of `klass`,
+  # Returns a Searcher. See apply_search for details.
+  def get_searcher(relation)
```
smoyth: Ruby styleguide says don't start methods with `get`. Perhaps `build_searcher`?

app/models/search/lex_token.rb
```diff
@@ -1,33 +1,47 @@
...
+    "#{@kind}(#{@content})"
+  end
+
+  def to_sql
+    @sql
+  end
+
+  def to_s_indented(level = 0)
+    ("  " * level) + to_s
+  end
```
smoyth: Yay cleanups!

app/searchers/responses_searcher.rb
```diff
@@ -91,6 +91,22 @@ def apply
...
 
+    save_filter_data(search)
+
     relation.where(sql)
   end
+
+  # Save specific data that can be used for search filters.
+  def save_filter_data(search)
+    search.expressions.each do |expression|
+      if expression.qualifier.assoc.include?(:forms)
```
smoyth: Here I think you can just check if the name is `"form"`. The exact_form one is only used by reports and is kind of a hack.

app/searchers/responses_searcher.rb
```diff
@@ -91,6 +91,22 @@ def apply
...
+        form_names = expression.values_list
+        form_ids = form_names.map { |name| Form.find_by(name: name).try(:id) }
+          .reject(&:blank?)
+        self.form_ids += form_ids
```
smoyth: How about `form_ids.concat(Form.where(name: form_names).pluck(&:id)`

app/searchers/responses_searcher.rb
```diff
@@ -91,6 +91,22 @@ def apply
...
+      else
+        self.advanced_text += " #{expression.qualifier_text}:(#{expression.values})"
```
smoyth: It's generally more efficient and idiomatic with strings to use `<<` (append) instead of `+=`.

Can we also not wrap values with parens if not necessary? Is it easy to check?

app/searchers/searcher.rb
```diff
@@ -4,11 +4,17 @@
...
 
+  # Parsed search values
+  attr_accessor :form_ids, :advanced_text
+
   def initialize(relation:, query:, scope: nil)
     self.relation = relation
     self.query = query
     self.scope = scope
+    self.form_ids = []
+    self.advanced_text = ""
```
smoyth: Cool

app/searchers/noop_searcher.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Class that acts as a pass-through Searcher.
+# It simply returns the original relation when applied.
+class NoopSearcher < Searcher
+  def apply
+    relation
+  end
+end
```
smoyth: Very nice OOD approach here!

app/searchers/responses_searcher.rb
```diff
@@ -91,6 +91,22 @@ def apply
...
+      end
+    end
+  end
```
smoyth: This is definitely moving in the right direction. We need to talk about an important edge case though—OR expressions. We should only be plucking stuff from qualifiers if they are part of top-level ANDed expressions. (I realized this is handwavy and we should talk more about it to make it less handwavy!)

smoyth: I think we also need some model specs testing this extraction stuff. Should be pretty easy to write.

app/searchers/responses_searcher.rb
cooperka: I'll fix the `<<` but plan to wait on the parens until https://redmine.sassafras.coop/issues/9393. The advanced_text var here is more of a proof-of-concept until then.

app/searchers/responses_searcher.rb
```diff
@@ -99,8 +99,31 @@ def apply
...
+      leaves = expression.values_tokens.expand
+      leaves.each do |lex_tok|
+        # if this is a value token descendant
+        if lex_tok.parent.is?(:value)
+          # if the previous token was also a value token, it's an implicit AND
+          if previous&.parent&.is?(:value)
+            token_ops.push(:and)
+          end
+
+          token_values.push(lex_tok.content)
```
smoyth: `<<`

app/searchers/responses_searcher.rb
```diff
@@ -99,8 +99,31 @@ def apply
...
   def save_filter_data(search)
     search.expressions.each do |expression|
-      if expression.qualifier.name == "form"
-        form_names = expression.values_list
+      op_kind = expression.op.kind
+      token_values = []
+      token_ops = []
+      previous = nil
+
+      leaves = expression.values_tokens.expand
```
smoyth: Do this inside instead of out here.

app/searchers/responses_searcher.rb
```diff
@@ -165,7 +165,10 @@ def maybe_filter_by_expression(expression, op_kind, token_values, is_filterable)
   def filter_by_expression(expression, op_kind, token_values)
     if expression.qualifier.name == "form" && equality_op?(op_kind)
       form_names = token_values
-      self.form_ids = form_ids.concat(Form.where(name: form_names).pluck(:id))
+      matched_form_ids = Form.where(name: form_names).pluck(:id)
+      return false if matched_form_ids.empty?
+
+      self.form_ids = form_ids.concat(matched_form_ids)
```
smoyth: `form_ids.concat(matched_form_ids)` is sufficient without the assignment since `concat` mutates the object.

PR #614, opened 2019-06-04
9746 DRY up filters

cooperka:
Mostly-straightforward refactoring. I'm happy with the cleanup. It might be easier to review commit-by-commit because they're each fairly standalone.

Recommend viewing with `?w=1` (ignore whitespace).

### DIFF
--- a/app/javascript/components/Filters/FormFilter/__snapshots__/component.test.js.snap
+++ b/app/javascript/components/Filters/FormFilter/__snapshots__/component.test.js.snap
@@ -19,41 +19,37 @@ exports[`renders as expected 1`] = `
     ]
   }
   id="form-filter"
-  overlay={
-    <FilterPopover
-      id="form-filter"
-      onSubmit={[MockFunction]}
-    >
-      <Select2
-        data={
-          Array [
-            Object {
-              "id": "1",
-              "text": "One",
-            },
-            Object {
-              "id": "2",
-              "text": "Two",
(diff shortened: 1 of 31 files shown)


app/javascript/components/Filters/FormFilter/__snapshots__/component.test.js.snap
```diff
@@ -1,78 +1,56 @@
...
-  "results": Array [
-    Object {
-      "type": "return",
-      "value": undefined,
-    },
-  ],
-}
+exports[`popover handles callbacks 1`] = `1`;
+
+exports[`popover handles callbacks 2`] = `
```
cooperka: These two snapshots were split up because now we get the actual `event` which has a MASSIVE stringified value. So it's logging the number of calls instead of the whole `calls` array now.

app/javascript/components/Filters/QuestionFilter/component.js
```diff
@@ -22,47 +20,29 @@ class QuestionFilter extends React.Component {
...
-        <div className="btn-apply-container condition-margin">
-          <Button
-            className="btn-apply"
-            onClick={onSubmit}
-          >
-            {I18n.t('common.apply')}
-          </Button>
-        </div>
-      </Popover>
+      <ConditionSetFormField />
```
smoyth: Look at all that nice deleted code!

PR #619, opened 2019-06-19
9782 print q codes

cooperka:
[NEMO Sample Form [before codes].pdf](https://github.com/thecartercenter/nemo/files/3306880/NEMO.Sample.Form.original.pdf)

[NEMO Sample Form [after codes].pdf](https://github.com/thecartercenter/nemo/files/3306879/NEMO.Sample.Form.5.pdf)

### DIFF
--- a/app/assets/stylesheets/local/forms/_printable.scss
+++ b/app/assets/stylesheets/local/forms/_printable.scss
@@ -76,6 +76,10 @@
       }
     }
 
+    div.qing-code {
+      text-align: right;
+    }
+
     div.condition {
       font-weight: bold;
     }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/local/forms/_printable.scss
```diff
@@ -10,6 +10,10 @@
     h1.title {
       display: block;
       margin-bottom: 3px;
+
+      i {
+        margin-#{$right}: 5px;
```
smoyth: Nice observance of convention here. Arabic speakers thank you :D

PR #621, opened 2019-06-25
9691 constraints field

smoyth:


app/serializers/conditional_logic_form/skip_rule_serializer.rb
```diff
@@ -1,32 +1,5 @@
...
-#  index_skip_rules_on_dest_item_id    (dest_item_id)
-#  index_skip_rules_on_source_item_id  (source_item_id)
-#
-# Foreign Keys
-#
-#  fk_rails_...  (dest_item_id => form_items.id)
-#  fk_rails_...  (source_item_id => form_items.id)
-#
-# rubocop:enable Metrics/LineLength
-
```
smoyth: I removed these because they got added when this file was in the root serializers directory but it's not there anymore so they're not getting maintained anymore.

app/javascript/components/conditions/SkipLogicFormField/component.js
```diff
@@ -48,10 +68,38 @@ class SkipLogicFormField extends React.Component {
...
+              onClick={this.handleAddClick}
+              tabIndex="0"
+            >
+            {/* eslint-enable */}
+              <i className="fa fa-plus" />
+              {' '}
+              {I18n.t('form_item.add_rule')}
+            </a>
+          </div>
+        </div>
```
smoyth: Here I combined SkipRuleSet and SkipLogic because there didn't seem to be a need for both and the extra layer of nesting was increasing complexity.

cooperka:
Note if you add a skip rule for "all cases", there's no button to remove it:

Screen Shot 2019-06-26 at 21 38 51

cooperka:
Tested and verified on ODK Collect as well. Merging for deploy!

PR #622, opened 2019-07-10
9392 + 9393: other filter parsing (reviewed, user, group, question)

cooperka:
### Filter summary

Already parsed:
- form

Newly parsed in this PR:
- reviewed
- user
- group
- question
    - Handles single, multiple, and cascading
    - Handles option set translations
- advanced search text
    - Handles cleaning up unnecessary parens e.g. `source:(x)` => `source:x`
    - Handles all types of operators

### What to expect during review

Screen Shot 2019-07-10 at 08 08 40

- All filter buttons should work to filter data
- All filter buttons should remember what they're originally set to and show it as a "hint" on the button itself
- Text manually typed in the advanced search box should be automatically converted into filter button UI **only** when possible to do so without losing information
- Design pass is not yet done; the question filter seems to be missing an "add" link and has some CSS issues that will be fixed later

### Known issues [all resolved]

- [x] If you filter by a select_one question, after refreshing the page you need to open the question filter again to trigger `onMount` to trigger data fetching. If you simply "search" again right away, the select_one question filter will be lost.
- [x] If you filter by a different form, the question filter gets reset. This is intentional, but the "hint" should not be cleared.

Note: integration testing plan is captured in https://redmine.sassafras.coop/issues/9775


app/searchers/responses_searcher.rb
```diff
@@ -162,16 +169,81 @@ def maybe_filter_by_expression(expression, op_kind, token_values, is_filterable)
...
+  def find_option_node_id(option_set_id, value)
+    partial_match = OptionNode.joins(:option).includes(:option)
+      .where(option_set_id: option_set_id)
+
+    # TODO: Can this be more efficient?
+    # Doesn't work with jsonb: `partial_match.where("options.name_translations ILIKE ?", value)`
+    full_match = partial_match.where("options.canonical_name ILIKE ?", value) +
+      partial_match.select do |node|
+        node.option.name_translations.values.map(&:downcase).include?(value.downcase)
+      end
```
cooperka: ^ Yuck, please let me know if you have a better way to do this. I couldn't find any way to match jsonb values, even though it seems like it should be trivial to cast them to a string and use `ILIKE`.

app/views/searches/_form.html.erb
```diff
@@ -3,10 +3,11 @@
...
     allForms: @all_forms || [],
-    selectedFormIds: @form_ids || [],
-    isReviewed: nil,
-    selectedUsers: Settings.filters_beta.present? ? [] : [],
-    selectedGroups: Settings.filters_beta.present? ? [] : [],
+    selectedFormIds: @selected_form_ids || [],
+    selectedQings: @selected_qings || [],
+    isReviewed: @is_reviewed,
+    selectedUsers: @selected_users || [],
+    selectedGroups: @selected_groups || [],
```
smoyth: I wonder if a cleaner way to do this would be to have a serializer for the ResponsesSearcher class.  ResponsesSearcher would then be responsible for all of this. allForms could be added, advancedSearch will be there eventually anyway, controllerName could just be hardcoded to responses. Seems to make sense, no?

spec/searchers/responses_searcher_spec.rb
```diff
@@ -19,12 +19,27 @@
...
+    let!(:r2) { create(:response, form: form) }
+
+    it "should work" do
+      expect(search(%(reviewed:1))).to contain_exactly(r1)
+
+      expect(searcher(%(reviewed:1))).to have_filter_data(is_reviewed: true, advanced_text: "")
+      expect(searcher(%(reviewed:yes))).to have_filter_data(is_reviewed: true, advanced_text: "")
+      expect(searcher(%(reviewed:"NO"))).to have_filter_data(is_reviewed: false, advanced_text: "")
+      expect(searcher(%(reviewed:("0")))).to have_filter_data(is_reviewed: false, advanced_text: "")
+      expect(searcher(%(reviewed:(1 0)))).to have_filter_data(is_reviewed: nil, advanced_text: "reviewed:(1 0)")
```
smoyth: Gorgeous test. Just what I want to see.

spec/searchers/responses_searcher_spec.rb
```diff
@@ -56,6 +71,23 @@
...
+    let!(:u1) { create(:user, name: "u1") }
+    let!(:u2) { create(:user, name: "u2 name") }
+    let!(:u3) { create(:user, name: "u3") }
+    let!(:r1) { create(:response, user: u1) }
+    let!(:r2) { create(:response, user: u2) }
+
+    it "should work" do
+      expect(search(%(submitter:#{u1.name}))).to contain_exactly(r1)
+
+      expect(searcher(%(submitter:#{u1.name}))).to have_filter_data(submitters: items_by_name_and_id(u1), advanced_text: "")
```
smoyth: I would consider getting rid of `items_by_name_and_id`. It's not shortening things much over `{id: u1.id, name: "u2 name"}` and it adds a layer of abstraction that makes the test harder to read. The reviewer test is so nice because it doesn't have this.

spec/searchers/responses_searcher_spec.rb
```diff
@@ -73,6 +105,54 @@
...
+      expect(searcher(%(text:apple))).to have_filter_data(
+        qings: [],
+        advanced_text: "text:(apple)"
+      )
+      expect(searcher(%({blue}:apple))).to have_filter_data(
+        qings: [{id: qing1.id, value: "apple"}],
+        advanced_text: ""
+      )
+      expect(searcher(%({GREEN}:apple {blue}:apple text:apple))).to have_filter_data(
+        qings: [{id: qing2.id, value: "apple"}, {id: qing1.id, value: "apple"}],
```
smoyth: Case in point: this is way easier to read!

spec/searchers/responses_searcher_spec.rb
```diff
@@ -73,6 +105,54 @@
...
+        advanced_text: "text:(apple)"
+      )
+      expect(searcher(%({red}:#{node3.option.canonical_name}))).to have_filter_data(
+        qings: [{id: qing3.id, option_node_id: node3.id}],
+        advanced_text: ""
+      )
```
smoyth: Wow, very nice test overall.

spec/searchers/responses_searcher_spec.rb
```diff
@@ -73,6 +105,54 @@
...
+    end
+  end
+
+  describe "question qualifier" do
+    let!(:opt_set) { create(:option_set, option_names: :multilevel) }
+    let!(:q1) { create(:question, qtype_name: "long_text", code: "blue", add_to_form: form) }
+    let!(:q2) { create(:question, qtype_name: "long_text", code: "Green", add_to_form: form) }
+    let!(:q3) do
+      create(:question, qtype_name: "select_one", code: "red", option_set: opt_set, add_to_form: form)
+    end
```
smoyth: The more conventional way to set this up is like this:

```
let(:form) { create(:form, question_types: %w[long_text long_text multilevel_select_one]) }
let(:qing1) { form.c[0] }
let(:node3) { form.c[2].option_set.c[0] } 
``` 

Note the second line may not be necessary if you can just use form.c[0] inline.

spec/searchers/responses_searcher_spec.rb
```diff
@@ -174,4 +254,12 @@ def search(query)
...
     ResponsesSearcher.new(relation: Response, query: query, scope: {mission: get_mission})
   end
+
+  def items_by_name_and_id(*items)
+    items.pluck(:id, :name).map { |id, name| {id: id, name: name} }
+  end
+
+  def qing_for_question(question)
+    FilterDataController.filter_unique(Questioning.where(question: question)).first
+  end
```
smoyth: I don't think this should be necessary unless I'm missing something.

app/searchers/responses_searcher.rb
smoyth: I think you can use `JSONB_EACH_TEXT`. See `app/models/results/answer_search_vector_updater.rb`. 

I assume you're using ILIKE just to do a case insensitive comparison and are not using the wildcards (`%`) at all? If so, using `LOWER()` with regular comparison will be faster I think. `ILIKE` should only be used if you need wildcards in my experience, which is not true here.

Let me know if you want to pair on this one. If you can get something working fairly quickly go for it, else let's pair.

cooperka: Good call on lower vs. ilike, I'll update. I tried `JSONB_EACH_TEXT` previously but it fails with `set-returning functions are not allowed in WHERE`:

```ruby
    full_match = partial_match.where("options.canonical_name ILIKE ?", value)
      .or(partial_match.where("JSONB_EACH_TEXT(options.name_translations) ILIKE ?", value))
```

```console
ActiveRecord::StatementInvalid (PG::FeatureNotSupported: ERROR:  set-returning functions are not allowed in WHERE
LINE 1: ... $1 AND ((options.canonical_name ILIKE 'Cat') OR (JSONB_EACH...
                                                             ^
: SELECT "option_nodes"."id" FROM "option_nodes" INNER JOIN "options" ON "options"."id" = "option_nodes"."option_id" WHERE "option_nodes"."option_set_id" = $1 AND ((options.canonical_name ILIKE 'Cat') OR (JSONB_EACH_TEXT(options.name_translations) ILIKE 'Cat'))):
```

cooperka: We can pair this week if you think there's another way to get it working (maybe different syntax outside of `where`)

app/javascript/components/Filters/model.js
```diff
@@ -51,11 +51,19 @@ class FiltersModel {
   }
 
   constructor(initialState = {}) {
+    const { selectedQings } = initialState;
+    if (!isEmpty(selectedQings)) {
+      this.conditionSetStore.resetConditionsFromQings(selectedQings);
+
+      // eslint-disable-next-line no-param-reassign
+      delete initialState.selectedQings;
```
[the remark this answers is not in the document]
cooperka: If any qings should be selected, it creates new conditions for them and then deletes the original property (since it's no longer needed, it was used to create the new conditions). I've added code comments.

app/searchers/responses_searcher.rb
```diff
@@ -162,16 +169,81 @@ def maybe_filter_by_expression(expression, op_kind, token_values, is_filterable)
...
+    when "submitter"
+      return filter_by_names(token_values, User, current_ids: submitters, include_name: true)
+    when "group"
+      return filter_by_names(token_values, UserGroup, current_ids: groups, include_name: true)
+    else
+      return false
     end
+  end
+
+  # Given a list of values, find all instances of this class that match,
```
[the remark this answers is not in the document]
cooperka: Yes it is, starting on line 112

spec/searchers/responses_searcher_spec.rb
cooperka: You're right -- I added it to use the deterministic quality of `filter_unique`, but during these tests there's only one questioning per question so the determinism isn't needed. I'll remove.

app/serializers/searchers/responses_searcher_serializer.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    def selected_groups
+      object.groups || []
+    end
+
+    def advanced_search_text
+      (Settings.filters_beta.present? ? object.advanced_text : object.query) || ""
+    end
+  end
+end
```
smoyth: I think you should not be using a `Searchers` namespace as it's inconsistent with other object types. e.g. there are no Models or Controllers or Decorators namespaces.

Namespaces should be about the domain, not the object type. So Odk, Results, Forms, etc. You could put this one in Results if you want to namespace it.

app/serializers/searchers/searcher_serializer.rb
```diff
@@ -0,0 +1,14 @@
...
+    attributes :advanced_search_text
+
+    format_keys :lower_camel
+
+    def advanced_search_text
+      # We can't assume advanced_text was actually parsed, so return the original query.
+      object.query || ""
+    end
+  end
+end
```
smoyth: Same here re: namespaces.

spec/searchers/responses_searcher_spec.rb
```diff
@@ -157,10 +250,34 @@
...
+      expect(searcher("foo")).to have_filter_data(advanced_text: "foo")
+      expect(searcher("text:foo")).to have_filter_data(advanced_text: "foo")
+      expect(searcher("text=it's")).to have_filter_data(advanced_text: "it's")
+      expect(searcher("bar bar")).to have_filter_data(advanced_text: "bar bar")
+      expect(searcher("reviewed:1 foo")).to have_filter_data(advanced_text: "foo")
+      expect(searcher("foo reviewed:1 123.4 source:(x)")).to have_filter_data(advanced_text: "foo 123.4 source:x")
+      expect(searcher("source:(\"x y\" z)")).to have_filter_data(advanced_text: "source:(\"x y\" z)")
+      expect(searcher("submit-date >= 2013-10-15 submit-date < 2013-10-20")).to have_filter_data(advanced_text: "submit-date>=2013-10-15 submit-date<2013-10-20")
+    end
+  end
```
smoyth: These specs are super dreamy

app/searchers/responses_searcher.rb
```diff
@@ -249,17 +249,16 @@ def qing_value(matched_question, token_values)
...
-    full_match = partial_match.select do |node|
-        node.option.name_translations.values.map(&:downcase).include?(value)
-      end
+    possibilities = OptionNode.joins(:option).where(option_set_id: option_set_id)
+    results = OptionNode.none
+
+    # Allow matching any translation in the mission's locales.
+    configatron.preferred_locales.each do |locale|
+      results = results.or(possibilities.where("LOWER(options.name_translations ->> ?) = ?", locale, value))
+    end
```
smoyth: Nicely handled here.

PR #627, opened 2019-07-22
9898 tally report query optimization

smoyth:
The reported bug was caused by duplicate join table entries. The number of duplicates would double each time the report were edited. The report in question must have been edited 14 times because it had 16,384 duplicates, and this was causing the generated query to be enormous.

The fix was related to permitted params. A migration was also needed to fix the existing duplicates and add a unique index to catch this if it happens again in future.

Assigning assuming CI passes. Please kick back to me if not.


app/controllers/reports_controller.rb
```diff
@@ -143,7 +143,7 @@ def report_params
       :unreviewed, :filter, :question_labels, :show_question_labels, :question_order, :text_responses,
       :percent_type, :unique_rows, :calculations, :option_set, :mission_id, :mission,
       :disagg_question_id, :group_by_tag,
-      option_set_choices_attributes: %i[option_set_id],
-      calculations_attributes: %i[id type report_report_id attrib1_name question1_id arg1 attrib1 question1 rank _destroy])
+      option_set_choices_attributes: %i[id option_set_id _destroy],
+      calculations_attributes: %i[id type attrib1_name question1_id arg1 attrib1 question1 rank _destroy])
```
cooperka: I don't quite understand why this fixes the bug?

db/migrate/20190722202039_remove_report_opt_set_choice_duplicates.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class RemoveReportOptSetChoiceDuplicates < ActiveRecord::Migration[5.2]
+  def up
+    result = execute("SELECT (ARRAY_AGG(id))[1] AS id, option_set_id, report_report_id, COUNT(*) AS num
+      FROM report_option_set_choices GROUP BY option_set_id, report_report_id HAVING COUNT(*) > 1")
```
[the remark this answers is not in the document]
smoyth: Postgres apparently does 1-based array indexing.

app/controllers/reports_controller.rb
smoyth: Yeah it's a bit esoteric. With Rails nested params, if you don't provide an ID to match on when you're updating, it creates a new record. Since we weren't permitting the ID to come through, Rails was happily creating a new join table record on each update instead of updating the existing one.

In newer versions of Rails, it complains loudly if it rejects a param. But in earlier versions it fails silently (probably due to the fact that strong parameters was a major major breaking change when it first came out so they were introducing it slowly).

PR #628, opened 2019-07-24
9733 filters design pass

cooperka:
## Relevant functionality

- Searching for basic text
  - Keyboard "enter" should submit
- Choosing a filter
  - Should submit automatically on dismiss popover (whether you click Apply or not)
- Choosing a form, then choosing a question from that form
  - Question choices are limited based on the selected form
- Clicking the info button
  - Search help modal should appear
  - Keyboard "enter" should also activate it
- Typing in some filterable text e.g. `reviewed:no`
  - Text should get consumed by the filter UI and no longer appear in the search box
- Typing in some non-filterable text e.g. `source:foo`
  - Text should NOT get consumed by the filter UI, and remain in the search box

## Potentially confusing behavior

Probably not worth addressing these, unless there's a quick win solution.

- If you change the selected form, the question filter gets cleared automatically (because the questions to filter on depend on the form)
  - And nothing happens **visually** until after the search is submitted, so this might be confusing to users
- If you filter by a select-multiple question, submit the search, then open the question filter AGAIN, then close it without making any changes, the search still submits automatically
  - This is because the question filter loads the select-multiple "levels" after being mounted, which makes it seem "dirty" and it thinks something changed even though you didn't change anything

## Known bugs (WIP)

- [ ] If you filter by a form, then ALSO filter by a cascading question, then ALSO submit the search a third time WITHOUT opening the question popover first, the question value is `Unknown`
- [x] rspec failure `question_index_spec.rb:29` (clear button removal)

## Original UI (without any filter buttons)

1 - original

## New UI (after design pass)

Screen Shot 2019-07-24 at 11 30 42
Screen Shot 2019-07-24 at 11 31 24
Screen Shot 2019-07-24 at 11 31 59


jpdimond:
Really wonderful work Kevin! I appreciate how well things work, and your attention to detail! I do like the hint where it is turns out :-)

image

A couple things for polish:

- Can we make the filter outlines and the search box outlines the same color?
- Can we make the width of the search box the same width as all of the filters? It's just off a little bit.
- Can we increase the margin-left of the i.hint to 15px? Right now, it would be easy to misclick because they are close together.

If time:
- Would be nice to have the options "Yes, No, Either" to have colors associated with them for easy scanning. A light green for yes, a light red for no, and keep either as gray.

Wahoo!!!! We are gonna have FILTERS.
🎆 🎉 🎉

cooperka:
Thanks Jill! I'll do all the above, with the note that the filter buttons will change width if (1) a filter is selected and it shows a hint or (2) once we add the date filter, or any other filters, or have more/less filters on a different page. But I think it does make sense to align the searchbox to a width of 4 filters so it at least aligns with something sometimes.

app/javascript/components/Filters/QuestionFilter/component.js
```diff
@@ -20,8 +22,20 @@ class QuestionFilter extends React.Component {
   }
 
   renderPopover = () => {
+    const { filtersStore: { allForms, selectedFormIds } } = this.props;
+    const formNames = selectedFormIds.map((id) => getItemNameFromId(allForms, id));
+    const formConstraintText = isEmpty(formNames)
+      ? null
+      : `Showing questions from ${formNames.join(', ')} only.`;
```
smoyth: This needs to be i18n'd right?

app/views/searches/_form.html.erb
```diff
@@ -2,10 +2,12 @@
   <%= react_component("Filters", @searcher_serializer.as_json.merge({
     controllerName: Settings.filters_beta.present? ? controller_name.to_json : nil,
   })) %>
-  <div class="search-footer">
-    <%= search_examples %>
-    <a href="#"><%= t('search.help_link') %></a>
-  </div>
+  <% unless Settings.filters_beta.present? && controller.is_a?(ResponsesController) %>
```
smoyth: Is it time to get rid of the filters_beta setting check? This thing is ready for full release after this PR, right?

cooperka:
Hmm 🙃 I'm sure some other combination of colors would look better -- this is bootstrap's default "success" and "danger" values for green and red:

Screen Shot 2019-07-25 at 18 36 10

Anything else will require more complex CSS (I'd need hex values for default, hover, and pressed states, as well as focused border "glow" color). Guessing we should just leave these gray @jpdimond.

cooperka:
After feedback:

Screen Shot 2019-07-25 at 18 48 27

PR #630, opened 2019-07-29
9945 cascading select filter bug

cooperka:
Bug description:
> If you filter by a form, then ALSO filter by a cascading question, then ALSO submit the search a third time WITHOUT opening the question popover first, the question value is `Unknown` and the search fails.

This PR:
- Fixes the bug by filtering `qings` by any selected `form_id`s before determining the `refableQing` ID to pass to the client
- Does some minor code cleanup

This is the last remaining piece needed before releasing the initial version of filters.

### DIFF
--- a/app/javascript/components/conditions/ConditionSetFormField/ConditionFormField/model.js
+++ b/app/javascript/components/conditions/ConditionSetFormField/ConditionFormField/model.js
@@ -53,6 +53,31 @@ class ConditionModel {
   @observable
   remove;
 
+  /** Return either the current value or the value of the deepest defined level. */
+  @computed
+  get currTextValue() {
+    if (this.optionSetId) {
+      const lastIndex = this.levels.length - 1;
+
+      for (let i = lastIndex; i >= 0; i -= 1) {
+        const { selected, options } = this.levels[i];
+
+        if (selected != null) {
+          const { name = '' } = options.find(({ id }) => selected === id) || {};
+          return name;
+        }
+      }
+
+      return null;
+    } if (this.optionNodeId) {
+      // If optionSet hasn't been loaded (for example, if question filter popover
+      // hasn't yet been mounted), fall back to the string value passed by backend.
+      return this.optionNodeValue;
+    }
+
+    return this.value;
+  }
+
   constructor(initialState = {}) {
     Object.assign(this, initialState);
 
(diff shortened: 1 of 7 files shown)


app/javascript/components/conditions/ConditionSetFormField/ConditionFormField/model.js
```diff
@@ -53,6 +53,35 @@ class ConditionModel {
...
+        const { selected, options } = this.levels[i];
+
+        if (selected != null) {
+          const { name = '' } = options.find(({ id }) => selected === id) || {};
+          return name;
+        }
+      }
+
+      return null;
+    } if (this.optionNodeId) {
```
smoyth: Is this supposed to be `else if`?

app/searchers/responses_searcher.rb
```diff
@@ -131,6 +131,14 @@ def clean_up
     self.form_ids = form_ids.uniq
     self.submitters = submitters.uniq
     self.groups = groups.uniq
+    # Map possible_ids to a single id based on any active form filters.
+    self.qings = qings.map do |qing|
+      return qing unless qing[:possible_ids]
+      qings = Questioning.where(id: qing[:possible_ids])
+      qings = qings.where(form_id: form_ids) if form_ids.present?
+      qing[:id] = qings.filter_unique.pluck(:id).first
+      qing.except(:possible_ids)
```
smoyth: This block is a bit tough to read. A few things: 

1. You are shadowing `self.qings`—perhaps choose a different var name for inside the loop
2. The block feels like it should be its own method with a nice descriptive name. A comment on the block like this is often a hint that a method is called for.
3. What was once a `clean_up` method is really now a second parsing pass. I would call it that perhaps. Or maybe rename `clean_up` to `dedupe` or something and then call that and then the new method all from `save_filter_data`. Or whatever feels right to you. But the narrowing of the question options feels like more than just cleaning up.

app/searchers/responses_searcher.rb
```diff
@@ -232,12 +240,10 @@ def filter_by_questions(qualifier_text, token_values)
     question_code = qualifier_text[1..-2]
     matched_question = Question.where("LOWER(code) = ?", question_code.downcase).first
     return false if matched_question.blank?
-    matched_qings = Questioning.where(question: matched_question)
-    return false if matched_qings.blank?
-    # Get the deterministic Questioning ID from the Question
-    matched_qing_id = matched_qings.filter_unique.first.id
+    matched_qing_ids = Questioning.where(question: matched_question).pluck(:id)
```
smoyth: Why pluck the IDs here and then redo this query moments later? Could you just store the relation in the `qings` array instead?

smoyth: Also a big fat comment seems worthwhile here explaining that this is an intermediate result and will be further refined in a second parsing pass once the form filter has been parsed or wahtever.

app/javascript/components/conditions/ConditionSetFormField/ConditionFormField/model.js
cooperka: There's a [lint check](https://eslint.org/docs/rules/no-else-return) that `else` after a `return` doesn't make sense because the else is implicit. Apparently airbnb explicitly changed the default to also prevent `else if`. But I don't like that rule, I'll turn it off and fix this.

PR #631, opened 2019-07-30
9909 filter media types

cooperka:
Only allow filtering by reportable questions (e.g. don't show images, audio, etc. as refableQings for the question search filter).

~~@smoyth you mentioned something in the call today about using `ref`, but I didn't see how that worked here. I had already come up with this solution, so submitting for feedback. Let me know if you prefer something else.~~

### DIFF
--- a/app/controllers/filter_data_controller.rb
+++ b/app/controllers/filter_data_controller.rb
@@ -9,7 +9,7 @@ class FilterDataController < ApplicationController
   def qings
     authorize!(:index, Response)
     qings = Questioning.for_mission(current_mission).joins(:question)
-      .includes(:question).order("questions.code")
+      .includes(:question).where(question: Question.reportable).order("questions.code")
     qings = qings.where(form_id: params[:form_ids]) if params[:form_ids].present?
     qings = qings.filter_unique
     render(json: ActiveModel::ArraySerializer.new(qings,
--- a/app/controllers/filter_data_controller.rb
+++ b/app/controllers/filter_data_controller.rb
@@ -9,7 +9,7 @@ class FilterDataController < ApplicationController
   def qings
     authorize!(:index, Response)
     qings = Questioning.for_mission(current_mission).joins(:question)
-      .includes(:question).where(question: Question.reportable).order("questions.code")
+      .includes(:question).with_type_property(:refable).order("questions.code")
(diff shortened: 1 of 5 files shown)


app/controllers/filter_data_controller.rb
```diff
@@ -9,7 +9,7 @@ class FilterDataController < ApplicationController
   def qings
     authorize!(:index, Response)
     qings = Questioning.for_mission(current_mission).joins(:question)
-      .includes(:question).order("questions.code")
+      .includes(:question).where(question: Question.reportable).order("questions.code")
```
smoyth: Oh, this is not what I meant, but good find! 

Please see app/models/question_type.rb. 

We should refactor this old scope in fact. Can you add another property (`reportable`) to `AVAILABLE_PROPERTIES` and then add a new scope to Question called `with_property(property)`? Then wherever `reportable` is used, use that instead. 

And then for your purposes here, please use the `refable` property because that also excludes location questions, which is appropriate here. `refable` is also used with the condition system, which you're reusing in the filters.

smoyth:
@cooperka Looks like we have a CI failure here. Any idea what that's about?

cooperka:
Seems like a recent flapper (for an old spec), I've seen it on at least one other PR and put it in the flappers sheet:

```
1309  1) response form reviewer notes admin user can submit notes
1310     Failure/Error: @full_dotted_rank ||= full_rank.join(".")
1311
1312     NoMethodError:
1313       undefined method `join' for nil:NilClass
1314     # ./app/models/form_item.rb:186:in `full_dotted_rank'
1315     # ./app/controllers/filter_data_controller.rb:15:in `qings'
1337
1338 rspec ./spec/features/results/reviewer_notes_spec.rb:30
```

Any thoughts? I can look into it unless it strikes you as something you've dealt with.

smoyth:
No I haven't seen it and yes could you look into it please.

app/controllers/filter_data_controller.rb
cooperka: I think I understand, and I took a whack at it, but the method delegation doesn't seem to work. Let's pair tomorrow or Fri.

```
undefined method `with_property' for #<Questioning::ActiveRecord_Relation>
```

cooperka:
Fixed via `with_type_property` as discussed, thanks for the help 🏅

PR #632, opened 2019-07-31
9957 flapping reviewer notes spec

cooperka:
Notes:
- Flapping started somewhere within https://github.com/thecartercenter/nemo/pull/617 (can't bisect precisely, not all commits compile, but somewhere in that range)
- The spec deterministically flaps **every other time it's run**, no matter what the seed value is

I'd prefer to do a more specific fix but I can't tell where `path` is actually getting set and it's too generic to search for. If you'd like to dive in feel free, but this fixes the flap and seems like a reasonably safe change.

### DIFF
--- a/app/controllers/filter_data_controller.rb
+++ b/app/controllers/filter_data_controller.rb
@@ -12,6 +12,13 @@ class FilterDataController < ApplicationController
       .includes(:question).with_type_property(:refable).order("questions.code")
     qings = qings.where(form_id: params[:form_ids]) if params[:form_ids].present?
     qings = qings.filter_unique
+
+    # Temporary guard clause to investigate a strange flapping spec.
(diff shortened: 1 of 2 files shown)


smoyth:
Hmm this is puzzling. `path` is provided by `has_ancestry`.

`full_rank` can only be `nil` if `path` is `[]`, which should never be the case.

I would want to figure out what is going on there. Doing `|| []` seems dangerous to me.

If you don't have capacity we can add it to the backlog and I'll take a look later.

cooperka:
`has_ancestry` is helpful, thanks. I'll unassign myself for now and take a look again on Friday if time. Since it fails every other time, this should rarely block PR merges since the CI is run twice on every PR 😉

cooperka:
Update: I probably won't have time to fix before leaving, and considering the priority of this, I will leave it for now (remaining unassigned).

PR #633, opened 2019-08-05
9958: Stop preserving search query between missions

cooperka:
Uses [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) to build, modify, and stringify the url params without the `search` query.

Note: Normally I'd use the more conventional npm `query-string` lib for this but it seems like the legacy JS code wouldn't deal well with that. This should be a robust solution, as far as I know. Related SO question [here](https://stackoverflow.com/a/901144/763231) as well.

### DIFF
--- a/app/assets/javascripts/legacy/views/mission_change_dropdown.js
+++ b/app/assets/javascripts/legacy/views/mission_change_dropdown.js
@@ -15,8 +15,11 @@
         var path = ELMO.app.params.mode == 'basic' ? '' : window.location.pathname;
 
         // Preserve query string and add missionchange param to it.
-        var qs = window.location.search;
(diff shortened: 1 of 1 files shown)


cooperka:
Done. I haven't specifically tested on a real IE browser, think it's worth downloading a VM and setting that up or do you have any other way to test?

smoyth:
Browserstack baby

cooperka:
Hah. All sorts of things are broken in IE, including all React and anything that uses the legacy JS `ELMO` global. Navigation is also significantly impaired because several important link handlers don't trigger.

I think IE may be a lost cause (and clearly no tears have been shed by users, at least not loudly), I propose reverting the last commit and officially denoting IE as unsupported. Thoughts?

PR #634, opened 2019-08-06
9775 filter integration specs

cooperka:
They've been a long time coming but they're finally here 🎊 

These integration tests confirm that:
- The filters are active on the responses page (but not the questions page)
- The filters parse search strings and display them using the UI
- The filters display hints
- The select2 fields can be cleared
- The filters submit automatically on dismiss popover

They're intended to be high-level and not comprehensive (e.g. there's no specific integration spec for choosing a question with the dropdowns since that's covered in more specific tests), but let me know if you think I should cover more cases or if I missed anything.

### DIFF
--- a/spec/features/search/response_search_spec.rb
+++ b/spec/features/search/response_search_spec.rb
@@ -62,5 +62,17 @@ feature "response search", js: true do
     expect(page).to have_field("search", with: "")
     expect(page).to have_css("#submitter-filter.active-filter", text: "Submitter (2 filters)")
     expect(page).to have_css("#reviewed-filter.active-filter", text: "Reviewed (No)")
+
+    new_search_for(%(form:"Form 1"))
+    click_on("Question")
+    expect(page).to have_content("Showing questions from Form 1 only.")
+    click_on("Form (Form 1)")
+    # Clear the form filter.
+    find(".select2-selection__clear").click
+    # The hint should still be there until the search is submitted.
+    expect(page).to(have_css(".active-filter"))
+    # Click off to dismiss the popover and automatically submit the search.
+    find("h1").click
+    expect(page).to_not(have_css(".active-filter"))
   end
 end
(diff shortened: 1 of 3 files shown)


spec/support/contexts/search.rb
```diff
@@ -6,6 +6,12 @@ def search_for(query)
     click_button("Search")
   end
 
+  def new_search_for(query)
+    # Clear the query params.
+    visit(current_path)
+    search_for(query)
```
cooperka: Perhaps `visit(current_path)` should be the default instead of adding this new method? Since `fill_in("search")` no longer clears the entire search (only the advanced search text).

workroommedia:
Thanks Kevin!

Administrators, Coordinators and Staffers are really excited about this one. 

Dottie

spec/features/search/response_search_spec.rb
```diff
@@ -41,4 +42,37 @@
...
+  scenario "search filters" do
+    login(user)
+    visit "/en/m/#{mission.compact_name}/responses"
+
+    search_for(%(foo))
+    expect(page).to have_field("search", with: "foo")
+
+    new_search_for(%(form:"Form 1"))
+    expect(page).to have_field("search", with: "")
+    expect(page).to have_css("#form-filter.active-filter", text: "Form (Form 1)")
```
smoyth: Should we be covering somewhere in here that the search is actually being applied? i.e. that the correct rows are showing and not showing? Maybe not for all the searches (since we are covering correct search results at the model level) but for one of them?

spec/support/contexts/search.rb
smoyth: I think it's fine like this for now. I wouldn't want to add another unnecessary page load, and implicitly checking can be fraught due to timing issues.

spec/features/search/response_search_spec.rb
cooperka: Done.

PR #638, opened 2019-08-09
9960 filter matching

cooperka:
This simplified a few things which was nice.

In short, this changes the search syntax used by the filters slightly e.g.

```diff
-form:("One"|"Two") {Q1}:"Cat" reviewed:1 submitter:"User B" foo
+form-id:("id1"|"id2") {Q1}:"Cat" reviewed:1 submitter-id:"id4" foo
```

Manually typing `form:"One"` no longer gets consumed by the filters UI, and remains in the advanced search box. `form-id:"id"`, however, does get consumed (`id` must be a UUID).

---

Note this doesn't implement exact-matching for "select" question option nodes. I originally implemented it, and had it working for the filters UI itself, but then realized that breaks the actual search results because of the more complex {QuestionCode} syntax... I reverted the commit and left questions as-is.

@smoyth if you think it's worth exact-matching option nodes, you could un-revert that commit and add new searcher logic that would be used **only** if the question referenced by {QuestionCode} has an `option_set_id`. I didn't know if this would be worth the time so leaving this up to you.


app/javascript/components/conditions/ConditionSetFormField/ConditionFormField/model.js
```diff
@@ -54,29 +54,29 @@ class ConditionModel {
...
 
   /**
-   * Return a string representation of the current value.
-   * Both user-facing and sent to the backend when performing a search.
+   * Return the current value to search for on the backend:
+   * either a basic string, or a selected option ID.
    */
   @computed
-  get currTextValue() {
+  get searchValue() {
```
cooperka: Renamed for clarity. It's no longer user-facing anywhere.

PR #639, opened 2019-08-13
9766 - Only email active users on SMS token change.

sdroadie:
Only send emails to active users or admins when an SMS token is changed for a mission.


smoyth:
Terrific!

PR #640, opened 2019-08-13
Updates to fix issues moving to Webpacker v4.

sdroadie:
Moving to Webpacker 4 forced a move to Babel 7, which broke a number of JavaScript package usages.


smoyth:
Wow nice work!

PR #641, opened 2019-08-20
9686 constraints rejection message modal

sdroadie:
Figured out most of this! Generating a modal with React, persisting the rejection message, and toggling the link text for the modal based on the presence of a custom message.

Still not really clear how the message is used, and I haven't been able to figure out how to get the `preferred_locales` to check against translated messages.

Added tests to `constraint_form_spec.rb`, but the new tests are failing on `wait_modal_to_be_visible`. I suspect it has something to do with the timeout waiting for the react element to render visibly.


app/javascript/components/conditions/RejectionModal/component.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import Modal from 'react-bootstrap/Modal';
```
smoyth: Nice

app/javascript/components/conditions/RejectionModal/component.js
```diff
@@ -0,0 +1,55 @@
...
+class RejectionModal extends React.Component {
+  static propTypes = {
+    conditionSetStore: PropTypes.object,
+    title: PropTypes.string,
+    show: PropTypes.bool,
+    handleClose: PropTypes.func,
+  };
+
+  static defaultProps = {
+    title: 'Add Rejection Message',
```
smoyth: Should be translated. But is this necessary?

app/javascript/components/conditions/RejectionModal/component.js
```diff
@@ -0,0 +1,55 @@
...
+import Modal from 'react-bootstrap/Modal';
+import Button from 'react-bootstrap/Button';
+import PropTypes from 'prop-types';
+import { inject, observer } from 'mobx-react';
+
+@inject('conditionSetStore')
+@observer
+class RejectionModal extends React.Component {
+  static propTypes = {
+    conditionSetStore: PropTypes.object,
```
smoyth: I'm not sure you'll need the store in this component. The only piece of data you need to share out is whether or not there are any translations set, right? So you could have that be a piece of state on this component that gets read by the link component.

app/javascript/components/conditions/RejectionModal/component.js
```diff
@@ -0,0 +1,55 @@
...
+
+  static defaultProps = {
+    title: 'Add Rejection Message',
+    show: false,
+  };
+
+  handleChange = (e) => {
+    const { conditionSetStore: { setRejectionMessage } } = this.props;
+    setRejectionMessage(e.target.value);
+  }
```
smoyth: So perhaps this just sets local state. I don't think we need the store here.

app/javascript/components/conditions/RejectionModal/component.js
```diff
@@ -0,0 +1,55 @@
...
+        <Modal.Header closeButton>
+          <Modal.Title>{title}</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <div className="elmo-form">
+            <div className="form-field">
+              <label className="main" htmlFor="rejection-message">Message:</label>
+              <div className="control">
+                <div className="widget">
+                  <input id="rejection-message" className="form-control" type="text" value={rejectionMsgTranslations.en} onChange={this.handleChange} />
```
smoyth: As long as the DOM elements for this lie inside the Questioning form, this input will get included in the submission to the server. Which is why we don't need the hidden inputs elsewhere if we have the right name attrib on this.

This should become a loop over `rejectionMsgTranslations`, with one text field per translation. The name attrib should look like `${namePrefix}[rejection_msg_translations][${locale}]`.

Then on the server it will arrive as `{..., rejection_msg_translations: {en: 'blah', fr: 'foo'}, ...}`, which will be permitted with the change I suggested above.

app/javascript/components/conditions/ConditionSetFormField/model.js
```diff
@@ -157,6 +160,11 @@ class ConditionSetModel {
     }));
   }
 
+  @action
+  setRejectionMessage = (message) => {
+    this.rejectionMsgTranslations.en = message;
+  }
+
```
smoyth: Don't think we need to use the store here actually. See below.

app/javascript/components/conditions/RejectionMessageLink/component.js
```diff
@@ -0,0 +1,53 @@
...
+        <a href="#" onClick={this.handleShow} tabIndex="0">
+          <i className="fa fa-pencil edit-rejection-message" />
+          {' '}
+          { I18n.t(messageId) }
+        </a>
+
+        <RejectionModal
+          show={show}
+          title={I18n.t(messageId)}
+          handleClose={this.handleClose}
```
smoyth: If not using the store, will need to pass down the existing translations as props. See below.

spec/features/forms/form_item/constraints_form_spec.rb
```diff
@@ -58,6 +61,10 @@
             select_operator("< less than")
             fill_in_value("25")
           end
+          click_link("Add Rejection Message")
+          wait_modal_to_be_visible
```
smoyth: Let's wait on diagnosing this spec stuff until we are feeling good about the stuff above.

app/serializers/conditional_logic_form/form_item_serializer.rb
```diff
@@ -3,7 +3,7 @@
 module ConditionalLogicForm
   # Serializes Questioning or QingGroup for creating display and skip rule form fields in browser
   class FormItemSerializer < ApplicationSerializer
-    attributes :id, :display_if, :code, :rank, :full_dotted_rank, :form_id, :type
+    attributes :id, :display_if, :code, :rank, :full_dotted_rank, :form_id, :type, :group_name_translations
```
smoyth: @sdroadie any ideas why this is here? I can't remember a reason for this to be here. Perhaps we can remove?

sdroadie: Ah, yeah, I forgot I added this. Totally not needed.

PR #643, opened 2019-08-29
10094 Option set modal height

ziedAb:


app/assets/stylesheets/local/_questions.scss
```diff
@@ -91,7 +91,8 @@ form.questioning_form {
   z-index: 500;
 
   .modal-body {
-    min-height: 400px;
+    height: 400px;
+    overflow: auto;
```
smoyth: @ziedAb Can you explain what happened here? Was the modal getting too tall? Is 400px a good size for the fixed height? Could we maybe go taller? Can you provide a screenshot with an option set with around 10 options on it?

ziedAb: 400px modal height

400px isn't the size of the whole modal, it is just the body

smoyth: Thank you.

PR #644, opened 2019-08-29
10097 fix right to left css bugs

ziedAb:


app/assets/stylesheets/global/_layout.scss
```diff
@@ -2,6 +2,10 @@ header {
   background-color: $header-background-color;
 }
 
+body {
+  text-align: #{$left};
```
smoyth: Can this just be `text-align: $left` as below? I think the `#{...}` syntax is just for keys?

app/assets/stylesheets/global/_layout.scss
```diff
@@ -57,7 +61,7 @@ header {
     display: inline-block;
     width: 190px;
     height: 31px;
-    margin-right: 10px;
+    margin-#{$right}: 10px;
```
smoyth: You know what would be cool (for a separate issue, not in this PR) would be a git hook on this repo that flags any non-RTL-friendly CSS rules on commit. So you'd match for anything like `-(right|left):` or `: (right|left);`. Otherwise stuff will keep creeping in... You'd need to test it on the existing stylesheets and fix any existing issues so we start out with a clean set... Thoughts?

smoyth: You could even do a blog post about this. I bet others could use it...

app/assets/stylesheets/global/_navbar.scss
```diff
@@ -18,6 +18,11 @@ nav.navbar-light {
     }
   }
 
+  // Overriding bootstrap navbar prefs causing RTL bug
```
smoyth: Can you add a bit more detail here and perhaps a URL reference, if you found this online? Are you certain this doesn't effect the navbar behavior for LTR users?

ziedAb: I didn't found this online, i have read about the property and did the appropriate fix, both RTL and LTR versions are working good.

smoyth: Cool.

PR #647, opened 2019-09-20
9400 add date search filter

sdroadie:
Adds a date search filter to the responses view, using [react-dates](https://github.com/airbnb/react-dates).

Updating dates in the filter will update the URL, and entering dates via the text search input will be parsed and appropriately handled in the filter.

If multiple dates are entered in the search input, the widest range of dates will be used.


sdroadie:
I think this is good to go. The remaining failing specs seem like flapping specs.

jpdimond:
UX review! Overall the functionality is awesome. Just some stylistic tweaks:

image
- Please make the height and font size of the input fields the same height as in the question drop down: 
image

- Please make the select color in the component the NEMO blue and light blue as opposed to the default green
image
Dark color: $accent-color
Light color: $light-blue

- Please align the apply button so it is on the same line as the fields

smoyth:
@cooperka There are some failing snapshots, a bunch of failed propType warnings, and a bunch of deprecation warnings when the JS tests are run. Could you clean these up please and take a quick look at the React code here to ensure it meets your standards?

cooperka:
@smoyth I pushed several tweaks, I recommend reading commit messages and reviewing once more before merge. Happy to discuss any questions.

Question for you: why the addition of stylesheet_pack_tag + extract_css (https://github.com/thecartercenter/nemo/commit/0e05c3b6423961c14a9be952bd4ea5fd12e65e8a), is it because there's CSS bundled with the new datepicker component that needs to be pulled into the main bundle?

app/javascript/components/Filters/model.js
```diff
@@ -86,14 +96,6 @@ class FiltersModel {
...
     Object.assign(this, initialState);
 
-    // Convert dates to moment if given
-    if (startDate) {
-      this.startDate = moment(startDate);
-    }
-    if (endDate) {
-      this.endDate = moment(endDate);
-    }
-
```
smoyth: Oops! Thanks for fixing!

PR #649, opened 2019-09-24
10151 show custom constraint rejection message

sdroadie:
Flapping specs seem to be failing, but the rest of this works for rendering to ODK.

### DIFF
--- a/app/controllers/concerns/condition_formable.rb
+++ b/app/controllers/concerns/condition_formable.rb
@@ -6,4 +6,10 @@ module ConditionFormable
   def condition_params
     [:id, :left_qing_id, :right_qing_id, :right_side_type, :op, :value, :_destroy, option_node_ids: []]
   end
+
+  def human_readable_conditions(codes: true, nums: true)
+    decorated_conditions = ConditionDecorator.decorate_collection(condition_group.members)
+    concatenator = condition_group.true_if == "all_met" ? I18n.t("common.AND") : I18n.t("common.OR")
+    decorated_conditions.map { |c| c.human_readable(codes: codes, nums: nums) }.join(" #{concatenator} ")
+  end
 end
--- a/app/controllers/concerns/condition_formable.rb
+++ b/app/controllers/concerns/condition_formable.rb
@@ -6,10 +6,4 @@ module ConditionFormable
   def condition_params
     [:id, :left_qing_id, :right_qing_id, :right_side_type, :op, :value, :_destroy, option_node_ids: []]
   end
-
-  def human_readable_conditions(codes: true, nums: true)
-    decorated_conditions = ConditionDecorator.decorate_collection(condition_group.members)
-    concatenator = condition_group.true_if == "all_met" ? I18n.t("common.AND") : I18n.t("common.OR")
-    decorated_conditions.map { |c| c.human_readable(codes: codes, nums: nums) }.join(" #{concatenator} ")
-  end
 end
--- a/app/decorators/odk/qing_decorator.rb
+++ b/app/decorators/odk/qing_decorator.rb
@@ -6,7 +6,7 @@ module Odk
     delegate_all
 
     def bind_tag(form, subq, xpath_prefix: "/data")
-      tag(:bind, nodeset: [xpath_prefix, subq.try(:odk_code)].compact.join("/"),
+      tag(:bind, nodeset: [xpath_prefix, subq&.odk_code].compact.join("/"),
                  type: binding_type_attrib(subq),
                  required: required? && visible? && subq.first_rank? ? required_value(form) : nil,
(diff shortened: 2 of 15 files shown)


spec/fixtures/odk/forms/conditional_logic.xml
```diff
@@ -151,7 +158,8 @@
       <bind nodeset="/data/*itemcode7*_2" type="select1" relevant="((/data/*itemcode3* &gt; 5) and (/data/*itemcode6* = '*optcode1*')) and (not((/data/*itemcode3* = 0)))"/>
       <bind nodeset="/data/*itemcode8*" type="select"/>
       <bind nodeset="/data/*itemcode9*" type="dateTime"/>
-      <bind nodeset="/data/*itemcode10*" type="int" constraint="(. &gt;= 10 and . &lt;= 100) and ((/data/*itemcode4* = 10) or (. = /data/*itemcode3*)) and ((. != 55))" jr:constraintMsg="Valid only if: Must be greater than or equal to 10 and less than or equal to 100; [DecimalQ4] is equal to 10 OR This question is equal to [IntegerQ3]; This question is not equal to 55"/>
+      <bind nodeset="/data/*itemcode10*" type="int" constraint="(. &gt;= 10 and . &lt;= 100) and ((/data/*itemcode4* = 10) or (. = /data/*itemcode3*)) and ((. != 55))" jr:constraintMsg="Must be greater than or equal to 10 and less than or equal to 100; Valid only if: [DecimalQ4] is equal to 10 OR This question is equal to [IntegerQ3]; This question is not equal to 55"/>
```
smoyth: This is not quite the right expectation. I would like to see:

`Must be greater than or equal to 10 and less than or equal to 100; Valid only if: [DecimalQ4] is equal to 10 OR This question is equal to [IntegerQ3]; Valid only if this question is not equal to 55`

smoyth: Note the extra 'valid only if` in the last clause. Without that, it doesn't make sense.

app/decorators/odk/qing_decorator.rb
```diff
@@ -68,13 +68,18 @@ def constraint
...
+          constraint.rejection_msg_translations[I18n.locale.to_s]
       end
-      str = msgs.compact.join("; ").presence
-      return nil if str.nil?
-      I18n.t("constraint.odk_message", conditions: str)
+      msgs = msgs.compact.join("; ").presence
+      custom_msgs = custom_msgs.compact.join("; ").presence
+      return nil if msgs.nil? && custom_msgs.nil?
+      [min_max_error_msg, msgs.nil? ? msgs : I18n.t("constraint.odk_message", conditions: msgs), custom_msgs]
+        .compact.join("; ")
```
smoyth: See below about changing the output slightly. I think this code can be simplified a good bit. You should be able to do it with just one `msgs` array I think.

spec/fixtures/odk/forms/conditional_logic.xml
```diff
@@ -106,6 +107,12 @@
           <text id="*itemcode10*:hint">
             <value>Question Hint 10</value>
           </text>
+          <text id="*itemcode11*:label">
+            <value>Integer Question Title 11</value>
+          </text>
+          <text id="*itemcode11*:hint">
+            <value>Question Hint 11</value>
+          </text>
```
smoyth: Nice job figuring out this fixture prep system.

spec/requests/odk/form_rendering_spec.rb
```diff
@@ -93,6 +93,14 @@
           {left_qing_id: form.c[9].id, op: "neq", value: 55}
         ]
       )
+
+      form.c[10].constraints.create!(
+        accept_if: "all_met",
+        conditions_attributes: [
+          {left_qing_id: form.c[10].id, op: "eq", value: 10}
+        ],
+        rejection_msg_translations: {"en": "Custom rejection message."}
```
smoyth: This looks great. Think you can use a symbol key on the `rejection_msg_translations`, which would be more conventional. Try it?

spec/fixtures/odk/forms/conditional_logic.xml
sdroadie: Ah, yeah, makes sense.

app/controllers/concerns/condition_formable.rb
```diff
@@ -6,4 +6,10 @@ module ConditionFormable
   def condition_params
     [:id, :left_qing_id, :right_qing_id, :right_side_type, :op, :value, :_destroy, option_node_ids: []]
   end
+
+  def human_readable_conditions(codes: true, nums: true)
```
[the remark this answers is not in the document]
sdroadie: Sure thing.

sdroadie:
@smoyth I think the `itext` stuff is mostly right.

PR #658, opened 2019-10-09
10225: Fix nil rejection_msg_translations issue

cooperka:
rejection_msg_translations already had a default, but was being set to `nil` due to empty strings for every locale.


cooperka:
Updated as discussed. The only usage I could find was within the React component, and `rejectionMsgTranslations` is already defaulted to `{}` in `ConditionSetModel`.

cooperka:
Confirmed by Tom, merging for hotfix.

PR #660, opened 2019-10-18
10199 odk version nums

cooperka:
> I was confused why it was passing at first but there's actually a separate spec that generates the XML for tests, and that was still using `code`. I think I found all the usages at this point but please be on the lookout for ones I may have missed.


app/models/form_version.rb
```diff
@@ -34,32 +37,35 @@ class FormVersion < ApplicationRecord
...
 
   private
 
-  # generates the unique random code
+  # Generates the unique version code using ODK date-revision convention
   def generate_code
-    # only need to do this if code not set
     return if code
+    # Today's date with a revision of 00
+    self.code = Date.now.strftime("%Y%m%d00")
```
cooperka: Should this be in any particular timezone? Only matters for the day (no hour)

db/migrate/20191018160023_change_form_version_code_to_odk_style.rb
```diff
@@ -0,0 +1,24 @@
...
+        revision += 1
+        break unless FormVersion.find_by(code: form_version.code)
+      end
+
+      form_version.save!
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
cooperka: It's not a big deal if someone wants to roll back; should this `raise` or simply noop?

smoyth:
Unfortunately this is not what I was after. `code` is still needed. The plan here was to add a *new* column called number for use with ODK. Let's discuss.

app/models/odk/response_parser.rb
```diff
@@ -178,7 +178,7 @@ def lookup_and_check_form(params)
 
       # if form version is outdated, error
       if form.current_version.code != params[:version] &&
-          form.current_version.sequence.to_s != params[:version]
+          form.current_version.number != params[:version]
```
cooperka: Can you explain the context behind [issue 8831](https://redmine.sassafras.coop/issues/8831) if you remember? Maybe this second check is obsolete now.

app/models/form_version.rb
```diff
@@ -15,51 +17,68 @@
...
 #
 #  form_versions_form_id_fkey  (form_id => forms.id) ON DELETE => restrict ON UPDATE => restrict
 #
 # rubocop:enable Metrics/LineLength
 
-# models a version number for a Form object. allows forms to have multiple uniquely identifiable versions.
-# provides 3 letter code for use with sms encoded forms.
+# Models a version number for a Form object. Allows forms to have multiple uniquely identifiable versions.
+# Provides unique 3-letter code for use with sms encoded forms,
+# plus unique 10-character number for use with odk xml forms.
```
smoyth: Ahh, much better!

app/models/form_version.rb
```diff
@@ -15,51 +17,68 @@
...
-  # inits a new FormVersion with same form_id
-  # increments sequence
-  # sets self.is_current = false
+  # Inits a new FormVersion with the same form_id
+  # and assigns it a new code + number
   def upgrade!
     upgraded = self.class.new(form_id: form_id, is_current: true)
     self.is_current = false
-    save
+    save!
```
smoyth: I think this method could maybe just be two lines:

```
update!(is_current: false)
self.class.create!(form_id: form_id, is_current: true)
```

Could you try that?

app/models/form_version.rb
```diff
@@ -15,51 +17,68 @@
...
-    true
+    while self.class.find_by(code: (self.code = Random.letters(CODE_LENGTH))); end
+  end
+
+  def ensure_unique_number
+    revision = number[-2, 2].to_i
+    # Increment the revision number until no match
+    while self.class.find_by(number: number)
+      revision += 1
+      raise RevisionTooHighError if revision >= 100
```
smoyth: Perfect

app/models/odk/response_parser.rb
smoyth: This will all be changing soon anyway. But yes you can just remove the second check.

db/migrate/20191021183012_add_odk_style_form_version_number.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class AddOdkStyleFormVersionNumber < ActiveRecord::Migration[5.2]
+  def change
+    add_column :form_versions, :number, :string
+    add_index :form_versions, [:number], unique: true
+
+    reversible do |dir|
```
smoyth: Wow never heard of this before!

db/migrate/20191021183012_add_odk_style_form_version_number.rb
```diff
@@ -0,0 +1,29 @@
...
+    # Set version number using ODK date-revision convention
+    loop do
+      raise RevisionTooHighError if revision >= 100
+      form_version.number = date + revision.to_s.rjust(2, "0")
+      revision += 1
+      break unless FormVersion.find_by(number: form_version.number)
+    end
+
+    form_version.save!
+  end
```
smoyth: Very nice

db/schema.rb
```diff
@@ -196,10 +196,11 @@
     t.datetime "created_at", null: false
     t.uuid "form_id", null: false
     t.boolean "is_current", default: true, null: false
-    t.integer "sequence", default: 1, null: false
+    t.string "number"
```
smoyth: This should be null false and have a reasonable limit. You'll have to add the null false constraint after you populate.

spec/models/form_version_spec.rb
```diff
@@ -34,6 +35,11 @@
     assert_match(/[a-z]{#{FormVersion::CODE_LENGTH}}/, fv.code)
   end
 
+  it "form version number generated on initialize" do
+    fv = FormVersion.new(form: form)
+    assert_match(/[0-9]{10}/, fv.number)
```
smoyth: Please convert all `assert_match` to `expect(...).to match(...)`

db/migrate/20191021183012_add_odk_style_form_version_number.rb
cooperka: Seems to be newish, saves a bit of time instead of manually undoing the add_column/index!

PR #663, opened 2019-11-01
10267 fix group condition issue [hotfix]

cooperka:
Repro steps:
- Go to a form with multiple groups, including one at the root e.g. https://staging.getelmo.org/en/admin/forms/5e38b7c5-1f04-48f6-95e2-3d0946214abb/edit
- Click to edit the first group
- Close modal, then click to edit second group
- Note the Name field is correct but the Display Logic field is wrong

This fix is a bit hacky, but it's the simplest solution I could think of given the unusual combination of technology we have in this scenario:
- Rails creating a React component
- React component creating a Mobx store
- Same React component being re-created when the modal goes away and comes back, thus needing to re-create the store as well

Reason it wasn't working:
- When the React component was re-created, it asked for an instance of the store, and was given the old instance of the store from the store cache
- The old code worked just fine on pages that render one instance of each component, but modals are an exception because they go away and come back with different data
- The component in the modal successfully re-rendered, but showed the old store data

Reason the fix works:
- Now there's no store cache; the store is always created from scratch, so the data is correct

Reason for adding the new nonce:
- When the `DisplayLogicFormField/injector` component is re-created from scratch by Rails, it somehow tries to resume some aspect of its old state, causing Mobx to warn loudly about store replacement **if the store has changed** (the reason for adding the store cache in the first place)
- This resuming of state when completely re-created doesn't seem to be typical React behavior, and is possibly due to some complexity added by the interaction between React/Rails/Mobx
- In order to prevent Mobx from complaining on re-mount, we need to force the component to reset by giving it a **unique key** every time it's mounted by Rails
    - React uses `key` to reconcile the virtual dom; it's an easy and conventional way to tell React "this component is different than before"

Performance implications:
- Hardly any, because the component should have been doing this in the first place

Specs coming momentarily.

### DIFF
--- a/app/javascript/components/Filters/injector.js
+++ b/app/javascript/components/Filters/injector.js
@@ -7,6 +7,8 @@ import { provideFiltersStore } from './utils';
 import { submitterType } from './SubmitterFilter/utils';
 import Filters from './component';
 
+let forcedResetNonce = 0;
+
 /**
  * Provides any needed stores to its children.
  */
@@ -38,8 +40,9 @@ function Injector(props) {
   });
   const { conditionSetStore } = filtersStore;
 
+  // eslint-disable-next-line no-return-assign
   return (
-    <Provider filtersStore={filtersStore} conditionSetStore={conditionSetStore}>
+    <Provider key={forcedResetNonce += 1} filtersStore={filtersStore} conditionSetStore={conditionSetStore}>
       <Filters {...props} />
     </Provider>
   );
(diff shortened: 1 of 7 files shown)


cooperka:
Only failing spec is "joe batt's arm" because now there are 2 locations returned by geocoding api instead of 1 🤕 releasing this as hotfix!

smoyth:
Thanks Kevin, for an elegant-as-possible solution. It pains me to see the knots we have to twist ourselves into, in 2019, to do a modest bit of interactivity on a server rendered app. I would love if you could think about an architecture direction that alleviates some of this pain.

PR #670, opened 2019-11-03
669: allow download of attached media with token

### LINKED ISSUE #669: Allow download of attached files/media using API Token
When we get a results list from an API call we do not have the link to the attached files. Also, the attached files/media of the responses can only be downloaded today if the user is logged in.

Our proposal is to allow download of media files using the API token (use the route media/objects#show).

viniciuscb:
Allows to download attached media when user is not logged in but the API token is provided in the headers of the request.

Resolves #669

### DIFF
--- a/app/controllers/media/objects_controller.rb
+++ b/app/controllers/media/objects_controller.rb
@@ -6,6 +6,7 @@ module Media
     include Storage
 
     before_action :set_media_object, only: %i[show destroy]
+    before_action :authenticate_token, only: %i[show]
     skip_authorization_check
 
     def show
@@ -69,6 +70,11 @@ module Media
       when "images" then Media::Image
       else raise "A valid media type must be specified"
       end
+    end
+
+    def authenticate_token
+      user = authenticate_with_http_token { |token, _options| User.find_by(api_key: token) }
+      @current_user = user if user
     end
   end
 end


app/controllers/media/objects_controller.rb
```diff
@@ -70,5 +71,10 @@ def media_class(type)
       else raise "A valid media type must be specified"
       end
     end
+
+    def authenticate_token
+      user = authenticate_with_http_token { |token, _options| User.find_by(api_key: token) }
+      @current_user = user if user
```
smoyth: I see what you are going for here, and I don't think it's a security issue because we are checking if the user has access to the media object inside the `show` method.

I'm not a fan of this authentication logic living in this controller though. It's a bit of duplication and not good separation of concerns.

Could you please:

1. Move this method to `app/controllers/concerns/application_controller/authentication.rb`
2. Rename to `authenticate_with_token`.
3. Add a guard clause to not do anything in this method if `@current_user` is already set.
4. Use `ActiveSupport::SecurityUtils.secure_compare(token, TOKEN)` (see https://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Token.html)
5. Refactor `app/controllers/api/v1/base_controller.rb` to use this new method (note this will involve replacing instances of `@api_user` with `current_user`.

A benefit of this is that the existing API request specs will cover the token auth code. Currently what you added has no coverage.

Thanks!

viniciuscb: Hi @smoyth as I've said in another PR I'm far from home now and will can see this only by the end of january. If there is no hurry, I would kindly ask you to wait a little bit more, otherwise I'd suggest that you make these modifications. Sorry.

cooperka:
Closing in favor of NEMO's new (existing) OData API as mentioned in the linked issue. Thank you again for this PR, and let us know if the OData API doesn't address your need!

Issue #673, opened 2019-11-03
Option sets child nodes are not loading (not authorized


viniciuscb:
This is happening when filling a form in web interface, there were two linked option sets (one main and one child, dependent) and this second option set is not loading for an enumerator user (who could fill all the other fields and submit form).

cooperka:
Thank you for the bug fix @viniciuscb, we don't have bandwidth right now to review and test but we hope to eventually get to it!

PR #672, opened 2019-11-03
671: Show download link for media fields in API requests

viniciuscb:
This suggestion adds the download link to API response requests on media fields, in the value field, which otherwise would be appearing as null.

Resolves #671

PR #674, opened 2019-11-03
673: added skip authorization on option sets child nodes

viniciuscb:
This commit solves the bug listed in #673

### DIFF
--- a/app/controllers/option_sets_controller.rb
+++ b/app/controllers/option_sets_controller.rb
@@ -9,6 +9,7 @@ class OptionSetsController < ApplicationController
   # authorization via cancan
   load_and_authorize_resource
   skip_authorization_check only: :child_nodes
+  skip_authorize_resource only: :child_nodes
 
   decorates_assigned :option_sets
 


app/controllers/option_sets_controller.rb
```diff
@@ -9,6 +9,7 @@ class OptionSetsController < ApplicationController
   # authorization via cancan
   load_and_authorize_resource
   skip_authorization_check only: :child_nodes
+  skip_authorize_resource only: :child_nodes
```
smoyth: Why are both of these needed?

viniciuscb: This is not related to our change in linked option sets, and is a bugfix in your code. The description of the bug is in the issue.

@smoyth I think I've not tested putting out skip_authorization_check.

cooperka:
(Note: PRs #674, #672, #670, #668 were automatically closed due to removal of old `develop` branch; I've updated their bases and re-opened them)

smoyth:
Closing for now due to age. Can be opened again later.

PR #677, opened 2019-11-12
10242 integrity warnings

smoyth:
image


app/models/form.rb
```diff
@@ -159,8 +159,8 @@ def version
     current_version.try(:code) || ""
   end
 
-  def any_questions?
-    questionings.any?
+  def data?
+    responses.any?
```
smoyth: The `any_questions?` method was not being used.

app/models/forms/integrity_warnings/question_warner.rb
```diff
@@ -0,0 +1,31 @@
...
+        # for standard copies.
+        %i[published data standard_copy]
+      end
+
+      def form_list
+        form_count = object.forms.size
+        more_suffix =
+          if form_count > MAX_FORMS
+            str = I18n.t("integrity_warnings.more_suffix", count: form_count - 3)
+            "(+#{str})"
```
cooperka: Note "3 more" might become "more 3" in other languages, so `+` at beginning might be better placed in the translation itself. Probably nbd though.

app/models/forms/integrity_warnings/warner.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+
+      def warnings(type)
+        # Expects the subclass method with name `type` to return an array of
+        # symbols or 1-element hashes of form [symbol => symbol].
+        # The first symbol is the boolean method to call on the object
+        # to check if the warning should be shown.
+        # The second symbol (in the array case) is a method to be called locally to provide extra
+        # info for I18n.
+        send(type).map do |params|
```
cooperka: Love the abstraction, but can this be made more explicit? E.g. named args `should_warn` and `i18n_info` instead of arrays of implicitly different symbols.

PR #680, opened 2019-11-20
10244 oldest accepted version setting

cooperka:
### Changes

- Adds new dropdown to form edit under "more settings"
  - Only if form has versions
- Adds new flag to FormVersion denoting "is oldest accepted version"
  - First version defaults to true, all future versions default to false
- Rejects ODK submissions that have outdated `number` e.g. `1999010100`
- Rejects SMS submissions that have a `code` from a FormVersion that has an outdated `number`
- Minor cleanup


app/models/form.rb
```diff
@@ -265,16 +275,31 @@ def current_version
     versions.current.first
   end
 
+  def oldest_accepted_version_id
+    return @oldest_accepted_version_id if defined?(@oldest_accepted_version_id)
+    versions.oldest_accepted.first&.id
```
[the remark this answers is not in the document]
smoyth: Maybe define a helper method `persisted_oldest_version_accepted` which returns `versions.oldest_accepted.first`. That might clean up a bunch of the below.

app/models/form.rb
```diff
@@ -265,16 +275,31 @@ def current_version
...
+
+  def oldest_accepted_version
+    versions.find(oldest_accepted_version_id) if oldest_accepted_version_id.present?
+  end
+
+  def update_oldest_accepted
+    if defined?(@oldest_accepted_version_id) &&
+        @oldest_accepted_version_id != versions.oldest_accepted.first.id
+      versions.oldest_accepted.first.update!(is_oldest_accepted: false)
+      versions.find(@oldest_accepted_version_id).update!(is_oldest_accepted: true)
```
smoyth: If you do as I suggest above, this could be:

```
  if defined?(@oldest_accepted_version_id)
    persisted_oldest_version_accepted&.update!(is_oldest_accepted: false)
    oldest_version_accepted.update!(is_oldest_accepted: true)
  end
```

Right? Rails won't do the update on the persisted one if it's already false.

app/models/form.rb
```diff
@@ -265,16 +275,31 @@ def current_version
...
     raise "standard forms should not be versioned" if standard?
 
-    if current_version
-      current_version.upgrade!
-    else
-      FormVersion.create(form_id: id, is_current: true)
-    end
+    has_current_version = current_version.present?
+    current_version.update!(is_current: false) if has_current_version
+    versions.create!(is_current: true, is_oldest_accepted: !has_current_version)
```
smoyth: Very smooth here!

app/models/form_version.rb
```diff
@@ -34,17 +35,20 @@ class FormVersion < ApplicationRecord
   after_initialize :generate_code, :generate_number
   before_create :ensure_unique_code, :ensure_unique_number
 
+  # Only one can be true per form.
+  validates :is_current, uniqueness: {allow_blank: true, scope: :form_id}
+  validates :is_oldest_accepted, uniqueness: {allow_blank: true, scope: :form_id}
```
smoyth: For future reference, Rails convention says you don't need `is_` in front of column names like this. Someday we can rename both.

app/models/form_version.rb
```diff
@@ -34,17 +35,20 @@ class FormVersion < ApplicationRecord
...
 
   CODE_LENGTH = 3
 
-  # Inits a new FormVersion with the same form_id
-  # and assigns it a new code + number
-  def upgrade!
-    update!(is_current: false)
-    self.class.create!(form_id: form_id, is_current: true)
+  # Version name for displaying to user
+  def name
```
smoyth: This should be in a decorator. Can you please add one real quick?

app/models/odk/response_parser.rb
```diff
@@ -178,10 +178,8 @@ def lookup_and_check_form(params)
       raise "xml submissions must be to versioned forms" if form.current_version.nil?
 
       # if form version is outdated, error
-      if form.current_version.code != params[:version] &&
-          form.current_version.number != params[:version]
-        raise FormVersionError, "Form version is outdated"
-      end
+      oldest_accepted_num = form.oldest_accepted_version.number.to_i
```
smoyth: `oldest_accepted_version` is feeling so verbose. Should we change to `minimum_version`? Then cols would be `minimum` and `current` (instead of `is_current`)? Feel free to skip if it would take too long.

app/models/sms/decoder/decoder.rb
```diff
@@ -125,7 +125,8 @@ def find_form
 
         # If form version outdated we must specify the form AND form_code in the error message
         # since they are different.
-        raise_decoding_error("form_version_outdated", form: v.form, form_code: code) unless v.is_current?
+        raise_decoding_error("form_version_outdated", form: v.form, form_code: code) unless
```
smoyth: Yikes, old one-letter var! Please rename `v` to `version` as a refactor.

app/models/sms/decoder/decoder.rb
```diff
@@ -125,7 +125,8 @@ def find_form
...
+          v.number >= v.form.oldest_accepted_version.number
```
smoyth: Please change to block form for multiline condition.

app/models/sms/decoder/decoder.rb
smoyth: There is also a bunch of LoD violations in here. Can you assign `form = version.form` further up once we know we have a form, and then maybe add a delegator in form to create `minimum_version_number` since we use that in lots of places?

app/views/forms/_oldest_accepted_version.html.erb
```diff
@@ -0,0 +1,5 @@
+<% if read_only %>
+  <%= form.object.oldest_accepted_version&.name %>
```
smoyth: Please use decorator here too.

app/views/forms/_oldest_accepted_version.html.erb
```diff
@@ -0,0 +1,5 @@
...
+<% else %>
+  <%= form.select(:oldest_accepted_version_id, form.object.possible_versions, {}, class: "form-control") %>
```
smoyth: This is where you'd use `@possible_versions` instead.

db/migrate/20191119223120_add_is_oldest_accepted_to_form_versions.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class AddIsOldestAcceptedToFormVersions < ActiveRecord::Migration[5.2]
+  def change
+    add_column :form_versions, :is_oldest_accepted, :boolean, null: false, default: true
+
+    # Revert the default before setting a single one to true (otherwise validation fails).
+    FormVersion.all.each { |fv| fv.update!(is_oldest_accepted: false) }
+    FormVersion.all.each { |fv| fv.update!(is_oldest_accepted: fv.is_current) }
```
smoyth: Nice approach. Another way to do this is to do null: true initially and then change_column_null after you do the update.

spec/models/form_spec.rb
```diff
@@ -188,6 +188,45 @@
...
+      code2 = form.current_version.code
+      number2 = form.current_version.number
+      expect(code1).not_to match(code2)
+      expect(number1).to be < number2
+
+      # make sure old v1 code/number didn't change
+      expect(form.versions.first.code).to match(code1)
+      expect(form.versions.first.number).to match(number1)
+    end
+  end
```
smoyth: Nice specs!

spec/models/odk/response_parser_spec.rb
```diff
@@ -52,8 +52,42 @@
...
+
+          it "should not error" do
+            expect do
+              Odk::ResponseParser.new(response: response, files: files).populate_response
+            end.to_not(raise_error)
+          end
+        end
+
+        context "future form version" do
+          let(:formver) { form.oldest_accepted_version.number.to_i + 1 }
```
smoyth: Nice job not getting hung up on the fact that this version doesn't really exist, and just sticking with the simplest test possible!

app/models/odk/response_parser.rb
```diff
@@ -178,10 +178,8 @@ def lookup_and_check_form(params)
...
+      raise FormVersionError, "Form version is outdated" if params[:version].to_i < oldest_accepted_num
```
smoyth: We still need to handle three letter codes for now b/c there will be forms out there with them. Perhaps the logic is:

* If it looks like a three letter code, it has to match one of the accepted versions' code exactly (you can put a comment on this that it can be removed once we're confident those are not in the wild anymore)
* Otherwise do the `<` comparison.

spec/models/odk/response_parser_spec.rb
```diff
@@ -52,8 +52,42 @@
...
+
+          it "should not error" do
+            expect do
+              Odk::ResponseParser.new(response: response, files: files).populate_response
+            end.to_not(raise_error)
+          end
+        end
+
+        context "invalid form version" do
+          let(:formver) { "wrong" }
```
smoyth: Instead of this, let's add a context block `with three-letter code version` and have examples for:

1. Accepted version
2. Valid but no-longer accepted version
3. Invalid version (like, never existed)

Does that sound right? We can delete these once we delete the code support.

spec/models/sms/decoder_spec.rb
```diff
@@ -66,14 +66,29 @@
...
+      end
+
+      it "outdated version should produce appropriate error" do
+        expect_decoding_fail(form, body: "#{v1.code} 1.15", error: "form_version_outdated")
+      end
 
-      # upgrade form version before submitting
-      old_version_code = form.current_version.code
-      form.upgrade_version!
+      it "current accepted version should be accepted" do
```
smoyth: This is a bit confusing since v2 is not the current version. Perhaps "minimum version should be accepted" ;)

spec/requests/odk/submission_spec.rb
```diff
@@ -77,7 +77,7 @@
   end
 
   context "with draft form" do
-    let(:formver) { "1234" }
+    let(:formver) { "1999010100" }
```
smoyth: Party like it's...!

app/models/odk/response_parser.rb
cooperka: Significantly over estimate already, so I'll plan to skip this for now, but completely agree that would read better

cooperka:
@smoyth done addressing feedback and ready for final review!

smoyth:
Great stuff here! Thanks for the attention to my comments. I made one commit to fix a safe navigation issue. Once CI passes we can merge this.

PR #683, opened 2019-12-05
10285 check correct mission

cooperka:
Ensure answers and option nodes don't refer to an item outside the current mission, for both web and ODK responses.

### DIFF
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -170,18 +170,6 @@ class Answer < ResponseNode
                      end
   end
 
-  # Raises a loud error if the OptionNode is not in the OptionSet (or the mission) for security purposes.
-  #
-  # Note: It would be simpler to reuse the logic that's in choice.rb to accomplish this,
-  # but `answer` isn't defined at that point for web responses, so we can't scope it there.
-  def choices_attributes=(attributes)
-    attributes.each do |(_index, item)|
-      scope = OptionNode.where(option_set_id: option_set_id)
-      raise ArgumentError if scope.find_by(id: item["option_node_id"]).nil?
-    end
-    super(attributes)
-  end
-
   # If this is an answer to a multilevel select_one question, returns the OptionLevel, else returns nil.
   def level
     option_set.try(:multilevel?) ? option_set.levels[(rank || 1) - 1] : nil
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -171,8 +171,9 @@ class Answer < ResponseNode
   end
 
   # Raises a loud error if the OptionNode is not in the OptionSet (or the mission) for security purposes.
-  # Note: This would be simpler to put in choice.rb,
-  # but `answer` isn't defined at that point in time, so we can't scope it there.
+  #
+  # Note: It would be simpler to reuse the logic that's in choice.rb to accomplish this,
+  # but `answer` isn't defined at that point for web responses, so we can't scope it there.
   def choices_attributes=(attributes)
     attributes.each do |(_index, item)|
       scope = OptionNode.where(option_set_id: option_set_id)
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -84,7 +84,8 @@ class Answer < ResponseNode
   accepts_nested_attributes_for(:choices)
 
   delegate :question, :qtype, :qtype_name, :required?, :hidden?, :multimedia?,
-    :option_set, :options, :first_level_option_nodes, :condition, :parent_group_name, to: :questioning
+    :option_set, :option_set_id, :options, :first_level_option_nodes, :condition,
+    :parent_group_name, to: :questioning
   delegate :name, to: :question, prefix: true
   delegate :name, to: :level, prefix: true, allow_nil: true
   delegate :name, to: :option, prefix: true, allow_nil: true
@@ -150,11 +151,11 @@ class Answer < ResponseNode
   # This is a temporary method for fetching option_node based on the related OptionSet and Option.
   # Eventually Options will be removed and OptionNodes will be stored on Answers directly.
   def option_node
-    OptionNode.where(option_id: option_id, option_set_id: option_set.id).first
+    OptionNode.where(option_id: option_id, option_set_id: option_set_id).first
   end
 
   def option_node_id
-    option_node.try(:id)
+    option_node&.id
   end
 
   # This is a temporary method for assigning option based on an OptionNode ID.
@@ -163,16 +164,18 @@ class Answer < ResponseNode
   # Raises a loud error if the OptionNode is not in the OptionSet (or the mission) for security purposes.
   def option_node_id=(id)
     self.option_id = if id.present?
-                       option_id = OptionNode.where(option_set: option_set).id_to_option_id(id)
+                       option_id = OptionNode.where(option_set_id: option_set_id).id_to_option_id(id)
                        raise ArgumentError if option_id.nil?
                        option_id
                      end
   end
 
   # Raises a loud error if the OptionNode is not in the OptionSet (or the mission) for security purposes.
+  # Note: This would be simpler to put in choice.rb,
+  # but `answer` isn't defined at that point in time, so we can't scope it there.
   def choices_attributes=(attributes)
-    attributes.each do |(index, item)|
-      scope = OptionNode.where(option_set: option_set)
+    attributes.each do |(_index, item)|
+      scope = OptionNode.where(option_set_id: option_set_id)
       raise ArgumentError if scope.find_by(id: item["option_node_id"]).nil?
     end
     super(attributes)
--- a/app/models/answer.rb
+++ b/app/models/answer.rb
@@ -151,7 +151,7 @@ class Answer < ResponseNode
   # This is a temporary method for fetching option_node based on the related OptionSet and Option.
   # Eventually Options will be removed and OptionNodes will be stored on Answers directly.
   def option_node
-    OptionNode.where(option_id: option_id, option_set_id: option_set_id).first
+    OptionNode.find_by(option_id: option_id, option_set_id: option_set_id)
   end
 
   def option_node_id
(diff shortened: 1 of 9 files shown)


spec/models/odk/response_parser_spec.rb
```diff
@@ -160,27 +172,54 @@
         end
 
         context "with invalid data types" do
-          let(:xml_values) { ["9.6", "The quick brown fox", "foo", "bar"] }
+          let(:xml_values) { ["9.6", "The quick brown fox", "foo", ""] }
```
cooperka: A consequence of verifying that the option node is in the current mission is that invalid option node IDs (like `bar`) will now fail. Is this acceptable? Seems reasonable unless we want to handle something like old forms being submitted with since-deleted values.

app/models/results/web_response_parser.rb
```diff
@@ -99,5 +107,17 @@ def form_items_in_mission
...
+      option_node_ids.all? { |id| option_in_mission?(id) }
+    end
+
+    def option_in_mission?(option_node_id)
+      option_node_id.blank? || option_nodes_in_mission[option_node_id].present?
+    end
+
+    def option_nodes_in_mission
+      @option_nodes_in_mission ||= OptionNode.where(mission: @response.mission).pluck(:id).index_by(&:itself)
+    end
```
cooperka: These 3 methods are duplicated now. It seems like they should go in `concerns/`, yeah? Not sure what to name it, the concept of "fooable" has never been very clear to me. Advice appreciated.

spec/models/odk/response_parser_spec.rb
```diff
@@ -143,6 +144,17 @@
             end.to raise_error(SubmissionError, /Submission contains group or question/)
           end
         end
+
+        context "response is in wrong mission" do
```
smoyth: I'm not sure if this is testing anything interesting. The mission of Response gets assigned by CanCanCan in the controller so it's always going to be appropriate. What we really need to be testing is that if you submit a response with a form ID inside the XML that does not match the mission you're submitting it to, there should be a SubmissionError. Is that covered anywhere already?

The way to do this would to setup a response _without a form_, setup the XML with a form ID from a different mission, and then let it run. In a typical flow, the form won't be set on Response when it's passed into the parser. You might even try getting rid of the `||=` on line 173 of the parser and seeing if the specs still pass. I don't get why we have that there...

spec/models/odk/response_parser_spec.rb
```diff
@@ -160,27 +172,54 @@
...
+          let(:other_form) { create(:form, mission: other_mission, question_types: question_types) }
+          let(:other_option_code) { "on#{other_form.c[3].option_set.c[0].id}" }
+          let(:xml_values) { ["Quick", "The quick brown fox", "9.6", other_option_code] }
+
+          it "should error" do
+            expect do
+              Odk::ResponseParser.new(response: response, files: files).populate_response
+            end.to raise_error(SubmissionError, /unidentifiable option node/)
+          end
+        end
```
smoyth: Having thought about this more, I think we actually can strengthen all of this to say the option node has to be not just in the mission, but also in the option set for the question. The integrity model says there can be a hard fail if you remove an option from an option set, or if you change the option set.

Sorry for not realizing this sooner! Would you agree?

app/models/concerns/report/gridable.rb
```diff
@@ -90,9 +90,9 @@ def apply_filter(rel)
   def fix_calculation_ranks
     # Need to reload calculations because otherwise the array may still contained destroyed ones.
     calculations
-      .select { |c| !c.destroyed? }
+      .reject(&:destroyed?)
       .sort_by(&:rank)
-      .each_with_index { |c,i| c.rank = i + 1 }
+      .each_with_index { |c, i| c.rank = i + 1 }
```
smoyth: Thanks for tidies!

app/models/odk/response_parser.rb
```diff
@@ -144,17 +144,41 @@ def add_group(xml_node, form_item, parent)
     end
 
     def add_answer(content, form_item, parent)
+      reject_invalid_options(content, form_item)
```
smoyth: Curious why you are doing this here instead of inside AnswerParser, which already has code for converting codes to nodes. It currently does go straight to Option ID though, so you'd have to edit it a bit.

See the line that says:

```
OptionNode.id_to_option_id(id)
```

You could change this to `OptionNode.where(option_set_id: form_item.option_set_id).id_to_option_id(id)` or something. You'd have to change the CodeMapper a bit to accept some scope information, but that seems. Reasonable. If this is confusing, let's pair on it.

I think we can also now get rid of the ` || OptionNode.old_id_to_option_id(id)` part and associated methods.

app/models/odk/response_parser.rb
```diff
@@ -144,17 +144,41 @@ def add_group(xml_node, form_item, parent)
...
+    end
+
+    def option_nodes_in_mission
+      @option_nodes_in_mission ||= OptionNode.where(mission: @response.mission).pluck(:id).index_by(&:itself)
+    end
+
     def form_item(name)
       form_item_id = form_item_id_from_tag(name)
-      unless FormItem.exists?(form_item_id)
+      form_item = FormItem.where(mission: response.mission).find_by(id: form_item_id)
```
smoyth: This looks like a good addition! I think we should also add something to `lookup_and_check_form` since that's an obvious place for some guards. We shouldn't really have to get this far before we notice the form is in the wrong mission and error out.

app/models/results/web_response_parser.rb
smoyth: May be moot if we switch to using a similar approach here, which is to use a scope when looking up the option ID. See line 163 in Answer.

app/models/odk/response_parser.rb
cooperka: Notes from pairing:
- this can be better addressed by updating answer.rb in a similar way
- old_id stuff will be done in a separate PR

spec/models/odk/response_parser_spec.rb
cooperka: [while pairing we agreed that either way is fine for this]

app/models/answer.rb
```diff
@@ -159,8 +164,23 @@ def option_node_id
...
   # Eventually Options will be removed and OptionNodes will be stored on Answers directly.
+  #
+  # Raises a loud error if the OptionNode is not in the OptionSet (or the mission) for security purposes.
   def option_node_id=(id)
-    self.option_id = id.present? ? OptionNode.id_to_option_id(id) : nil
+    self.option_id = if id.present?
+                       option_id = OptionNode.where(option_set: option_set).id_to_option_id(id)
+                       raise ArgumentError if option_id.nil?
+                       option_id
+                     end
```
cooperka: Note: As discussed, it would be more efficient to use `OptionNode.where(mission: mission)` for scoping (since option_set is delegated to questioning), but `response` isn't defined at this point so we can't use it to get the mission.

We could use `mission: option_set.mission` or `mission: questioning.mission` instead of `option_set: option_set` but that probably doesn't change much.

app/models/choice.rb
```diff
@@ -62,8 +62,20 @@ def option_node_id
...
+                       # Answer can't always be set in advance, so only scope it if it's set.
+                       scope = if answer.present?
+                                 OptionNode.where(option_set_id: answer.option_set.id)
+                               else
+                                 OptionNode
+                               end
+                       option_id = scope.id_to_option_id(id)
+                       raise ArgumentError if option_id.nil?
+                       option_id
+                     end
```
cooperka: This assertion is new since pairing. It's required in order to validate ODK select-multiple answers, since they skirt around the call to `choices_attributes=`.

I found a seemingly safe way to pass `answer` here at least for ODK; if we come up with a way to pass it in for web as well, then we could remove the `scope` conditional here and also remove `choices_attributes=` from Answer since this would cover it.

The reason it's difficult to pass `answer` in from web is because it uses a more generic call to `tree_parent.children.build` (compared to a more explicit `answer.choices.build` from ODK).

app/models/odk/answer_parser.rb
```diff
@@ -41,10 +41,9 @@ def populate_answer_value(answer, content, form_item)
       answer
     end
 
-    def option_id_for_option_node_code(option_node_id)
-      Odk::CodeMapper.instance.item_id_for_code(option_node_id)
-    rescue SubmissionError
-      nil
+    # Get the item ID, unless it's nil (normally that would raise an error).
+    def safe_item_id_for_code(code)
+      code.present? ? CodeMapper.instance.item_id_for_code(code) : nil
```
cooperka: New since pairing. We can't entirely remove `option_id_for_option_node_code` because it's doing some error prevention (CodeMapper raises if `code` is nil). Was this a collective oversight, or did you already suggest a different fix for this?

app/models/results/web_response_parser.rb
```diff
@@ -80,6 +80,13 @@ def update_or_add_node(web_hash_node, tree_parent)
 
     def new_tree_node_attrs(web_hash_node, tree_parent)
       clean_params = web_hash_node.slice(*TOP_LEVEL_PARAMS).permit(PERMITTED_PARAMS)
+      if clean_params.key?(:choices_attributes)
+        clean_params[:choices_attributes].keys.each do |key|
+          # Include mission_id FIRST so it can immediately be used in models to validate things.
+          value = clean_params[:choices_attributes][key]
+          clean_params[:choices_attributes][key] = {mission_id: @response.mission_id}.merge(value)
```
cooperka: Note: We originally had `attrs.merge!(mission_id: foo)` which sets mission last; mission needs to be set first so it exists when option_node_id is set.

cooperka:
CI is passing and we've paired on most of this by now. I'm merging to prevent conflicts with upcoming changes over the next couple days. Will make sure Tom looks this over when he's back later this week.

PR #684, opened 2019-12-05
10296 qing edit issues

cooperka:
Added failing spec then fixed it. In my manual testing, this issue only appeared for media questionings (e.g. image, video).


app/assets/javascripts/views/questioning_form_view.js.coffee
```diff
@@ -30,7 +30,7 @@ class ELMO.Views.QuestioningFormView extends ELMO.Views.QuestionFormView
     @showDefault() && (@fieldValue('default') || '').trim() != ''
 
   showRequired: ->
-    !@fieldValue('read_only') && @super.metadataTypeBlank.call(this)
+    (!@fieldValue('read_only') || @fieldValue('read_only') == 'No') && @super.metadataTypeBlank.call(this)
```
cooperka: I couldn't determine why this regressed (or maybe it's always been this way) but this seems safe. More risky would be to remove the first part of the condition if we know they'll all be `No` (rather than empty or `false`), but I didn't want to make that assumption.

smoyth: I figured out what is going on here. See `app/helpers/elmo_form_builder.rb:178`. questionings.read_only can either be nil or false in the negative case. There is no null constraint on the DB b/c read_only doesn't apply to QingGroups. This may have been a mistake. But I don't think it's worth trying to change that right now. May cause weird side effects. 

If it's false, things work as planned. If it's `nil`, then the `data-val` attrib doesn't get included and that breaks the `@fieldValue` check. The solution I think is to change `val` from `nil` to `false` around line 149 of that file. Let me know what you think...

spec/features/forms/form_item/questionings_form_spec.rb
```diff
@@ -128,15 +128,36 @@
...
+        it "should display all fields as not editable" do
+          visit(edit_qing_path)
+          expect_visible("required", true)
+          expect_editable("required", false)
+          expect_visible("hidden", true)
+          expect_editable("hidden", false)
+          expect_visible("display_logic", true)
+          expect_editable("display_logic", false, field_type: "select")
+          expect_visible("skip_logic", true)
+          expect_editable("skip_logic", false, field_type: "select")
```
smoyth: Specs look good!

app/assets/javascripts/views/questioning_form_view.js.coffee
cooperka: Wonderful, simpler fix implemented (and fixup'd into commit history). Thanks for diagnosing this.

PR #685, opened 2019-12-06
10311 fix questioning redirect on mission change

cooperka:
- Fixed error on redirect from questioning
- Added specs for every main route that can be redirected so this doesn't happen again
- Cleaned up old specs as discussed in pairing

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -64,6 +64,9 @@ Metrics/AbcSize:
 Metrics/PerceivedComplexity:
   Max: 10
 
+Metrics/ParameterLists:
+  CountKeywordArgs: false
+
 Style/AsciiComments:
   Enabled: false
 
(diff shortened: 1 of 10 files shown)


.rubocop.yml
```diff
@@ -64,6 +64,9 @@ Metrics/AbcSize:
 Metrics/PerceivedComplexity:
   Max: 10
 
+Metrics/ParameterLists:
+  CountKeywordArgs: false
```
smoyth: Nice

spec/support/helpers/request_spec_helpers.rb
```diff
@@ -24,25 +26,7 @@ def do_api_request(endpoint, params = {})
...
-    assert_response(:success)
-  end
-
-  def post_s(path, opts = {})
-    if opts.empty?
-      post path
-    else
-      post path, opts
-    end
-    assert_response(:success)
```
smoyth: Hooray!!!

PR #687, opened 2019-12-11
10327 remove old "q" support from CodeMapper

cooperka:
Split out of #683 for easier, separate review. Also easier to see which specs are affected.


cooperka:
Addressed feedback from Tom; spec passed locally; merging as discussed 🎉

PR #689, opened 2019-12-12
10330: Toggle password fields on render

cooperka:
Fixes a long-standing bug where password fields were hidden if the page refreshed, e.g. when submitting a form that has errors. Previously they were only shown on a change in status of the select dropdown.


smoyth:
Cool!

smoyth:
Cool!

PR #694, opened 2020-01-02
10356 decaffeinate

cooperka:
Changes:
- Auto-converted all coffeescript to JS
- Manually cleaned up some low-hanging fruit after the conversion
- Fixed 2866/4767 of global JS lints (continuation of #693)

Resources:
- https://decaffeinate-project.org/
- https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md


smoyth:
This is so exciting! Thank you for getting this done and for I'm sure all the little details you had to deal with to get there!

There are two of the suggestions I think we should deal with now:

1. The initClass thing. This is a pretty gross pattern that I don't want to see new files following. I pushed a commit showing how we can fix this. See also https://benmccormick.org/2015/07/06/backbone-and-es6-classes-revisited if you're interested. Hopefully your fancy IDE can help you streamline this refactor somehow? ;)
2. The IIFE thing. There are relatively few of these, the code is gross, and the fix is easy. See https://github.com/decaffeinate/decaffeinate/blob/master/docs/suggestions.md#ds205-consider-reworking-code-to-avoid-use-of-iifes. You can please doublecheck but I don't think we need the "treat as expression" behavior anywhere those are appearing.

Then we can merge these! Hooray!

cooperka:
Whoa, great fix! I tried about 3 different things to get rid of initClass and none of them worked with Backbone. Thanks I'll address your comments.

PR #696, opened 2020-01-14
Remove miniracer

smoyth:
Everything compiles swimmingly without this dependency. Since we will always have node I think we can nix it. It was throwing an error on the pi for some reason.


cooperka:
Discussed on slack, merging

PR #698, opened 2020-01-16
10384 report run btn styles

cooperka:
Styles weren't being changed effectively on https://staging.getelmo.org/en/m/sandboxkevin/reports/new. Fixes:
- Remove `secondary` class when adding `primary`, to prevent style conflicts
- Restore `disabled` attribute when un-enabling

### DIFF
--- a/app/assets/javascripts/legacy/views/report/edit_view.js
+++ b/app/assets/javascripts/legacy/views/report/edit_view.js
@@ -195,6 +195,11 @@
       }
     } else {
       button.on('click', () => { return false; });
+      button.attr('disabled', true);
+      if (name === 'run') {
+        button.addClass('btn-secondary');
+        button.removeClass('btn-primary');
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/legacy/views/report/edit_view.js
```diff
@@ -156,43 +160,59 @@
 
   // applies a given function to all panes
   klass.prototype.pane_do = function (func_name) {
-    for (let i = 0; i < this.panes.length; i++) if (this.panes[i][func_name]) this.panes[i][func_name](Array.prototype.slice.call(arguments, 1));
```
smoyth: You didn't like my 146 char line that I wrote in 2012? :P

cooperka: Fabulously dense 👏

PR #701, opened 2020-01-22
10399 smarter offline mode for maps

smoyth:
Adds a smoother experience when using the dashboard on the field server. If a connection is detected and a valid key is present, the map will load, else a nice message will be displayed.


app/helpers/application_helper.rb
```diff
@@ -94,12 +94,14 @@ def javascript_doc_ready(&block)
...
-    if configatron.has_key(:google_maps_api_key)
-      api_key = configatron.google_maps_api_key
-      javascript_include_tag("https://maps.googleapis.com/maps/api/js?key=#{api_key}&v=3")
-    end
+  def google_maps_key_missing?
+    !configatron.has_key?(:google_maps_api_key)
+  end
+
+  def google_maps_js
+    return "" if google_maps_key_missing? && !offline?
```
cooperka: Is this right? Worth a comment. It seems to return an (invalid) maps URL if the key is missing and you **are** offline.

smoyth: Rats. That should be an OR! I'll fix.

smoyth: I blame the virus...

PR #702, opened 2020-01-24
10325 hidden qings

cooperka:
- Change hidden => disabled
- Add new column + checkbox for actually hidden
- Change checkbox hint wording based on designs
- Remove exceptions to the disabled rule (e.g. timestamps); make them hidden instead
- Add "eye-slash" icon for hidden qings
- Add optional "checkbox sandwich" UI with info to the right of the checkbox instead of inside the "i" button

Before/after checkbox sandwiches:

<div>

</div>


app/decorators/odk/qing_group_decorator.rb
```diff
@@ -50,6 +50,8 @@ def header_bind_tag(xpath_prefix: "/data")
     def body_tags(xpath_prefix:, **_options)
       return main_body_tags(xpath_prefix) if root?
 
+      return safe_str unless visible? && enabled?
```
cooperka: This guard is new since pairing, to prevent rendering hidden qing groups in the ODK body. I don't quite understand:
1. why this guard wasn't previously necessary with disabled qing groups (though the guard was used before for regular qings, see qing_decorator above)
2. why `visible?` isn't defined here (causes spec failures) even though both qing and qing_group inherit from form_item where `visible?` is defined

Can you please help me understand? Happy to pair if it's easier.

config/locales/en/main.yml
```diff
@@ -650,9 +651,10 @@ en:
...
         display_logic: "Conditions which must be met in order for this question to be shown to the user."
-        hidden: "Whether this question will be shown on this form. Use this to disable a question if you can't delete it. This value will override the condition and required settings below."
-        read_only: "If checked, the enumerator won't be able to change the pre-filled value when using ODK. Does not affect web or SMS responses."
-        required: "Whether this question must be completed on this form or is optional. Only enforced if question is not hidden the condition specified above is met."
+        hidden: "Hide question but still collect default answer (ODK only)."
```
cooperka: More specific question about wording: `Hide question but still collect default answer (ODK only)` makes me assume the checkbox has **no effect** outside of ODK, when in fact it causes the question to behave as if it's **disabled** for web submissions.

Like you said, maybe we can tweak wording when we demo the checkbox sandwich.

app/helpers/elmo_form_builder.rb
```diff
@@ -50,6 +50,7 @@ def field(field_name, options = {})
       label_tag: elmo_field_label(field_name, options),
       field_html: elmo_field(field_name, options) + (options[:append] || ""),
       hint_html: elmo_field_hint(field_name, options),
+      inline_hint_html: elmo_field_inline_hint(field_name, options),
```
smoyth: 😄

app/models/form_item.rb
```diff
@@ -130,8 +131,8 @@ def base_item
     self
   end
 
-  def visible_children
-    sorted_children.select(&:visible?)
+  def visible_and_enabled_children
```
smoyth: Can we keep the name `visible_children` here? For the sake of practical use, I think that is implied. The implementation would stay the same. I could be sold on adding normalization and/or a DB constraint so that visible always implies enabled.

app/models/results/blank_response_tree_builder.rb
```diff
@@ -26,7 +26,9 @@ def build
 
     def add_level(form_nodes, response_node)
       form_nodes.each do |form_node|
-        unless form_node.hidden
+        # If we ever add hidden functionality to web forms this will need to get more granular,
+        # and the hidden check will need to happen in the partials instead.
+        unless form_node.hidden || form_node.disabled
```
smoyth: Worth a comment explaining why these don't have `?`s after them. I can't remember why.

app/models/form_item.rb
```diff
@@ -218,6 +219,10 @@ def top_level?
     depth == 1
   end
 
+  def enabled?
+    !disabled?
+  end
+
   def visible?
     !hidden?
```
smoyth: Maybe we should just define `visible?` as `!hidden? and !disabled?` That would solve a good deal of our nomenclature issues. Then you can just grep for `visible? && enabled?` and replace with `visible?`.

app/views/questionings/_form.html.erb
```diff
@@ -16,13 +16,12 @@
...
-    <%= f.field(:all_levels_required, type: :select, prompt: false,
-                                      read_only: cannot?(:update_core, @questioning),
-                                      options: translate_options(%w[false true],
-                                        [:form_item, :all_levels_required_options])) %>
+    <%= f.field(:required, type: :check_box, read_only: cannot?(:update_core, @questioning), inline_hint: true) %>
+    <%= f.field(:read_only, type: :check_box, read_only: cannot?(:update_core, @questioning), inline_hint: true) %>
+    <%= f.field(:hidden, type: :check_box, read_only: cannot?(:update_core, @questioning), inline_hint: true) %>
+    <%= f.field(:disabled, type: :check_box, read_only: cannot?(:update_core, @questioning), inline_hint: true) %>
+    <%= f.field(:all_levels_required, type: :select, read_only: cannot?(:update_core, @questioning),
+        prompt: false, options: translate_options(%w[false true], [:form_item, :all_levels_required_options])) %>
```
smoyth: Smooth

db/migrate/20200124001444_flip_odk_disabled_to_hidden.rb
```diff
@@ -0,0 +1,26 @@
...
+          if behind_the_scenes?(qing) && qing.hidden?
+            qing.update!(disabled: true, hidden: false)
+          end
+        end
+      end
+    end
+  end
+
+  def behind_the_scenes?(qing)
+    qing.metadata_type == "formstart" || qing.metadata_type == "formend" || qing.default.present?
```
smoyth: Nice

spec/features/forms/form_item/questionings_form_spec.rb
```diff
@@ -55,6 +56,20 @@
           expect_visible("hidden", true)
         end
 
+        it "should hide disabled option when metadata field has a value" do
```
smoyth: Shouldn't it be hiding _hidden_ instead of disabled when metadata has a value? Hidden is implied in that case, but disabled is not. You could still disable a metadata question if you don't want to collect data for it anymore.

spec/features/forms/form_item/questionings_form_spec.rb
```diff
@@ -170,6 +189,7 @@
       visit(edit_qing_path)
       expect_editable("required", true)
       expect_editable("hidden", true)
+      expect_editable("disabled", true)
```
smoyth: A quick-ish way to simplify this file, which is getting a bit unwieldy, might be to combine `expect_visible` and `expect_editable` into one call, `expect_field_visibility("foo", value)`, where `value` is one of `:hidden`, `:read_only`, or `:editable`. Thoughts?

spec/models/forms/condition_computer_spec.rb
```diff
@@ -151,7 +153,7 @@
       #   Q6.3.1                    SR5
       #   Q6.3.2                            SR7
       # Q7                                  SR7
-      # Q8          (nothing because it's a hidden question)
+      # Q8          (nothing because it's a disabled question)
```
smoyth: Is this true or should this stay `hidden` and you add another line to the comment for `Q9` which is `disabled`?

spec/models/forms/condition_computer_spec.rb
```diff
@@ -27,13 +27,15 @@
...
+        "integer", # Hidden question
+        "integer" # Disabled question
       ])
     end
     let(:form_items) { computer.preordered_form_items }
 
     before do
-      form.c[7].update!(hidden: true)
+      form.c[6].update!(hidden: true)
+      form.c[7].update!(disabled: true)
```
smoyth: Are you sure you're disabling the right question to match the data structure above? Seems you added another question above as the disabled one, so it should be `c[8]`, whilst `c[7]` stays as the hidden one?

spec/requests/odk/form_rendering_spec.rb
```diff
@@ -274,6 +274,28 @@
...
+
+      before do
+        form.c[2].c[0].update!(disabled: true, required: true)
+        form.c[3].c[0].update!(disabled: true)
+        form.c[4].update!(disabled: true)
+      end
+
+      it "should render proper xml" do
+        expect_xml(form, "empty_and_disabled")
+      end
```
smoyth: Nice work on this.

app/models/results/blank_response_tree_builder.rb
cooperka: Probably no reason! I'll add `?`s. Originally written here https://github.com/thecartercenter/nemo/commit/62eb4dfe047aa8fe9c01a5474c1e795b38b11f7f

spec/features/forms/form_item/questionings_form_spec.rb
cooperka: Ah good catch. Both were being hidden for metadata, I'll make sure `disabled` checkbox is always rendered.

spec/models/forms/condition_computer_spec.rb
cooperka: Yep! Silly mistake. I've fixed this but am not sure if the passing spec is testing the correct behavior... I haven't explicitly added a skip rule to Q8 but it ends up being associated with SR7, is that because it's the "end" that SR7 is skipping to? I don't think I understand this spec very well.

```
      # Q8          SR7 (hidden question)
      # Q9          (nothing because it's a disabled question)
```

app/models/form_item.rb
cooperka: Will do [discussed DB constraint on slack and not implementing that]

PR #703, opened 2020-01-27
10404 wysiwyg icons

cooperka:
Webpacker seems not to include `vendor/assets/` images by default (based on https://stackoverflow.com/a/14195512/763231), this fixes that. Confirmed via:

```bash
NODE_ENV=production RAILS_ENV=production rails assets:precompile
find public/assets -name *.svg
```

The editor icons were working on staging, probably from stale assets... not sure why staging used stale assets and health got cleaned, but images definitely get compiled now locally.


smoyth:
Nice find. Thanks for the speed!

PR #707, opened 2020-02-04
10441_fix_keyboard_enter_publish

cooperka:
No visual changes, just causes the 'save' button to be the default action on enter key instead of 'save and publish'.

Also fixes similar issue with 'save and mark reviewed' for response edit.

### DIFF
--- a/app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
+++ b/app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
@@ -33,4 +33,10 @@
     font-weight: bold;
     margin-bottom: 5px;
   }
+
+  .submit-buttons {
+    .float-right {
+      margin-left: 3px;
+    }
+  }
 }
--- a/app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
+++ b/app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
@@ -33,10 +33,4 @@
     font-weight: bold;
     margin-bottom: 5px;
   }
-
-  .submit-buttons {
-    .float-right {
-      margin-left: 3px;
-    }
-  }
 }
--- a/app/views/forms/form.html.erb
+++ b/app/views/forms/form.html.erb
@@ -53,7 +53,7 @@
       <div class="less-settings"><a href="#"><%= t("common.less_settings") %></a></div>
       <% unless read_only %>
         <div class="submit-buttons">
-          <%= f.submit(class: "btn btn-primary float-right") %>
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
```diff
@@ -33,4 +33,10 @@
     font-weight: bold;
     margin-bottom: 5px;
   }
+
+  .submit-buttons {
+    .float-right {
+      margin-left: 3px;
+    }
+  }
```
smoyth: This is a clever solution! However, instead of float, could we use flex order? It has less side effects, I think, than float:

https://getbootstrap.com/docs/4.0/utilities/flex/#order

app/views/forms/form.html.erb
```diff
@@ -53,11 +53,11 @@
       <div class="less-settings"><a href="#"><%= t("common.less_settings") %></a></div>
       <% unless read_only %>
         <div class="submit-buttons">
+          <%= f.submit(class: "btn btn-primary float-right") %>
```
smoyth: Can you please include a comment here about why we are doing this?

app/assets/stylesheets/global/html_forms/elmo_forms/_general.scss
cooperka: Adding `d-flex` makes the buttons left-aligned and removes some default padding between them, which needs to be manually undone as needed... so to support button ordering in a reusable way I'll add styles via SCSS instead of using Bootstrap, and test to make sure existing buttons work fine (including with "upload progress" elements in some cases). There are 17 existing `.submit-buttons` that I want to be careful about, but flexbox will make them all consistent instead of just here 🙌

cooperka: Done and tested in RTL language as well

app/views/user_sessions/new.html.erb
```diff
@@ -4,6 +4,8 @@
   <%= f.field(:verify, content: recaptcha_tags) if captcha_required? %>
   <div class="submit-buttons">
     <%= f.submit(:login, class: "btn btn-primary") %><br/><br/>
+  </div>
+  <div class="d-flex justify-content-end mt-3">
     <%= link_to(t("user.forgot_password"), new_password_reset_path) unless offline_mode? %>
   </div>
```
cooperka: (Moved to its own div because the old way was a bit hacky; this renders the same as it used to)

cooperka:
Discussed at standup, merging

PR #708, opened 2020-02-08
10450 fix printable_form show [hotfix]

cooperka:
Changes:

- Fix method call that was renamed
- Wait for load in spec to make sure print page renders without error (this would have caught the original issue)
- Fix incorrect JS date format `2020-0-1` => `2020-01-01`
- Bring specs back from the dead (from https://github.com/thecartercenter/nemo/commit/f562a5c56438a675c72df8ae4435d031fd9818fd) now that they work with updated webdriver
- Clear localstorage to fix flapping spec

### DIFF
--- a/spec/features/forms/form/printable_spec.rb
+++ b/spec/features/forms/form/printable_spec.rb
@@ -14,22 +14,26 @@ feature "forms", js: true do
   end
 
   describe "print from index" do
-    scenario "should work and show tip the first time" do
-      visit(forms_path)
+    context "first time" do
+      before do
+        visit(forms_path)
+      end
 
-      # First time printing should show tips.
-      find("a.print-link").click
-      expect(page).to have_css("h4", text: "Print Format Tips")
+      it "should work and show tip" do
+        find("a.print-link").click
+        expect(page).to have_css("h4", text: "Print Format Tips")
+      end
     end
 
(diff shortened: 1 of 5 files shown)


spec/features/forms/form/printable_spec.rb
```diff
@@ -14,26 +14,37 @@
...
-    #     expect(page).to have_content('Form')
-    #     date = Date.today.strftime("%Y-%m-%d")
-    #     page.execute_script("window.localStorage.setItem('form_print_format_tips_shown', '#{date}')")
-    #
-    #     # Second time printing should not show tips.
-    #     find('a.print-link').click
-    #     expect(page).not_to have_css('h4', text: 'Print Format Tips')
-    #   end
-    # end
+    context "with shown flag set" do
```
smoyth: Hmm, thanks for your energy toward getting this "back from the ☠️"! But it still seems to be failing on CI... Perhaps investigate a bit more but not sink hours into it?

cooperka:
Yeesh, apparently Selenium just needs 10 seconds (on Travis) instead of 2 seconds (locally) to render the loading element. I've changed the setting and specs pass.

Can you confirm the way I've gone about changing the setting looks reasonable?

cooperka:
Discussed at standup, merging

PR #710, opened 2020-02-11
10451: Set up Scout

cooperka:


cooperka:
Merging into today's release branch, will make this a setting later

PR #713, opened 2020-02-20
10467: Save uploaded response XML in tmp file

cooperka:


app/controllers/responses_controller.rb
```diff
@@ -168,16 +170,19 @@ def web_create_or_update
   end
 
   def handle_odk_submission
-    return render_xml_submission_failure("No XML file attached.", 422) unless params[:xml_submission_file]
+    submission_file = params[:xml_submission_file]
+    return render_xml_submission_failure("No XML file attached.", 422) unless submission_file
```
smoyth: How about

```
unless (submission_file = params[:xml_submission_file])
  return render_xml_submission_failure("No XML file attached.", :unprocessable_entity)
end
```

app/controllers/responses_controller.rb
```diff
@@ -193,6 +198,17 @@ def handle_odk_submission
...
+  # Copy the uploaded file to a temporary path we control so that if saving the response fails,
+  # we can inspect the XML to find out why.
+  def copy_to_tmp_path(submission_file)
+    upload_path = submission_file.tempfile.path
+    upload_name = File.basename(upload_path)
+    tmp_path = TMP_UPLOADS_PATH.join(upload_name)
+    FileUtils.mkdir_p(TMP_UPLOADS_PATH)
+    FileUtils.cp(upload_path, tmp_path)
+    tmp_path
+  end
```
smoyth: Very smooth.

app/controllers/responses_controller.rb
```diff
@@ -168,16 +170,19 @@ def web_create_or_update
...
+      tmp_path = copy_to_tmp_path(submission_file)
+
       @response.user_id = current_user.id
       @response.device_id = params[:deviceID]
-      @response.odk_xml = params[:xml_submission_file]
+      @response.odk_xml = submission_file
       @response = odk_response_parser.populate_response
       authorize!(:submit_to, @response.form)
-      @response.save(validate: false)
+      FileUtils.rm(tmp_path) if @response.save(validate: false)
```
smoyth: I think you can just do 

```
@response.save(validate: false)
FileUtils.rm(tmp_path)
```

`save` doesn't return anything meaningful if called with `validate: false`

spec/requests/media/odk_response_spec.rb
```diff
@@ -25,6 +25,20 @@
...
+
+    it "should save a temp file for failures" do
+      submission_file = prepare_and_upload_submission_file("no_version.xml")
+
+      post submission_path, params: {xml_submission_file: submission_file},
+                            headers: {"HTTP_AUTHORIZATION" => encode_credentials(user.login, test_password)}
+      expect(response).not_to have_http_status(:created)
+
+      tmp_files = Dir.glob(ResponsesController::TMP_UPLOADS_PATH.join("*.xml"))
+      expect(FileUtils.rm(tmp_files)).not_to be_empty
```
smoyth: You should probably clean out the `TMP_UPLOADS_PATH` dir before each of these specs (in a before block? use `FileUtils.rm_rf`?). Else you will end up with cross contamination between runs potentially.

PR #721, opened 2020-03-18
10479 answer option id refactor

smoyth:


cooperka:
Note: I just rebased with numerous conflicts. In particular, @smoyth can you please verify the changes from https://github.com/thecartercenter/nemo/commit/5fcd84c09f3237ead51a7a37358048e3d882780d are adequately kept?

- `prep_form_vars` was previously added to a method that was removed here
- `unless huge?` no longer exists so some of your comment tweaks here were dropped

app/models/option_node.rb
```diff
@@ -240,8 +240,12 @@ def arrange_as_json(hash = nil)
...
         #
-        # This should ideally consult the `destroy` permission for option_node in the authorization system,
-        # but it all needs to be refactored anyway; it's not worth doing more gross things just for this.
+        # This should ideally consult the `destroy` permission for option_node in the authorization system.
+        # But Ability is not easily accessed from models. Also, the option set form is due for a refactor
+        # and will likely not need JSON serialization in the future. So for now we'll just duplicate
+        # the logic to test if removal is allowed. See the `user_independent_permissions` method
+        # in Ability for the source of this.
+        # branch[:removable] = !data? && !in_use?
         branch[:removable] = true
```
cooperka: Again drawing attention to this area with rebase conflicts:

`= true` was on develop. I left your new code commented out here, otherwise it would not exist in git history.

app/models/option_set.rb
```diff
@@ -204,8 +204,14 @@ def published?
...
-  # or if any copies are used in at least one question.
+  # Whether this OptionSet appears on any Questionings that have data.
+  # Since Questionings can't change OptionSet once they have data, this is sufficient.
+  def data?
+    !standard? && questionings.any?(&:data?)
+  end
+
+  # Whether this OptionSet is used any Questions. We don't need to check conditions
+  # because a condition can't refer to a node in this set without the set being used in a question.
   def in_use?
```
cooperka: All these new comments are really helpful, I appreciate it 👏

app/models/choice.rb
```diff
@@ -63,46 +58,6 @@ def coordinates?
...
-    self[:option_id] =
-      if node.nil?
-        nil
-      else
-        node.option_id # Temporary until we get rid of option_id column.
-      end
-    super
-  end
-
-  def option_node_id=(id)
```
cooperka: 🎉

db/migrate/20200226173054_populate_new_option_node_id_columns.rb
```diff
@@ -0,0 +1,134 @@
...
+    null_choices = Choice.where(option_node_id: nil).where.not(option_id: nil)
+      .includes(answer: {form_item: :question})
+    return [] if null_answers.none? && null_choices.none?
+
+    pairs = null_answers.map { |a| [a.option_set_id, a.option_id] }
+    pairs.concat(null_choices.map { |c| [c.answer.option_set_id, c.option_id] })
+  end
+
+  def print_missing_pairs(pairs)
+    puts "Some answers/choices refer to options not contained in the expected option set:\n"
```
cooperka: I ran into this while migrating locally; it's unclear at a glance when this is a problem or not. Can you clarify how to know if a manual fix is needed vs. when to just touch the flag?

Is this related to the security checks we recently added to prevent submitting invalid options?

db/migrate/20200226173054_populate_new_option_node_id_columns.rb
```diff
@@ -0,0 +1,134 @@
...
+          AND choices.option_node_id IS NULL
+          AND choices.answer_id = answers.id
+          AND answers.questioning_id = form_items.id
+          AND form_items.question_id = questions.id
+          AND questions.option_set_id = option_nodes.option_set_id
+          AND option_nodes.option_id = choices.option_id
+    SQL
+  end
+
+  def execute_partitioned(query, table, parts = 100)
```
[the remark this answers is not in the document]
smoyth: Yes I tested on a big old DB. No guarantee something out there won't break it but that's why it logs verbosely.

app/models/option_node.rb
smoyth: Yes I think it's fine.

smoyth: I removed the commented code. `= true` is the most recent.

db/migrate/20200226173054_populate_new_option_node_id_columns.rb
smoyth: Kind of related to the security fix.

Manual fix is needed if it seems like high stakes data. I'll add more text to clarify:

```
    puts "If the option set is not in active use, feel free to use the automated fix."
    puts "If the option set is in active use, consider the effects of adding new options to it."
    puts "You may prefer to manually delete the answers/choices in question.
```

smoyth:
@cooperka We will perhaps need to work together on resolving those missing option errors in the migration. Deploy will be slower this time around.

smoyth:
@cooperka and yes, it's DONE :D

PR #724, opened 2020-04-10
10461 server to server first steps

smoyth:


smoyth:
@cooperka this code is not super finished but I'd like to get it in the mainline so it doesn't deviate from develop. It's not exposed to the frontend at all. We'll likely build on it in the future. What do you think?

db/migrate/20200410153801_remove_original_id_foreign_keys.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class RemoveOriginalIdForeignKeys < ActiveRecord::Migration[5.2]
+  def change
+    # Removing these because there is not a big downside and the upside is we can
+    # still use original_id across systems if we don't have to meet this requirement.
+    remove_foreign_key "forms", "forms"
```
cooperka, line 7: I don't think I understand the motivation for this, can you clarify?

app/models/cloning/importer.rb
```diff
@@ -0,0 +1,26 @@
...
+    end
+
+    def import
+      ApplicationRecord.transaction do
+        # Defer constraints so that constraints are not checked until all data is loaded.
+        SqlRunner.instance.run("SET CONSTRAINTS ALL DEFERRED")
+        Zip::InputStream.open(zip_file) do |io|
+          while (entry = io.get_next_entry)
+            klass = entry.name.sub(/.csv$/, "").tr("_", ":").constantize
+            klass.copy_from(io)
```
cooperka, line 20: 🚀

db/migrate/20200323001444_change_foreign_keys_to_deferrable.rb
```diff
@@ -0,0 +1,139 @@
...
+    remove_foreign_key "taggings", name: "taggings_question_id_fkey"
+    remove_foreign_key "taggings", name: "taggings_tag_id_fkey"
+    remove_foreign_key "tags", name: "tags_mission_id_fkey"
+    remove_foreign_key "user_group_assignments", name: "user_group_assignments_user_group_id_fkey"
+    remove_foreign_key "user_group_assignments", name: "user_group_assignments_user_id_fkey"
+    remove_foreign_key "user_groups", name: "user_groups_mission_id_fkey"
+    remove_foreign_key "users", name: "users_last_mission_id_fkey"
+    remove_foreign_key "whitelistings", name: "whitelistings_user_id_fkey"
+
+    add_foreign_key "answers", "form_items", column: "questioning_id", deferrable: true
```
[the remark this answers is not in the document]
smoyth: Hmm I'm glad you pointed this out. It turns out that this doesn't actually work and gets silently ignored. I need to take out this migration I think. Or at least the deferred: true part. For now any time we need this functionality we'll have to hand-migrate the database in question.

But to answer your question, no it shouldn't change anything else.

db/migrate/20200410153801_remove_original_id_foreign_keys.rb
smoyth:
```
    # The only thing this foreign key is used for is for tracking which original object
    # copies point to so that they are not duplicated during subsequent copies.
    # (e.g. so the same question is not copied twice when two different forms that use
    # it are imported to a mission).
    # But if there is a constraint on the foreign key then it means we can't use it across 
    # different servers, which could end up being useful.
    # So let's remove the constraint.
```

PR #725, opened 2020-04-20
10654: Fix answer map bug, add spec, clean up code

smoyth:


smoyth:
Thanks @cooperka for reviewing. Let me know if the CI fails.

PR #735, opened 2020-05-27
Explicitly require eventmachine (for homebrew)

cooperka:
Fixes:

```
Unable to load the EventMachine C extension; To use the pure-ruby reactor, require 'em/pure_ruby'
nemo/vendor/bundle/ruby/2.4.0/gems/activesupport-5.2.4.2/lib/active_support/dependencies.rb:291:in `require': dlopen(nemo/vendor/bundle/ruby/2.4.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle, 9): Library not loaded: /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib (LoadError)
  Referenced from: nemo/vendor/bundle/ruby/2.4.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle
  Reason: image not found - nemo/vendor/bundle/ruby/2.4.0/gems/eventmachine-1.2.7/lib/rubyeventmachine.bundle
```

I tried quite a few other suggested workarounds to no avail; this comment did it for me https://github.com/oneclick/rubyinstaller2/issues/96#issuecomment-548249647

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,7 @@ gem "dalli", "~> 2.7.4" # memcache
 gem "delayed_job_active_record", "~> 4.1.3"
 gem "descriptive_statistics", "~> 2.5.1", require: "descriptive_statistics/safe" # mean, median, etc.
 gem "draper", "~> 3.0.1"
+gem "eventmachine", "~> 1.2", platform: :ruby
 gem "exception_notification", "~> 4.2"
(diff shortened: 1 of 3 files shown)


smoyth:
My only concern here would be performance (switching from native to Ruby) but it seems eventmachine is only needed by `thin` which is not used in production, so this should be fine.

PR #736, opened 2020-05-28
10595 OData resources + cached json

cooperka:
This includes all the remaining code from the demo plus more specs and cleanup.

### Changes

- The `response_json_generator` can turn a Response into an OData JSON blob that matches the metadata (thanks Tom!)
- The `PopulateResponseCachedJson` migration can cache all existing responses via the new `cached_json` column
  - Any response that has already been cached will be skipped; env `FORCE_REDO` can disable this feature
  - All responses will be migrated by default; env `SUBSET` can be used to toggle filter logic to only migrate responses from a specific mission (for testing/debugging/demos)
- Resource routes (e.g. `/en/m/foo/odata/v1/Responses-formuuid123`) now spit out the `cached_json` for each response
- Pseudo group codes and other strings are generated using a temporary `vanilla` method that removes characters that may break Power BI
  - We still want form names to look like `Responses: My mission - with symbols and spaces` when displayed in Power BI, but SOME symbols break Power BI

### For a future PR

- Groups with the same name may collide, which may break Power BI; this will be fixed once we enforce unique `group_code`s
- We need a better way to deal with `vanilla`, it's smelly
- This PR is tested and verified in Power BI with basic repeat groups. More testing needs to be done before public release
  - For example with complex types like cascading geographic, and what happens when you add/remove qings from a form
  - I have these things written down to validate as part of 10704

### DIFF
--- a/config/initializers/string.rb
+++ b/config/initializers/string.rb
@@ -11,6 +11,11 @@ class String
     mb_chars.normalize(:kd).gsub(/[^\x00-\x7F]/n, "")
   end
 
+  # Temporary method to rid a string of pesky characters that might annoy Power BI.
+  def vanilla
+    normalize.gsub(/[^a-z0-9._\- ]/i, "").to_s
+  end
+
   def ucwords
     split(" ").map(&:capitalize).join(" ")
   end
--- a/db/migrate/20200508155813_populate_response_cached_json.rb
+++ b/db/migrate/20200508155813_populate_response_cached_json.rb
@@ -6,15 +6,17 @@ class PopulateResponseCachedJson < ActiveRecord::Migration[5.2]
   disable_ddl_transaction!
 
   def up
-    # TODO: Temporary filter; sandbox haitimalaria riverblindnessnigeria2017
-    mission_ids = %w[sandbox haitimalaria riverblindnessnigeria2017].map do |compact_name|
+    # Optionally limit to a subset of missions (if env flag is set).
+    mission_ids = %w[sandbox].map do |compact_name|
       Mission.find_by(compact_name: compact_name).id
     end
-    # form_ids = Form.live.distinct(:id).pluck(:id)
-    form_ids = Form.live.distinct(:id).map do |form|
-      form.id if form.name.match?(/CDDAtt/i)
+
+    # Optionally limit to a subset of live forms (if env flag is set).
+    form_ids = Form.live.map do |form|
+      form.id if form.name.match?(/.*/i)
     end.compact
-    responses = Response.where(mission_id: mission_ids, form_id: form_ids)
+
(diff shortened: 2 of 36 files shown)


config/initializers/string.rb
```diff
@@ -11,6 +11,11 @@ def normalize
     mb_chars.normalize(:kd).gsub(/[^\x00-\x7F]/n, "")
   end
 
+  # Temporary method to rid a string of pesky characters that might annoy Power BI.
+  def vanilla
+    normalize.gsub(/[^a-z0-9._\- ]/i, "").to_s
+  end
```
cooperka, line 17: As noted in PR description, this is smelly. I should figure out how to remove it but need to prioritize other work this week.

It's used many places because Power BI rejects certain symbols, BUT we don't want to remove everything or the string will look ugly to humans using the interface (top level `ResponsesMymissionwithsymbols` vs. `Responses: My mission - with symbols and spaces`)

cooperka:
Spec failure is flapper

config/initializers/string.rb
smoyth: Yeah the way to do this later I think will be decorators. Fine for now as long as you have a note somewhere to get rid of it later.

db/migrate/20200508155813_populate_response_cached_json.rb
```diff
@@ -0,0 +1,44 @@
...
+  def up
+    responses = Response.where(filters)
+    remaining_responses = ENV["FORCE_REDO"] ? responses : responses.where(cached_json: nil)
+    cache_responses(remaining_responses)
+  end
+
+  def filters
+    if ENV["SUBSET"]
+      {
+        mission_id: %w[].map do |compact_name|
```
smoyth, line 17: Not sure what is going on here. `[].map` anything is just `[]` right?

spec/requests/odata/resource_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+def json_for(form, response, **answers)
+  {
+    "ResponseID": response.id,
+    "ResponseShortcode": response.shortcode,
+    "FormName": form.name,
+    "ResponseSubmitterName": response.user.name,
+    "ResponseSubmitDate": response.created_at.iso8601,
+    "ResponseReviewed": false
+  }.merge(answers)
+end
```
smoyth, line 61: Nice specs!

db/migrate/20200508155813_populate_response_cached_json.rb
cooperka: Yes this is just a placeholder (was useful before for demo and may be useful during the deploy migrations but currently is blank behind an env flag)

PR #741, opened 2020-06-09
10597 - tweaks after manual testing

cooperka:
Fixes for Power BI (for several scenarios with nested repeat groups), minor cleanup, and minor performance improvements.

### DIFF
--- a/app/models/results/response_json_generator.rb
+++ b/app/models/results/response_json_generator.rb
@@ -21,7 +21,7 @@ module Results
       object["ResponseReviewed"] = response.reviewed?
       root = response.root_node_including_tree(:choices, form_item: :question, option_node: :option_set)
       add_answers(root, object) unless root.nil?
-      add_nil_answers(object, response)
+      add_nil_answers(response.form, object)
       object
     end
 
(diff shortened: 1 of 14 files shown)


app/models/results/response_json_generator.rb
```diff
@@ -49,7 +50,37 @@ def add_group_answers(group, object)
...
       else
-        add_answers(group, object[group_key(group)] = {})
+        subgroup = object[node_key(group)] = {}
+        add_answers(group, subgroup)
+      end
+    end
+
+    # Make sure we include everything specified by metadata in our output,
+    # even if an older Response didn't include that qing/group originally.
+    def add_nil_answers(node, object)
```
smoyth, line 60: Can we call this var `item` or `form_item` to distinguish it from `node` which is short for `ResponseNode`.

Also, looking back, `object` is a terrible var name to use throughout here. My bad. `json_object` would be much preferable. Would you mind fixing?

cooperka:
Variable names improved, and I also cherry-picked the parallelization code for the migration.

db/migrate/20200508155813_populate_response_cached_json.rb
```diff
@@ -33,12 +33,13 @@ def cache_responses(responses)
     ActiveRecord::Base.logger.level = 1
 
     total = responses.count
-    curr = 0
-    responses.find_each do |response|
-      puts "Updating #{response.shortcode}... (#{curr += 1} / #{total})"
+    start = Time.now
+    Parallel.each_with_index(responses.find_each, in_processes: Etc.nprocessors) do |response, index|
```
smoyth: 😎

PR #742, opened 2020-06-17
7817 upgrade ruby

cooperka:
- [Release notes](https://www.ruby-lang.org/en/news/2019/12/25/ruby-2-7-0-released/)

Note: Some deprecation warnings are logged on start. Most should be fixed by upgrading the related dependency after patches are released; several have already been addressed; here's what remain:

```
gems/draper-3.0.1/lib/draper/delegation.rb:10: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
gems/activesupport-5.2.4.3/lib/active_support/core_ext/module/delegation.rb:154: warning: The called method `delegate' is defined here
gems/activerecord-5.2.4.3/lib/active_record/type.rb:27: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
gems/activerecord-5.2.4.3/lib/active_record/type/adapter_specific_registry.rb:9: warning: The called method `add_modifier' is defined here
gems/actionpack-5.2.4.3/lib/action_dispatch/middleware/stack.rb:37: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
gems/actionpack-5.2.4.3/lib/action_dispatch/middleware/static.rb:111: warning: The called method `initialize' is defined here
gems/will_paginate-3.1.8/lib/will_paginate/railtie.rb:67: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
gems/activesupport-5.2.4.3/lib/active_support/rescuable.rb:51: warning: The called method `rescue_from' is defined here
```


cooperka:
Confirmed via call

PR #744, opened 2020-06-29
10823 fix server 500 with unicode usernames

cooperka:
Collect does not yet support sending UTF-8 encoded Basic auth, but [is open to adding support](https://forum.getodk.org/t/support-for-special-characters-in-usernames-basic-auth/27696/4) if we submit a PR. Either way, we should not throw up a 500 error if a username is submitted with invalid encoding.

This PR
1. converts the input to UTF-8 if necessary,
2. allows a broader set of characters to be in the username (specifically `[:word:]` and `.`), and
3. properly returns 401 if the username doesn't exist, regardless of encoding.

Do you agree this is a reasonable route to go? If so, TODO:
- [x] Add spec for Basic auth flow
- [x] Update a couple of error message strings to reflect the newly allowable characters

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -50,7 +50,7 @@ gem "webpacker", "~> 4.2"
 
 # Authz and Authn
 gem "activerecord-session_store", "~> 1.1.1"
-gem "authlogic", "~> 4.4.2"
+gem "authlogic", "~> 6.1"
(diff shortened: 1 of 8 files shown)


cooperka:
Hound config wasn't reading our version correctly; it's now fixed.

cooperka:
Approved on Slack

PR #746, opened 2020-07-14
10920: Fix integers missing from truncated CSV export

cooperka:
Detect `long_text` answers and only truncate those vals. Somehow this was completely overlooked when I originally implemented it and **every** `value` field was truncated 😊 

Will hotfix onto health.


cooperka:
[Approved]

PR #748, opened 2020-07-17
10837: some OData loose ends

cooperka:
See commit messages for info.


config/initializers/o_data.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module OData
+  NAMESPACE = "NEMO"
+  BASE_PATH = "/odata/v1"
+end
```
cooperka, line 6: Note: It's in `initializers` because routes.rb depends on NAMESPACE and that seemed to be the only place that routes could safely load from.

db/migrate/20200508155813_populate_response_cached_json.rb
```diff
@@ -35,7 +35,7 @@ def cache_responses(responses)
     total = responses.count
     start = Time.zone.now
     num_procs = ENV["NUM_PROCS"] ? ENV["NUM_PROCS"].to_i : Etc.nprocessors
-    Parallel.each_with_index(responses.find_each, in_processes: num_procs) do |response, index|
+    Parallel.each_with_index(responses.find_each, in_processes: num_procs, isolation: true) do |response, index|
```
cooperka, line 38: Note: without `isolation`, global `I18n.locale` is shared across re-used processes.

PR #750, opened 2020-07-28
10907 odata performance improvements

cooperka:
18 hours straight to migrate GW on 16 cores seems excessive... This is an attempt at low-hanging fruit. Total speedup is ~20-40% depending on form structure.


cooperka:
Failure was a flapper

Gemfile
```diff
@@ -113,13 +113,21 @@ gem "ruby-jmeter", "~> 2.13.4"
...
   gem "thin", "~> 1.7.0"
+
+  # Great for debugging i18n paths. Uncomment temporarily when needed.
+  # gem "i18n-debug", "~> 1.1"
+
+  # Profiling. Uncomment temporarily when needed.
+  # gem "rack-mini-profiler", "~> 2.0"
+  # gem "memory_profiler", "~> 0.9.14" # Append to URL: ?pp=profile-memory
+  # gem "flamegraph", "~> 0.9.5" # Append to URL: ?pp=flamegraph
+  # gem "stackprof", "~> 0.2.15"
```
smoyth, line 130: Cool!

app/models/application_record.rb
```diff
@@ -14,4 +14,12 @@ def self.clone_options(**options)
     end
     class_variable_set("@@clone_options", options)
   end
+
+  # Identical to calling update! but with validate: false.
+  def update_without_validate!(**args)
+    with_transaction_returning_status do
+      assign_attributes(**args)
+      save!(validate: false)
```
smoyth, line 22: I don't think ! makes sense if validations are off. You can just do `save`

db/migrate/20200508155813_populate_response_cached_json.rb
```diff
@@ -40,8 +40,8 @@ def cache_responses(responses)
     total = responses.count
     start = Time.zone.now
     num_procs = ENV["NUM_PROCS"] ? ENV["NUM_PROCS"].to_i : Etc.nprocessors
-    Parallel.each_with_index(responses.find_each, in_processes: num_procs,
-                                                  isolation: true) do |response, index|
+    Parallel.each_with_index(responses.each, in_processes: num_procs,
```
smoyth, line 43: The change to `each` will mean all responses and answers are loaded into memory at once won't it?

Check this out: https://stackoverflow.com/a/57514911

cooperka: Theoretically yes, but in practice it changed nothing. I tried on my local machine with the full Health dataset and it ran smoothly either way. I wonder if Parallel batches internally.

Note: this change was necessary to support `order` above.

PR #755, opened 2020-08-17
10996 fix blank lists after searching

cooperka:
Motivation:
- Previously, `page` was not reset when doing a search with the React search filter component. If you were on page 2 and searched for something that only had 1 page of results, you'd see a **blank** page saying e.g. "showing all 8 questions".

Core changes:
- Reset `page` on search
- Clean up messy snapshot tests a bit

Additional tweaks:
- Link to responses using the React search UI (`form-id:`) instead of textbox (`form:`)
- Fix `ActionController::UnknownFormat` when clicking standard from report links to responses


cooperka:
Spec is a rare flapper.

PR #756, opened 2020-08-17
9948 clear level_names on save

cooperka:
> Looks like this is happening when a single level option set has an option level name defined, which is not a typical scenario (would involve checking the 'multilevel' box filling in the name, then unchecking it). And this shouldn't really even be possible. issue to correct this: "Clear option level names for single-level option sets on save"

Diagnostics:
- The checkbox has been disabled for years, so I'm not sure how this would have happened in the first place: [blame for option_set_form.js](https://github.com/thecartercenter/nemo/blame/cc3b8ec979f7301b4f3e11bd14c3e1ea40b28815/app/assets/javascripts/legacy/views/option_set_form.js#L95)
- Ah, it must be from OptionSet imports! The header was used for `level_names` even if it was single-level

### DIFF
--- a/app/models/option_set.rb
+++ b/app/models/option_set.rb
@@ -355,6 +355,7 @@ class OptionSet < ApplicationRecord
   def normalize_fields
     self.name = name.strip
     self.allow_coordinates = false unless geographic?
+    self.level_names = nil unless multilevel?
     true
   end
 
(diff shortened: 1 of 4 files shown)


cooperka:
Spec failures are real, will work on it.

PR #757, opened 2020-09-02
10596 automated re-caching when things change

cooperka:
Changed:
- Added Wisper to listen for changes that would require updating the OData cached_json
- Ignore non-live responses when caching batches for performance

Future work:
- Ability to isolate metadata updates (instead of full response caching) for performance (includes benchmarking JSONB updates etc.)

### DIFF
--- a/app/jobs/cache_o_data_job.rb
+++ b/app/jobs/cache_o_data_job.rb
@@ -69,7 +69,7 @@ class CacheODataJob < ApplicationJob
   end
 
   def cache_batch
-    responses = Response.dirty.order(created_at: :desc).limit(BATCH_SIZE)
+    responses = Response.dirty.live.order(created_at: :desc).limit(BATCH_SIZE)
     responses.each_with_index do |response, index|
       CacheODataJob.cache_response(response, logger: Delayed::Worker.logger)
       update_notes if (index % NOTES_INTERVAL).zero?
@@ -94,7 +94,8 @@ class CacheODataJob < ApplicationJob
 
   # Self-enqueue a new batch if there are responses left to cache.
   def loop_or_finish
-    if Response.exists?(dirty_json: true)
+    # TODO: Is this join performant enough with big data?
+    if Response.live.exists?(dirty_json: true)
       self.class.perform_later
     else
       complete_operation
--- a/spec/models/odata/cache_listener_spec.rb
+++ b/spec/models/odata/cache_listener_spec.rb
@@ -1,0 +1,110 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+describe OData::CacheListener do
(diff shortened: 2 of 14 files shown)


app/jobs/cache_o_data_job.rb
```diff
@@ -94,7 +94,8 @@ def enqueue_operation
 
   # Self-enqueue a new batch if there are responses left to cache.
   def loop_or_finish
-    if Response.exists?(dirty_json: true)
+    # TODO: Is this join performant enough with big data?
+    if Response.live.exists?(dirty_json: true)
```
cooperka, line 98: I will benchmark this

spec/models/odata/cache_listener_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      expect_responses_dirty(false, false, true)
+    end
+  end
+
+  # Reload all responses and expect certain ones to be dirty.
+  def expect_responses_dirty(*values)
+    responses.each(&:reload)
+    expect(responses.pluck(:dirty_json)).to eq(values)
+  end
+end
```
smoyth, line 97: Very smooth test!

smoyth (commented):
Looks great to me! Great application of Wisper!

cooperka:
> Why are you ignoring non-live responses in this PR?

This is just a performance tweak so that we don't keep re-caching responses to unpublished forms (only live forms can be accessed via the odata api).

If you prefer, we could instead "order by" so that they're still cached in the background and are ready as soon as it's re-published, but are prioritized *after* live responses?

smoyth:
I didn't realize only live forms were available on API. Paused forms should be too. In fact paused forms may be the primary type that ends up being accessed that way. I'm fine with not showing responses to draft forms.

spec/requests/odata/root_spec.rb
```diff
@@ -20,14 +20,17 @@
...
-      names = ["Responses: #{form.name}", "Responses: #{form_with_no_responses.name}"]
-      urls = %W[Responses-#{form.id} Responses-#{form_with_no_responses.id}]
+      forms = [form, form_with_no_responses, paused_form]
+      values = forms.map do |form|
+        {
+          name: "Responses: #{form.name}",
+          kind: "EntitySet",
+          url: "Responses-#{form.id}"
+        }
+      end
```
smoyth, line 30: Is this stuff supposed to be in this commit?

smoyth (commented):
Looks good. Just one question.

spec/requests/odata/root_spec.rb
cooperka: This is just a small refactor after I added a third form `paused_form` to the array here and I wanted to DRY up the logic

PR #761, opened 2020-09-16
11106 sentry

cooperka:
Changes:
- Install Sentry with basic user/version context and async reporting (uses DelayedJob instead of blocking the UI to send error reports)
- Use pry-byebug per [this rails guide](https://guides.rubyonrails.org/debugging_rails_applications.html#debugging-with-the-byebug-gem). It has replaced pry-nav as the default recommendation; it has a larger community and the same commands (plus dynamic breakpoints)

Screen Shot 2020-09-16 at 09 08 50


cooperka:
Spec is a flapper

smoyth (approved):

smoyth:
@cooperka One comment I just thought of—what happens if someone with a NEMO instance is not using Sentry? Will this just silently be ignored?

cooperka:
Discussed on Slack and at standup; merging!

PR #762, opened 2020-09-16
11052 report caching

cooperka:
Large reports were not being cached because they exceeded the memcached limit. Enabling compression makes them ~5-10x smaller in practice, with a minor CPU overhead that seems worthwhile.

Even with compression, many reports will still exceed the default 1 MB memcached limit. We should probably increase that in production via `/etc/memcached.conf`

Proposal:
- `-m 128` (2x the default total memory limit, in MB)
- `-I 16m` (16x the default per-item memory limit -- may as well? We cache relatively few things that are shared by many users, and relatively few reports are actually this big)

Some reports (like LEON "MOY Analysis") uncompressed are **137.4 MB**!! Compressed is 10.1 MB, and data over the wire to the client is 3.6 MB. There's nothing I can think of to speed up ridiculous reports aside from heavy limiting and pushing people toward CSV exports.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -15,11 +15,15 @@ ELMO::Application.configure do
   config.consider_all_requests_local = true
 
   # Caching may need to be turned on when testing caching itself.
-  # To do so, run `rails dev:cache` which will create this tempfile.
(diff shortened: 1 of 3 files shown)


smoyth (commented):
Nice change and documentation!

PR #763, opened 2020-09-18
11084: Load config from ENV vars optionally

smoyth:
Moving toward using ENV for everything. This as a first step lets one do so on new servers. 

Next step would be to change all servers to do it this way. Once that's stable, we could refactor code to use ENV directly instead of configatron or Settings.


config/initializers/paperclip.rb
```diff
@@ -3,20 +3,25 @@
...
     path: upload_path,
     storage: "fog",
     fog_credentials: {
       provider: "AWS",
-      aws_access_key_id: Settings.aws.access_key_id,
-      aws_secret_access_key: Settings.aws.secret_access_key,
-      region: Settings.aws.region,
+      aws_access_key_id: ENV["NEMO_AWS_ACCESS_KEY_ID"] || Settings.aws.access_key_id,
+      aws_secret_access_key: ENV["NEMO_AWS_SECRET_ACCESS_KEY"] || Settings.aws.secret_access_key,
+      region: ENV["NEMO_AWS_REGION"] || Settings.aws.region,
```
cooperka, line 21: I alluded to this at standup, but I would love to keep all ENV config as named variables rather than accessing dynamically every time we use them. We could call it `Settings` or `Constants` or anything, that doesn't matter to me. For a few reasons
1. Easier to type `Settings.foo` vs. `ENV["FOO"]`
1. Easier to autocomplete assigned variables vs. plain strings
1. Allows defaulting them in one place and using the default later. For example I could see `ENV["NEMO_OFFLINE_MODE"] == "true"` proliferating, then one day we want to default it to `"true"` in some specific scenario if it's undefined
1. Allows viewing all possible ENV settings in one place. For example if we add a new feature flag `ENV["SPECIAL_FEATURE"]` but forget to add it to `.env.example`, the only way a new developer would know about it would be grep the whole codebase for `"ENV"` and figure out which settings are possible to use. I've run into this far too many times on open source codebases!

`.env` would still be the one single place where servers are configured, but `Settings` would be where Rails learns about the config. Thoughts?

cooperka (approved):
Paired and agreed to use OpenStruct in an initializer to define settings. We'll clean up all the ENV throughout the codebase after that's done (later). First server that uses this new .env will need full testing. PR LGTM 👍

PR #764, opened 2020-10-09
11139: Overhaul caching on dashboard

smoyth:
Overhauls dashboard caching to use low-level caching instead of fragment caching, following in the direction pioneered by @cooperka  :)

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -41,65 +41,39 @@ class WelcomeController < ApplicationController
 
   private
 
-  def cache_key
-    @cache_key ||= [
-      # we include the locale in the cache key so the translations are correct
-      I18n.locale.to_s,
-
-      # obviously we include this
-      Response.per_mission_cache_key(current_mission),
-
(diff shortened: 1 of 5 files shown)


app/controllers/welcome_controller.rb
```diff
@@ -123,6 +97,29 @@ def dashboard_index
     end
   end
 
+  # Yields to a block and caches the result and stores in an ivar with the given `name`.
+  # Computes a cache key based on given dependencies (looked up in `cache_keys`) and on `name`.
+  def instance_variable_cache(name, dependencies: %i[responses enumerator_id], **options)
+    key = dependencies.map { |d| cache_keys.fetch(d) } << name
+    instance_variable_set(name, Rails.cache.fetch(key, **options) { yield })
+  end
```
cooperka, line 105: This is pretty clever, passing dependencies to base the keys on 💯

cooperka (approved):
Tested with heavy data locally and it seems to work great for both the map and the reports 🙌 This is some great cleanup.

I don't love the indirection of explicit variables `@responses` to dynamic variables `"@responses"`` but I see how it's useful.

PR #766, opened 2020-10-12
11166 fix OData default value for Collections

cooperka:
It was hard to find conclusive docs; I finally found some context in https://issues.oasis-open.org/browse/ODATA-1071. OData explicitly disallows Collections from being null, requiring that they be empty arrays instead. For Collections, the Nullable attribute applies to its children instead of the property itself.

Hotfixing and re-caching on Secure2 immediately so Zied can use it.


cooperka:
Specs are flappers

smoyth (commented):
Nice find and nice fix!

PR #767, opened 2020-10-13
11180: Add OData responses_slug feature flag

cooperka:
This will be flagged on the Mali server only.

```yml
use_data_factory: true
```

### Changes

If toggled on:
- Use a url-safe slug instead of user-friendly string for entity names
- Stringify complex types like arrays and objects (this affects both metadata and output)

### Example

New example path: http://nemo.test/en/m/fakemission7718/odata/v1/Responses-Thousands-2d8a93ba-494e-4d84-9c6e-0329fb7045fb

Screen Shot 2020-10-22 at 08 52 51

Tested and verified with Azure Data Factory.


config/initializers/o_data.rb
```diff
@@ -3,4 +3,27 @@
...
+      responses_slug(form_id, form_name)
+    else
+      "Responses-#{form_id}"
+    end
+  end
+
+  # Both user-friendly-ish and URL safe.
+  def self.responses_slug(form_id, form_name)
+    "Responses-#{form_name.gsub(/[^a-z1-9]/i, '-')}-#{form_id}"
+  end
```
smoyth, line 28: Seems like this file should be in the `lib` directory, not in initializers.

Or better yet, it seems like it really should be a class that has a `form` instance variable and wraps Form to provide this functionality. The way it's written now is not very OO. I realize this may be temporary but it also might not be.

smoyth (commented):
Thanks for the quick fix. I added a suggestion about how to structure.

config/initializers/o_data.rb
cooperka: Hah, I didn't even think where to put them when I made this fix... Thanks for pointing it out.

I moved them to `form_decorator` which doesn't seem quite right (since they're OData specific functions rather than generic form behavior) but maybe it's okay since all the other decorators are named after models too. The code certainly reads simpler this way than before. Further suggestions welcome.

cooperka: By the way this initializer was written this way in the first place because it's used by `routes.rb` which needs it right away on startup. The new methods shouldn't have gone there though.

smoyth: Yeah I think you could use a decorator but it should be a different one, not the main one. Any time you have a bunch of prefixed methods like that in a class it's often a smell/suggestion for a new class.

cooperka:
Specs are flappers

smoyth (commented):
Looks fine to me. It's all temporary, right? Except the form decorator?

smoyth:
Decorator looks nice BTW!

cooperka:
It's "temporary" as in a workaround until Data Factory meets our needs. Hopefully the call ~tomorrow gets us further in that direction.

PR #771, opened 2020-10-29
11091 all the cops

cooperka:
All these changes were automatic and should be safe. Separating this from the actual Rails 6 upgrade PR.


cooperka:
Specs are flappers

smoyth:
Cool! So glad this is done!

PR #772, opened 2020-10-29
11091 rails 6

cooperka:
I wrote a lot of info in the commit messages so I truly recommend going one-by-one for this one (there are a lot but they're each fairly straightforward).

Hoping to get this core work released this sprint, then I have more deprecation fixes and upgrades coming soon. SO MANY DEPRECATIONS! 🚀


cooperka:
`Remove redundant save! on create_root_group` has caused additional spec failures, I'll look into those today.

cooperka:
Specs are only flappers now!

PR #773, opened 2020-10-29
11124 fix slow location query

smoyth:


cooperka (approved):
That weird thing happened again where I couldn't load these CSS changes until I ran `rm -rf public/ && gco HEAD public/ && bundle exec rake theme:preprocess` and restarted the server... Anyway, looks great and works great!

Since this performs so well, would you say this is **instead of** or a **precursor to** the new cache table with mission_id, choice_id, answer_id, lat, lng, etc? I still have our pairing notes on that in case they're relevant to next steps fyi.

smoyth:
Yeah this is fine. I don't think we need the separate table now. Better to reduce the scope of the feature (no select multis) than increase code complexity. That is my theme for NEMO right now!

smoyth:
Thanks for testing it out!

smoyth:
@cooperka Before I merge this, I want to make sure that disabling the dashboard for enumerators is ok. My hunch/hope is that it is, but I need to check first.

PR #777, opened 2020-11-10
Add .env.example and document in setup instructions

cooperka:
Now that settings.local.yml.example is gone, shouldn't we have .env.example for local development with all the old values defaulted there instead? (If I understand correctly, the `nemo-deploy` server will control `.env` on cloud servers so this change is only for local development)

Just trying to understand the migration path. Several things have changed and I don't quite remember what else is coming down the pipe.

### DIFF
--- a/docs/development-setup.md
+++ b/docs/development-setup.md
@@ -112,7 +112,7 @@ git checkout develop
 1. Install the required gems by running `bundle install` in the project directory.
 1. Install the required Node modules by running `yarn install` in the project directory.
 1. Run `cp config/database.yml.example config/database.yml`.
-1. Run `cp .env.example .env` and adjust settings as appropriate.
+1. (optional) Run `cp .env.development .env.development.local` and adjust settings as appropriate.
 1. Run `cp config/initializers/local_config.rb.example config/initializers/local_config.rb` and adjust settings as appropriate. Note that the reCAPTCHA and Google Maps API Key must be valid keys for those services in order for tests to pass.
 1. Setup the UUID postgres extension:
     1. On Linux: `sudo -u postgres psql nemo_development -c 'CREATE EXTENSION "uuid-ossp"'`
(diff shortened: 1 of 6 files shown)


docs/production-setup.md
```diff
@@ -367,6 +367,10 @@ Upgrading should be done in stages. Start with the stage closest to your current
 1. [Sentry](https://sentry.io/for/rails/) error tracking is now enabled by default. This means NEMO will automatically send us diagnostic logs when Rails errors occur. These error reports help us fix bugs faster, but may unintentionally reveal information to us about your servers or users. Please disable Sentry in `config/application.rb` (or switch to your own personal URL) if you wish to opt out.
     1. You can still configure the server to email you when errors occur via `config/local_config.rb`; this will not disable Sentry.
 
+#### Upgrading to v11.11
+
+1. Move any relevant configs from `config/settings.local.yml` to `.env`
+
```
smoyth, line 373: Ooh good thinking.

docs/development-setup.md
```diff
@@ -112,7 +112,7 @@ git checkout develop
 1. Install the required gems by running `bundle install` in the project directory.
 1. Install the required Node modules by running `yarn install` in the project directory.
 1. Run `cp config/database.yml.example config/database.yml`.
-1. Run `cp config/settings.local.yml.example config/settings.local.yml` and adjust settings as appropriate.
+1. Run `cp .env.example .env` and adjust settings as appropriate.
```
smoyth, line 115: What if we have a `.env.development` file that is checked into the repo with settings that work for development? Then we could eliminate this step. And in production, one sets up `.env` and the `.env.development` is ignored?

If development things need to be overridden, folks can use `.env.local` or `.env.development.local`.

`.env.*.local` and `.env.local` need to be added to .gitignore.

cooperka:
It's working now locally using `.env.development.local`, I'll get rid of my `settings.local.yml` 🎉

cooperka:
Specs are flappers

PR #778, opened 2020-11-11
11159: Fix user association destruction issues and add specs

smoyth:


app/models/user.rb
```diff
@@ -61,6 +61,10 @@ class User < ApplicationRecord
   attr_accessor(:password_confirmation)
 
   has_many :responses, inverse_of: :user
+  has_many :reviewed_responses, class_name: "Response", foreign_key: :reviewer_id,
```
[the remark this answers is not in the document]
cooperka: Seems relevant, please fix

cooperka (commented):
Looks great besides hound.

smoyth:
Fails = flaps

app/controllers/concerns/application_controller/crud.rb
```diff
@@ -9,10 +9,15 @@ def destroy_and_handle_errors(obj, options = {})
     obj.send(options[:but_first]) if options[:but_first]
     obj.destroy
     flash[:success] = "#{obj.class.model_name.human} #{t('errors.messages.deleted_successfully')}"
-  rescue DeletionError
+  rescue DeletionError # Deprecated
     flash[:error] =
       t($ERROR_INFO.to_s, scope: [:activerecord, :errors, :models, obj.class.model_name.i18n_key],
                           default: t("errors.messages.generic_delete_error"))
+  rescue ActiveRecord::DeleteRestrictionError => e # Preferred method
+    assn_name = error.to_s.split(" ")[-1]
```
cooperka, line 17: Is `error` a rails magic variable or did you mean to use `e` above?

docs/destruction.md
```diff
@@ -0,0 +1,39 @@
+# Project Approach to Object Destruction
```
cooperka, line 1: This is great! Can we link to it from the development readme or somewhere?

config/locales/en/main.yml
```diff
@@ -857,7 +857,9 @@ en:
               blank_unless_goto_end: "This field is required unless you choose 'skip to end of form'."
         user:
           at_least_digits: "Please enter at least %{num} digits."
-          cant_delete_if_responses: "You can't delete this user because they have associated responses."
+          cant_delete_if_has_responses: "You can't delete this user because they have associated responses."
+          cant_delete_if_has_reviewed_responses: "You can't delete this user because they have reviewed responses."
+          cant_delete_if_has_sms_messages: "You can't delete this user because they have associated SMS messages."
```
cooperka, line 862: Will we know to add a new string here if we add a new type of user association? Maybe we could at least note it in the new `destruction.md`...

cooperka: Alternatively could we call it `cant_delete_if_has_associated` and end with `they have an associated: %{type}` so that we don't need to do dynamic key lookup?

cooperka (commented):
Tricky stuff! The dynamic i18n key worries me for long-term maintenance (suggestion in the comments), otherwise this is great and well documented.

docs/destruction.md
smoyth: Hmm, I'm not ready to link to it yet. Not sure where it would be linked to from. It's sort of the first of its kind.

config/locales/en/main.yml
smoyth: The issue is then we need to translate the `type` parameter, which becomes a whole ordeal and we then have to remember to do that. I actually tried it out after reading your suggestion but it gets messy fast.

Generally passing meaty params to i18n like this is almost always problematic due to pluralization rules or things like this. 

I did document it in the .md file though. That's a good idea. Thanks!

smoyth:
Going to wait for CI but I think this is ready to go.

PR #780, opened 2020-11-13
11177: Fix standard report to mark self populated

smoyth:
Apparently this has broken for at least an olympiad...

The standard report was not marking itself populated properly after running, so the view wasn't rendering itself. Only happened in cases where the report data is sent in the AJAX response rather than after a refresh, when it's run anew.


smoyth:
Fails = flaps

cooperka (approved):
Nice sleuthing!

PR #783, opened 2020-11-13
11298: Fix flapping SentryJob spec

smoyth:


smoyth:
Fail is a **BRAND NEW** flapper 🤬 🤬 🤬

I logged it, shouldn't be a hard fix I think.

spec/jobs/sentry_job_spec.rb
```diff
@@ -2,17 +2,23 @@
...
-    Raven.capture_message("Test")
-    expect(Delayed::Job.count).to eq(1)
+    VCR.use_cassette("sentry", match_requests_on: %i[method uri host path]) do
+      begin
+        1 / 0
+      rescue ZeroDivisionError => e
+        Raven.capture_exception(e)
+      end
+      Delayed::Worker.new.work_off
+    end
```
cooperka, line 22: How does this now assert that it really creates the job? Would VCR throw if it doesn't end up handling any requests?

cooperka (approved):
Just a question, VCR seems great for this!

spec/jobs/sentry_job_spec.rb
smoyth: Yes. See the VCR settings in rails_helper. There is an allow_unused_requests or something.

PR #784, opened 2020-11-13
11300 odata flappers

smoyth:


spec/support/contexts/o_data.rb
```diff
@@ -36,12 +36,25 @@ def expect_fixture(filename, forms: [], substitutions: {})
 end
 
 shared_context "odata with basic forms" do
-  let!(:form) { create(:form, :live, mission: mission, question_types: %w[integer select_one text]) }
-  let!(:form_with_no_responses) { create(:form, :live, mission: mission, question_types: %w[text]) }
-  let(:paused_form) { create(:form, :paused, mission: mission, question_types: %w[text]) }
-  let(:draft_form) { create(:form, :draft, mission: mission, question_types: %w[text]) }
+  # We have to explicitly name these forms so we can be sure of their lexical ordering.
+  # If we let the factory assign names with a counter, we may get 8, 9, 10, which are
+  # lexically ordered 10, 8, 9, which causes issues.
```
cooperka, line 41: I used `:reset_factory_sequences` in a few specs to handle this, apparently it wasn't good enough? I don't quite understand why it flaps but appreciate the simple fix!

cooperka (approved):

spec/support/contexts/o_data.rb
smoyth: Yeah not sure about that one!

PR #786, opened 2020-11-17
11300: Set explicit ordering on responses

smoyth:
This one came up again. 

There were actually two sources of indeterminate ordering:

1. The order of the responses in the ODataController
2. The order of the responses in the spec itself.

I set up explicit ordering for both. Hopefully this should fix it. @cooperka please note how I did this in the spec, by just storing the responses in a let var instead of saving and reloading them. If doing the latter, you'd need to use an `order` clause, but storing in a let is more efficient anyway.


smoyth:
Waiting for CI before assigning this one, in case explicit ordering breaks something else.

cooperka (approved):
🥇

PR #787, opened 2020-11-18
11101 some option set efficiencies

smoyth:
Low hanging fruit.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -16,7 +16,7 @@
 #  smsable               :boolean          default(FALSE), not null
 #  standard_copy         :boolean          default(FALSE), not null
 #  status                :string           default("draft"), not null
-#  status_changed_at     :datetime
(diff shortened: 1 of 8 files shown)


app/models/form.rb
```diff
@@ -233,8 +233,14 @@ def last_qing
 
   def update_status(new_status)
     return if new_status == status
-    # Don't run validations in case form has become invalid due to a migration or other change.
-    update_columns(status: new_status, status_changed_at: Time.current)
+
+    old_status = status&.to_sym
+    new_status = new_status.to_sym
+
+    if old_status == :draft || new_status == :draft
```
smoyth, line 240: We only want to update this timestamp if form is going into or out of draft status. That's what really matters in terms of data integrity. Paused <=> live does not.

cooperka (approved):
🐎

PR #788, opened 2020-11-20
10929 sms guide draft form

smoyth:
1. Hide the link in index page and show/edit page if form is draft
2. Show a graceful error message if they visit the URL anyway

### DIFF
--- a/app/controllers/forms_controller.rb
+++ b/app/controllers/forms_controller.rb
@@ -80,6 +80,8 @@ class FormsController < ApplicationController
   end
 
   def sms_guide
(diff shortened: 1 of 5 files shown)


smoyth:
Assuming CI will pass. If not, please fix yourself if it's something straightforward. Thanks :D

cooperka (approved):
Hooray!

The guide is actually viewable for draft forms as long as they have a version (revert to draft => still has sms guide) so maybe that change is not strictly necessary, but I'm happy with it. The UI may be more intuitive this way.

Error message in particular will help a lot because it sounds like most people get here manually by typing in the URL.

cooperka:
Oops, assigning back to you to merge assuming you're okay with the comment.

PR #799, opened 2021-01-06
11418: Show better DropZone error message for unauthorized file uploads

cooperka:
Dropzone shows "There was an error uploading the file" when really you just got signed out due to inactivity and need to refresh the page. The user import dropzone in particular requires re-authorization so this happens in a fairly short time window.

Note: the `RECENT_LOGIN_REQUIRED` string is defined in `authorization.rb`


smoyth:
Fancy!

PR #800, opened 2021-01-07
11371: Switch from Roo to Rails CSV (deprecate XLSX imports)

cooperka:
Todo:
- Update user docs, I can do that around release time


app/models/tabular_import.rb
```diff
@@ -35,10 +35,8 @@ def failed?
   protected
 
   def open_sheet
-    self.sheet = Roo::Spreadsheet.open(file).sheet(0)
-  rescue TypeError, ArgumentError => e
-    raise e unless /not an Excel 2007 file|Can't detect the type/.match?(e.to_s)
-    add_run_error(:wrong_type)
+    self.file = Paperclip.io_adapters.for(file).read if file.class == Paperclip::Attachment
```
cooperka, line 38: I don't love this but it will be going away soon. This is required for feature specs only.

spec/models/tabular_import_spec.rb
```diff
@@ -5,24 +5,24 @@
...
+  let(:import) { TestImport.new(file: file).tap(&:run) }
   let(:run_errors) { import.run_errors }
 
   context "with invalid file type" do
     let(:filename) { "not_tabular.jpg" }
 
     it "should correctly report error" do
       expect(import).not_to be_succeeded
-      expect(run_errors).to eq(["The uploaded file must be a valid XLSX-type Excel file."])
+      expect(run_errors).to eq(["There was an issue with your CSV file: Invalid byte sequence in UTF-8 in line 1."])
```
cooperka, line 25: Not a friendly error message, but users should never see this because their upload will be rejected earlier based on content type. They will see `The uploaded file was not an accepted format.`

cooperka:
Specs are flappers. It just needed an encoding! Ready for review at your leisure, I'll base my ActiveStorage work on this branch.

cooperka:
Travis is not picking up this change... I'll try again tomorrow morning.

config/locales/en/main.yml
```diff
@@ -1501,10 +1501,14 @@ en:
...
-
   option_set_import:
     instructions_html: |-
-      Upload a spreadsheet with your option set. Only XLSX-type Excel spreadsheets and CSV files are accepted.
+      Upload a spreadsheet with your options.
+
+      Notes:
+
+      * Only CSV files are accepted. You can export XLSX files to CSV using any standard spreadsheet software.
+      * You can find more documentation <a href="https://elmo-nemo.readthedocs.io/en/main/content/formsoptionsets/formsoptionsets.html#import-new-option-set">here</a>.
```
smoyth, line 1511: Fancy!

config/locales/en/main.yml
```diff
@@ -2266,18 +2270,18 @@ en:
 
   user_import:
     instructions_html: |-
-      Upload a spreadsheet with your users. Only XLSX-type Excel spreadsheets and CSV files are accepted.
-
-      A spreadsheet template is available in <a href="%{xlsx_url}">XLSX</a> and <a href="%{csv_url}">CSV</a> formats as a starting point.
+      Upload a spreadsheet with your users. A <a href="%{csv_url}">CSV template</a> is available as a starting point.
```
smoyth, line 2273: Can we say "CSV file" instead of "spreadsheet"? The former is the more common way to refer to it I think. "spreadsheet" usually means .xlsx or .numbers or whatever. I don't think people think of CSV files as spreadsheets, do they?

spec/fixtures/user_imports/french.csv
```diff
@@ -1,4 +1,4 @@
-IDENTIFIANT,nom complet,numéro de téléphone principal,Autre numéro de téléphone,E-mail,Année de naissance,Genre,Nationnalité,Observations,Groupes
-user0,User0,17098885555,17098885556,email0@email.com,1981,Homme,Canada,note_0,Les habitants
-user1,User1,,,email1@email.com,,,,note_1,
-user2,User2,,,email2@email.com,,,,note_2,
```
smoyth, line 4: Trying to understand what changed with this file... Line endings? If so how come? If not then what?

smoyth (approved):
Stellar! So glad to see this simplified! Just a few small things... Great work!

smoyth:
@cooperka Thought: did you do a final sweep to make sure XLSX is not mentioned anywhere else in e.g. i18n strings, etc.

I also realize that the fact that opt set exports still go to XLSX makes very little sense. Would it be much work to convert them to CSV now as part of this and just have https://redmine.sassafras.coop/issues/11372 be about the background job piece? Kind of poor planning on my part. But just rendering as CSV instead of XLSX shouldn't be too bad, no?

spec/fixtures/user_imports/french.csv
cooperka: Likely line endings, this must have happened when I was cleaning up trailing commas (auto-added by XLS => CSV conversion for other files), I'll revert it.

config/locales/en/main.yml
cooperka: Agreed, the code comments are meant to be generic so that we don't need to change them again later, but user-facing docs should say CSV for clarity. Will do!

cooperka:
2 new commits to review; waiting on specs...

app/controllers/option_sets_controller.rb
```diff
@@ -151,6 +148,22 @@ def create_or_update
...
+    CSV.generate(headers: true) do |csv|
+      I18n.with_locale(locale) do
+        csv << headers
+
+        rows.each do |row|
+          csv << row
+        end
+      end
+    end
+  end
```
smoyth, line 165: 😎

smoyth (commented):
Glad this was straightforward! Well done!

PR #801, opened 2021-01-13
11425 fix odk formList caching

cooperka:
Fixes 2 issues related to ODK formList caching:
- List doesn't update when form name changes
- List doesn't update when you access it via proxy, so the URLs are invalid and it leads to really strange low-level Android bugs in the Collect app. Relevant when debugging locally but would also apply to any regular proxy in production.


smoyth (approved):
Nice fixes. Thanks!

PR #803, opened 2021-01-28
11415 gh actions

smoyth:


db/migrate/20210126195023_copy_active_storage_from_paperclip_cloud.rb
```diff
@@ -0,0 +1,74 @@
...
+    attachment = record.send(title).attachment
+    if attachment && attachment.created_at > ENV["SKIP_MINUTES_AGO"].to_f.minutes.ago
+      puts "Skipping existing #{filename}"
+      return
+    end
+  end
+
+  puts "Copying #{filename}... (#{index + 1} / #{total})"
+  puts "  at #{url}" if ENV["VERBOSE"]
+  record.send(title).attach(io: URI.open(url), filename: filename,
```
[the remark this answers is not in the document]
cooperka: Ah, I did fix but URI.open has the same risk as Kernel#open -- rubocop did not know about this until v1.0 which was recently upgraded

.github/workflows/verify.yml
```diff
@@ -0,0 +1,89 @@
...
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          # Restore keys will be matched if the main key fails. This may restore older gems to the
+          # dir but we don't care because bundle install will check the gems anyway.
+          restore-keys: |
+            ${{ runner.os }}-gems-
+      - name: Set bundle path
+        run: bundle config path vendor/bundle
+      - name: Install gems
+        # Save some time by not even running bundle install if we had a positive cache hit
+        if: steps.cache-gems.outputs.cache-hit != 'true'
```
cooperka, line 46: Neat 🐎 have you tried this same thing for yarn install?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          cp config/database.yml.ci config/database.yml
+          bin/rails db:setup
+      - name: Install Chromedriver
+        uses: nanasess/setup-chromedriver@master
+      - name: Configure Chromedriver
+        run: |
+          export DISPLAY=:99
+          chromedriver --url-base=/wd/hub &
+      - name: RSpec & JS Tests
```
cooperka, line 75: I think we lost `yarn run lint:react` during this migration, could we do that before specs? (should only take about 5 seconds to run)

.github/workflows/verify.yml
```diff
@@ -0,0 +1,89 @@
...
+        env:
+          # A real Google Maps key is needed for some specs. This is stored in the Github repo
+          # under Settings > Secrets.
+          NEMO_GOOGLE_MAPS_API_KEY: ${{ secrets.NEMO_GOOGLE_MAPS_API_KEY }}
+        shell: bash {0}
+        run: |
+          yarn run i18n:export
+          bundle exec rspec --format RSpec::Github::Formatter --format documentation
+          RSPEC_RESULT=$?
+          yarn run test
```
cooperka, line 87: React tests take about 7 seconds to run, any objection to making them run first so we see output sooner?

.env
```diff
@@ -0,0 +1,57 @@
+# This file holds default settings. They should be sufficient to get a development
+# environment up and running. In some cases the settings will be valid values and suitable
+# for production, such as maximum upload size. In other cases, some settings will be nonsense
+# values, such as access keys.
+#
+# To override these defaults in development (e.g. for developing against an API with real credentials)
+# use the .env.development file, which is git-ignored. To override these settings in production,
+# use the .env.production file, also git-ignored.
```
cooperka, line 8: Ah, I love it -- https://github.com/bkeepers/dotenv#what-other-env-files-can-i-use

(The dotenv library for Node very annoyingly enforces [12-factor app architecture](https://www.npmjs.com/package/dotenv#should-i-have-multiple-env-files) which disallows this sort of inheritance)

cooperka (approved):
Just some minor improvements suggested. Thanks for making this work!!

PR #804, opened 2021-01-29
11093 azure migration

cooperka:
This has already been released to cloud servers. Related PR: #802 

New warning as of Rails 6.1 when generating ERD:
> Warning: Ignoring invalid association :mission on ResponseNode (Could not find the inverse association for mission (:response_nodes in Mission)


cooperka:
Specs are flappers

smoyth (commented):
Looks good to me! Thanks for fixing Rails ERD and the other stuff you cleaned up!

smoyth:
I believe the spec failures are flappers. I rebased 11415 against this branch and it passed in GH actions so pretty sure we're good to go here.

PR #805, opened 2021-02-04
11486 json generator nil bug

smoyth:


cooperka (approved):
Code LGTM, thanks for cleanup too

PR #807, opened 2021-02-10
11414: Gracefully handle InvalidAuthenticityToken

cooperka:
Verification steps: visit the login page and manually muck with the DOM, then try to sign in:

```html
<input type="hidden" name="authenticity_token" value="CHANGE_ME">
```

### DIFF
--- a/spec/features/users/login_spec.rb
+++ b/spec/features/users/login_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+feature "login", js: true do
+  let(:user) { create(:user) }
+
+  before do
+    # Specs have security off by default.
+    ActionController::Base.allow_forgery_protection = true
+  end
+
+  after do
+    ActionController::Base.allow_forgery_protection = false
+  end
+
+  it "should login and show welcome" do
+    real_login(user)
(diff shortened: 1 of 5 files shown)


cooperka:
Specs are flappers

smoyth (commented):
Cool! Should we have a test for this? Seems like the kind of thing that will eventually break or get removed and then re-solved in 5 years if we don't have a test :D

spec/features/users/login_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  it "should login and show welcome" do
+    real_login(user)
+    expect(page).to have_current_path(
+      mission_root_path(mission_name: get_mission.compact_name, locale: "en")
+    )
+  end
+
+  it "should return a user-friendly error when stale" do
+    visit(login_path(locale: "en"))
+    page.execute_script("$('[name=\"authenticity_token\"]').val('invalid')")
```
cooperka, line 26: Tricky spec! I manipulate with JS here because e.g. `first("input").set("invalid authenticity token")` results in Selenium ElementNotInteractableErrors.

smoyth: Nice move!

smoyth (approved):

PR #808, opened 2021-02-15
10398 config overhaul

smoyth:
Sorry for the massive PR!


smoyth:
@cooperka this is ready now. Not sure what Hound is complaining about. It won't show me the errors! I did fix a bunch of lint things earlier.

lib/config_manager.rb
```diff
@@ -0,0 +1,121 @@
...
+  def recaptcha_public_key
+    ENV["NEMO_RECAPTCHA_PUBLIC_KEY"]
+  end
+
+  def recaptcha_private_key
+    ENV["NEMO_RECAPTCHA_PRIVATE_KEY"]
+  end
+
+  def storage_service
+    storage_type = ENV.fetch("NEMO_STORAGE_TYPE")
```
cooperka, line 116: Is the deciding factor that `ENV[]` is for optional configs and `ENV.fetch` is for mandatory configs because it throws? This could easily be mistaken in the future, maybe we should add a more explicit helper or at least class docs.

config/initializers/config_check.rb
```diff
@@ -0,0 +1,56 @@
...
+
+# Check key settings in production to make sure they have safe and sensible values.
+
+# rubocop:disable Style/IfUnlessModifier
+# Don't abort if we're running via a rake task b/c we need to be able to run migrations on deploy
+# since migrations will copy the config to .env.
+if Rails.env.production? && File.split($0).last != "rake"
+  secret_key_base = Rails.configuration.secret_key_base
+  if secret_key_base.blank? || secret_key_base.match?(/XXXXXXXXXXXXXXXX/)
+    abort("NEMO_SECRET_KEY_BASE must be set to a real value")
```
cooperka, line 11: 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

cooperka: Oh, fyi more flexible regex: `/(.)\1{8}/` to detect any obscenely repeated characters. Maybe we could just use this as a constant everywhere.

[Capture any `.`, refer to it with `\1`, then check if it happens 8 times.]

config/initializers/active_record_query_trace.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+ActiveRecordQueryTrace.enabled = true if Rails.env.development? || Rails.env.test?
```
cooperka, line 3: [pending deletion if we use Rails config for this instead, original question above]

lib/tasks/theme.rake
```diff
@@ -17,17 +17,7 @@ namespace :theme do
...
-  desc "Migrates old theme setup to new one."
-  task migrate: :environment do
-    Themeing::Migrator.new.run
-  end
 end
 
-# For now, run migrate on assets:precompile to provide a smooth upgrade path.
-# Can remove this once we think everyone has upgraded to the new themeing setup.
-# This may result in preprocessing being run twice but who cares.
-Rake::Task["assets:precompile"].enhance(["theme:migrate"])
```
cooperka, line 30: I CARE. Thanks Tom!

cooperka (approved):
So much dreamy cleanup! 🔥🔥🔥 Please address some useful remaining hounds (you should be able to see them in the full diff page, GitHub hides them here), otherwise most of my feedback is minor.

.rubocop.yml
```diff
@@ -53,6 +53,11 @@ Layout/LineLength:
 Lint/RedundantCopDisableDirective:
   Enabled: false
 
+Lint/UnusedMethodArgument:
+  # Otherwise we have to remove named args from the method signature altogether which seems to reduce
+  # clarity of the code.
+  AllowUnusedKeywordArguments: true
```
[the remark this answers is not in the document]
smoyth: You can't do that with a keyword argument unless I'm missing something here...

Gemfile
```diff
@@ -182,6 +182,7 @@ group :development, :test do
   # gem "i18n-debug", "~> 1.1" # Great for debugging i18n paths. Uncomment temporarily when neeeded.
 
   # Misc
+  gem "active_record_query_trace", "~> 1.8"
```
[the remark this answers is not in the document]
smoyth: For some reason I wasn't getting what I needed with verbose query logs. Maybe I was in test mode. I'll take it out and if I need it again I can add it back in with an explanation. Thanks!

app/controllers/concerns/application_controller/authorization.rb
```diff
@@ -69,12 +69,12 @@ def handle_recent_login_required(_exception)
   end
 
   def offline_mode?
-    configatron.offline_mode
+    Cnfg.offline_mode?
```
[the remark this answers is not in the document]
smoyth: Config is already taken by something. It's easy to remember—Config with no vowels :D

config/environments/development.rb
```diff
@@ -5,6 +5,7 @@
 
   # Allow local puma-dev domains.
   config.hosts << "nemo.test"
+  config.hosts << "nemo.frappe.sassafras.coop"
```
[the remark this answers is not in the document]
smoyth: Testing incoming SMS :D

lib/config_manager.rb
smoyth: I'll make this clearer.

config/initializers/config_check.rb
smoyth: The main thing I'm trying to catch here is if folks haven't overridden the `.env` file values, which we control. So I don't think we need this extra complexity. I would also have to test it, and testing this stuff is a pain. So I'd like to leave as is for now.

PR #809, opened 2021-02-16
11549: Update odata gem and related gems

smoyth:
The rest of these updated when I ran `bundle update odata_server`. I guess b/c they are dependencies or something.


cooperka (approved):
✨

PR #810, opened 2021-02-22
11542: Fix grid mode support

smoyth:


smoyth:
I haven't tested the output yet on ODK. I'll do that tomorrow.

cooperka (approved):
Code LGTM, glad we could fix this without depending on Collect updates. Definitely want to confirm manual testing with the app before merge.

smoyth:
Tested. Was broken. Fixed. As soon as CI passes this is ready for deploy.

PR #811, opened 2021-02-23
11551 config overhaul cleanup

smoyth:
While it may seem jarring, the majority of the config migration script is now obsolete since configatron is gone. Same for the config migrate down script I think.


Gemfile.lock
```diff
@@ -257,44 +252,6 @@ GEM
...
-      dry-equalizer (~> 0.3)
-      dry-inflector (~> 0.1, >= 0.1.2)
-      dry-logic (~> 1.0, >= 1.0.2)
-    dry-validation (1.5.6)
-      concurrent-ruby (~> 1.0)
-      dry-container (~> 0.7, >= 0.7.1)
-      dry-core (~> 0.4)
-      dry-equalizer (~> 0.2)
-      dry-initializer (~> 3.0)
-      dry-schema (~> 1.5, >= 1.5.2)
```
smoyth, line 297: Wow who knew we'd get rid of all this old guff just by removing the `config` gem!

config/initializers/config_check.rb
```diff
@@ -3,9 +3,7 @@
 # Check key settings in production to make sure they have safe and sensible values.
 
 # rubocop:disable Style/IfUnlessModifier
-# Don't abort if we're running via a rake task b/c we need to be able to run migrations on deploy
-# since migrations will copy the config to .env.
-if Rails.env.production? && File.split($PROGRAM_NAME).last != "rake"
+if Rails.env.production?
```
smoyth, line 6: Don't need to check for rake anymore b/c we assume the migration is already complete by this point. If it isn't then a bunch of other things will break anyway.

cooperka (approved):
🔥 🔥 🔥

PR #812, opened 2021-02-23
11458: Dont hide popover when popover itself is clicked

smoyth:


cooperka (approved):
Beautiful! It works great, this had been a pain point for me especially when testing things.

PR #819, opened 2021-03-15
added download and zip

jpdimond:
Backend for the bulk image download.

The check file size was tested for performance on staging and guinea worm. Performance for zipping the files has not been tested yet.


Gemfile.lock
```diff
@@ -546,6 +546,8 @@ GEM
       activesupport (>= 4.0)
       sprockets (>= 3.0.0)
     sync (0.5.0)
+    sys-filesystem (1.4.1)
+      ffi (~> 1.1)
```
cooperka, line 550: Strange issue to note: This change seems to break the GitHub Actions cache for branches that don't include `ffi`, so automated tests fail to run on ALL branches except for this one. I'm not sure why, just wanted to let you know in case tests fail on other branches.

Should be resolved after this PR is merged, I'll review soon.

PR #820, opened 2021-03-15
11626: Fix SMTP for password reset

cooperka:
Fix SMTP by filtering out blank SMTP options. Deployed in branch `release-12.2.1`.

Moved some code around but the only new thing is the `select` filter.

Specs failing due to this reason: https://github.com/thecartercenter/nemo/pull/819#discussion_r594769156


cooperka:
Approved on Slack

PR #824, opened 2021-03-24
11593: Don't load Sentry in offline mode

cooperka:
And make the DSN configurable because why not. That also makes it easier to manually disable if users want to.


cooperka:
Specs are flappers.

smoyth:
Thank you!

PR #828, opened 2021-04-09
11713: Add admin feature to re-cache form OData

cooperka:
Context: [this Slack thread](https://sassafras.slack.com/archives/C02K5DWP2/p1617997475055600?thread_ts=1617825469.048200&cid=C02K5DWP2) with [this form](https://health.getnemo.org/en/m/malariamonthlyreporting/forms/917d5a48-6d71-4596-96fe-22789bb11683).

Related: I also added a stub server for debugging in https://github.com/thecartercenter/nemo/commit/31cb218c9eb74e9f4eb017bbb9b5f33644c00401.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -66,8 +66,10 @@ class Ability
     can(%i[index show clear destroy], Operation, creator_id: user.id)
     can(:submit_to, Form) { |f| user.admin? || user.assignments.detect { |a| a.mission == f.mission } }
 
-    user.admin? ? admin_permissions : non_admin_permissions
+    # Toggle permissions in order of increasing level of role power.
+    non_admin_permissions unless user.admin?
     mission_dependent_permissions if mission
+    admin_permissions if user.admin?
 
     cannot(:change_assignments, User, id: user.id) unless user.admin?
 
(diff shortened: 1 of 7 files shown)


spec/models/ability/form_ability_spec.rb
```diff
@@ -24,13 +24,25 @@
...
 
+    context "for admin" do
+      let(:user) { create(:user, admin: true) }
+
+      it "should allow re-caching forms" do
+        expect(ability).to be_able_to(:re_cache, Form)
+      end
+    end
+
     context "for coordinator" do
```
jpdimond, line 35: do we need to add another test for other roles to make sure they can't re-cache? Or is this good enough?

cooperka: The below test for enumerators should cover everything else (since only coordinators and admins are touched in abilities).

app/models/ability.rb
```diff
@@ -160,10 +164,13 @@ def coordinator_permissions
...
       end
-      can(:condition_form, Constraint, mission_id: mission.id)
+      # This is intentionally redundant with `manage` as we're slowly migrating to explicit permissions.
       can(:change_status, Form, mission_id: mission.id)
+      cannot(:re_cache, Form)
       can(:update_core, OptionSet, mission_id: mission.id)
 
+      can(:condition_form, Constraint, mission_id: mission.id)
+
       can(%i[read create update update_code update_core export bulk_destroy],
```
jpdimond, line 174: Cool re: moving to explicit actions

PR #829, opened 2021-04-14
11488 option val dynamic

jpdimond:
Complete.

### DIFF
--- a/app/decorators/odk/form_decorator.rb
+++ b/app/decorators/odk/form_decorator.rb
@@ -62,17 +62,17 @@ module ODK
 
     # returns array of option sets that are referenced by the default dynamic calculation
     def referenced_option_sets(qing)
-      # parsing code, belongs in the pattern parser
-      codes = qing.default.scan(/\$(\w+):value+/)
+      codes = qing.default.scan(ODK::DynamicPatternParser::CODE_ONLY_REGEX)
       questions = codes.flatten.map { |c| Question.with_code(c).first }
-      questions[0].nil? ? [] : questions.map { |q| q.option_set.presence }
+      questions.any? ? questions.map { |q| q.option_set.presence } : []
     end
 
     # returns array of option sets needed for dynamic calculations
     def option_sets_for_instances
       opt_sets = enabled_questionings.map do |qing|
-        # parsing code
-        referenced_option_sets(qing) if qing.default.present? && qing.default.include?("value")
+        if qing.default.present? && qing.default =~ ODK::DynamicPatternParser::VALUE_REGEX
+          referenced_option_sets(qing)
+        end
       end
       opt_sets.flatten.compact.uniq
     end
--- a/app/decorators/odk/form_decorator.rb
+++ b/app/decorators/odk/form_decorator.rb
@@ -60,7 +60,6 @@ module ODK
       enabled_questionings.any?(&:preload_last_saved?)
     end
 
-    # TODO: these methods may go elsewhere, e.g. optionset decorator
     # returns array of option sets that are referenced by the default dynamic calculation
     def referenced_option_sets(qing)
       # parsing code, belongs in the pattern parser
--- a/app/decorators/odk/form_decorator.rb
+++ b/app/decorators/odk/form_decorator.rb
@@ -60,20 +60,22 @@ module ODK
       enabled_questionings.any?(&:preload_last_saved?)
     end
 
-    # returns array of option sets that are referenced by the default dynamic calcultion
-    def ref_option_sets(qing)
+    # TODO: these methods may go elsewhere, e.g. optionset decorator
+    # returns array of option sets that are referenced by the default dynamic calculation
+    def referenced_option_sets(qing)
+      # parsing code, belongs in the pattern parser
       codes = qing.default.scan(/\$(\w+):value+/)
       questions = codes.flatten.map { |c| Question.with_code(c).first }
-      questions.map { |q| q.option_set.presence }
+      questions[0].nil? ? [] : questions.map { |q| q.option_set.presence }
     end
 
     # returns array of option sets needed for dynamic calculations
     def option_sets_for_instances
-      opt_sets = []
-      enabled_questionings.each do |qing|
-        opt_sets |= ref_option_sets(qing) if qing.default.present? && qing.default.include?("value")
+      opt_sets = enabled_questionings.map do |qing|
+        # parsing code
+        referenced_option_sets(qing) if qing.default.present? && qing.default.include?("value")
       end
-      opt_sets
+      opt_sets.flatten.compact.uniq
     end
   end
 end
--- a/app/decorators/odk/form_decorator.rb
+++ b/app/decorators/odk/form_decorator.rb
@@ -59,5 +59,21 @@ module ODK
     def needs_last_saved_instance?
       enabled_questionings.any?(&:preload_last_saved?)
     end
+
+    # returns array of option sets that are referenced by the default dynamic calcultion
+    def ref_option_sets(qing)
+      codes = qing.default.scan(/\$(\w+):value+/)
+      questions = codes.flatten.map { |c| Question.with_code(c).first }
+      questions.map { |q| q.option_set.presence }
+    end
+
+    # returns array of option sets needed for dynamic calculations
+    def option_sets_for_instances
+      opt_sets = []
+      enabled_questionings.each do |qing|
+        opt_sets |= ref_option_sets(qing) if qing.default.present? && qing.default.include?("value")
+      end
+      opt_sets
+    end
   end
 end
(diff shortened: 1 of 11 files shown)


spec/models/odk/response_pattern_parser_spec.rb
```diff
@@ -144,6 +144,30 @@
     end
   end
```
jpdimond, line 146: ignore this spec for now

app/decorators/odk/option_set_decorator.rb
```diff
@@ -27,9 +27,27 @@ def translation_tags(lang)
       tags.reduce(&:<<)
     end
 
+    # Returns <instance> for option sets that have dynamic calcs only
+    # Only supports level 1 option sets
+    def instances_for_dynamic_calcs
+      content_tag(:instance, id: "#{odk_code}_level1") do
```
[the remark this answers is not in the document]
jpdimond: cascading instance blocks collisons

app/models/odk/dynamic_pattern_parser.rb
```diff
@@ -3,8 +3,8 @@
 module ODK
   # Abstract parent class for classes that parse $-style patterns for ODK.
   class DynamicPatternParser
-    # Basic regex for codes like $Question7
-    CODE_REGEX = /[$]!?[A-z]\w+/.freeze
+    # Basic regex for codes like $Question7 or $Question3:value
+    CODE_REGEX = /[$]!?[A-z]\w+:value|[$]!?[A-z]\w+/.freeze
```
cooperka, line 7: Tip: You can simplify to `/[$]!?[A-z]\w*(:value)?/`

Nice tool that validates and explains each and every symbol in your expression: https://regex101.com/

cooperka: Separate issue I just noticed: this doesn't actually support single-character question codes like `$x`, so we should also replace the `+` with `*` (this is already done above in my modified suggestion).

app/models/odk/dynamic_pattern_parser.rb
```diff
@@ -96,6 +99,14 @@ def process_code(code)
       codes_to_outputs[code] = calculated? ? "(#{unwrapped})" : unwrapped
     end
 
+    def option_value?(code)
+      code =~ /:value/
+    end
+
+    def qing_code(code)
+      code.match(/\w+/)[0]
```
cooperka, line 107: Nice 😎

app/models/odk/response_pattern_parser.rb
```diff
@@ -18,9 +18,14 @@ def join_tokens(tokens)
...
-      xpath = src_item.xpath_to(target_qing_or_subqing(other_qing))
-      other_qing.has_options? ? itext_expr(xpath) : xpath
+    def build_output(other_qing, option_value: false)
+      xpath = if option_value
+                other_qing.xpath_to_option_value
+              else
+                src_item.xpath_to(target_qing_or_subqing(other_qing))
+              end
+
+      other_qing.has_options? && !option_value ? itext_expr(xpath) : xpath
```
cooperka, line 28: Since we just return `xpath` directly if option_value, let's simplify this and just do it earlier:

```ruby
return other_qing.xpath_to_option_value if option_value
# ...code that was there before
```

app/views/forms/odk/_instances.xml.erb
```diff
@@ -11,9 +11,14 @@
   <% end %>
 </instance>
 
+<%# Instances for first level option sets if needed for dynamic calculations %>
+<% @form.option_sets_for_instances.each do |option_set| %>
+  <%= ODK::DecoratorFactory.decorate(option_set).option_instances_for_dynamic_calcs %>
+<% end %>
```
cooperka, line 17: The structure of this block and the block below differ even though they seem to be doing very similar things, and I wonder if we can make them more consistent for readability.

Can we assign a new `@option_sets_for_instances` in the same way that `@option_sets` (used below) gets assigned? And if it gets decorated like that one too, we'd just do `option_set.option_instances_for_dynamic_calcs`

spec/decorators/odk/form_decorator_spec.rb
```diff
@@ -54,4 +54,40 @@
...
+
+      it "should return zero option sets" do
+        expect(decorated_form.option_sets_for_instances).to eq([])
+      end
+    end
+
+    describe "with value question code" do
+      before do
+        qing = form.questionings.last
+        qing.default = "calc($likertvalue)"
```
cooperka, line 83: I'm not clear what's being tested here, is it intentionally missing the `:` in `:value`?

spec/models/odk/response_pattern_parser_spec.rb
```diff
@@ -144,6 +144,46 @@
...
+      let(:likert_question) { create(:question, code: "likert1", qtype_name: "select_one", option_set: likert_options) }
+      let(:likert_question2) { create(:question, code: "likert2", qtype_name: "select_one", option_set: likert_options2) }
+      let(:score) { create(:question, code: "score1", qtype_name: "integer") }
+      let(:form) { create(:form, :live, name: "Dynamic answers for option sets", questions: [likert_question, likert_question2, score]) }
+      let(:q1) { ODK::QingDecorator.decorate(form.c[0]) }
+      let(:q2) { ODK::QingDecorator.decorate(form.c[1]) }
+      let(:q3) { ODK::QingDecorator.decorate(form.c[2]) }
+
+      let(:src_item) { q3 }
+      let(:pattern) { "calc($likert1:value + $likert2:value)" }
```
cooperka, line 177: Nice spec 👍

cooperka (commented):
WOW this looks like the result of a tremendous amount of painstaking work. I really appreciate the documentation, specs, and eye for detail here.

You note in the description:

> This is the first part of the issue that generates the instances in the XML. The calculations do not actually work yet.

Can you briefly explain what's left to get the calculations will work? We can chat on Slack if it's easier.

Most of my comments are just for general enhancements, nothing major 🐝

app/decorators/odk/form_decorator.rb
```diff
@@ -59,5 +59,22 @@ def needs_external_csv?
     def needs_last_saved_instance?
       enabled_questionings.any?(&:preload_last_saved?)
     end
+
+    # returns array of option sets that are referenced by the default dynamic calculation
+    def referenced_option_sets(qing)
+      # parsing code, belongs in the pattern parser
```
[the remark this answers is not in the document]
jpdimond: ah! i was going use the pattern parser code to do that. yes usually I do use a TODO, thanks for catching that.

app/decorators/odk/qing_decorator.rb
```diff
@@ -63,6 +63,11 @@ def constraint_msg(locale)
       msgs.compact.join("; ").presence
     end
 
+    def xpath_to_option_value
+      "instance('#{decorated_option_set.odk_code}_numeric_values')/root/item[itextId=/data/"\
+        "#{odk_code}]/numericValue"
```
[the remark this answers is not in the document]
jpdimond: done

app/models/odk/dynamic_pattern_parser.rb
jpdimond: Yup, i was using that tool!

Cool, i couldn't find a simpler OR operator, so yay for simplification. thank you!

app/models/odk/response_pattern_parser.rb
jpdimond: done!

spec/decorators/odk/form_decorator_spec.rb
jpdimond: yup, just making sure that it isn't read as $likert:value

app/views/forms/odk/_instances.xml.erb
jpdimond: cool, done!

app/models/odk/dynamic_pattern_parser.rb
jpdimond: i tried your suggestion, but it didn't work. i'm also using this tool to ensure i'm using the ruby regex https://rubular.com/

spec/models/odk/response_pattern_parser_spec.rb
jpdimond: :D

app/models/odk/dynamic_pattern_parser.rb
jpdimond: ah learned I needed a non-capturing group!

cooperka (approved):
Wooooo!!! 

Have you tested this on the actual collect app with the latest changes? Whenever I build features for the app, it almost always breaks before merge due to silly untestable things 🙃 

If tested and verified, please go ahead and merge this!

PR #830, opened 2021-04-19
11737: media download checkbox hotfix

jpdimond:
When I originally wrote this, the checkbox parameters would not show up in the controller unless it was checked. So I'm not sure what happened here!


cooperka (approved):
Verified and deploying now! Thanks for the quick fix Jill!

PR #832, opened 2021-04-26
11746: added rescue if duplicate files, added a test

jpdimond:
I'm not sure how the original case happened, perhaps there are some duplicate files? Anyhow this is the best I can think of to deal with this edge case, to just skip it if there is a duplicate.


cooperka:
Closing in favor of https://github.com/thecartercenter/nemo/pull/833

PR #833, opened 2021-04-28
11737 (part 2): add warning alert for Zip::EntryExistsError

jpdimond:
Note from Kevin: This allows us to know how often this rare issue occurs, while still allowing the download to be generated. See also issue 11746.


spec/models/utils/bulk_media_packager_spec.rb
```diff
@@ -5,50 +5,52 @@
...
-    ]
-  end
-
-  describe "bulk image packager" do
-    it "should return the correct size of images" do
-      ability = Ability.new(user: operation.creator, mission: operation.mission)
-      packager = described_class.new(ability: ability, search: nil, operation: operation)
-      size = packager.calculate_media_size
-      expect(size).to equal(1_819_435)
+  context "happy paths" do
```
jpdimond, line 8: this was just adding a context, that's all.

cooperka (approved):
Thanks for the placeholder specs and getting this to the finish line Jill

PR #840, opened 2021-05-13
11782: Fix dynamic values with groups

smoyth:


cooperka (approved):
LGTM, this is super tricky, thanks for the detailed spec updates. I will deploy this change shortly and verify with the production form with Collect.

cooperka:
Spec failures are all known issues.

PR #845, opened 2021-06-01
11805 zip bug fix

jpdimond:


cooperka (approved):
Works great, spec is clear, thanks Jill!

cooperka:
Oops, I was also supposed to address whether we need a migration for this -- yes indeed, and we already have one that does almost exactly what we need so I duplicated it and added your new code here: https://github.com/thecartercenter/nemo/commit/87455b3ebadf6447de1de30d347a4d84d1b0c36d

I'll deploy to staging today as planned to verify.

PR #847, opened 2021-06-07
11818: only accept alphanumeric form names

jpdimond:
I'm sure there are more broken tests with this, so going to push and let the CI run to see where it fails.

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -75,7 +75,7 @@ class Form < ApplicationRecord
 
   after_save :update_minimum
 
-  validates :name, presence: true, length: {maximum: 32}
+  validates :name, presence: true, length: {maximum: 32}, format: {with: /\A[A-Za-z0-9]+\z/}
   validate :name_unique_per_mission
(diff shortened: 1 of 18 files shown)


cooperka:
FYI looks like a lot of spec failures re: `ActiveRecord::RecordInvalid: Validation failed: Name: is invalid`

app/models/form.rb
```diff
@@ -75,7 +75,7 @@ def self.receivable_association
 
   after_save :update_minimum
 
-  validates :name, presence: true, length: {maximum: 32}
+  validates :name, presence: true, length: {maximum: 32}, format: {with: /\A[A-Za-z0-9]+\z/}
```
cooperka, line 78: Can we also allow space, dash, underscore? Those should ideally be safe. I also wonder if this is too restrictive because it will disallow any accented characters like é which should ideally also be safe.

Have you already checked with Dottie to see which form exactly is breaking the API, and what the error is? It may actually be a very specific kind of symbol that does it, such as a slash... I know Zied has reported that specific symbol in the past.

PR #848, opened 2021-06-09
[hotfix] 11830 include conditional qing codes in CSV exports

cooperka:
~The main change here is that we now pull in `cached_json.keys` and include those in the CSV header~

^ That didn't actually work for all cases, but what does work is using `form.descendants` to get a list of all questions in the form.

See the spec file I added before and after the second commit to see the user-facing change here -- `IntegerQ2` is now present in the CSV headers even though it was never submitted as an answer (because the display logic hides it).

I will deploy this to the health server after specs pass; we can deploy this to all servers next cycle, no rush.

### DIFF
--- a/app/models/o_data/simple_entities.rb
+++ b/app/models/o_data/simple_entities.rb
@@ -14,6 +14,12 @@ module OData
       FormName: :string
     }.freeze
 
+    # These have been added on since an earlier full re-cache,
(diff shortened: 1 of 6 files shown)


cooperka:
Specs are flappers.

jpdimond:
This looks good to me Kevin!!! Nice work here!

PR #849, opened 2021-06-22
11818: restricted form name to chars that are breaking API

jpdimond:


cooperka (approved):
Looks great, thanks for your flexibility here! I think users will be more understanding of this smaller restriction.

PR #850, opened 2021-07-06
11804: added threshold in ui for bulk media download

jpdimond:


app/assets/javascripts/views/export_csv_view.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  spaceLeft() {
+    $.ajax({
+      url: ELMO.app.url_builder.build("media-size"),
+      method: "get",
+      data: "",
+      success: (data) => {
+        $(".calculating-info").hide();
```
cooperka, line 34: For readability and reliability, I like to delegate "ownership" of specific DOM elements to specific methods -- e.g. the `.calculating-info` could always be hidden/shown in the `spaceLeft` method so the logic is self-contained. That way you know that both the `calculateMediaSize` AND `spaceLeft` method are "pure" and don't leave DOM state half-way modified with the assumption that some other method will finish modifying it later (easy for bugs to creep in that way).

So my specific suggestion here would be to move `$(".calculating-info").show();` into this method above `ajax`, and do similar tweaks for other elements if reasonable. It doesn't need to be perfect.

app/assets/javascripts/views/export_csv_view.js
```diff
@@ -0,0 +1,48 @@
...
+        $(".media-info").show();
+
+        if (data.space_on_disk) {
+          $("input[type=submit]").removeAttr("disabled");
+
+        } else {
+          $(".error-info").show();
+        }
+        return data;
+      }
```
cooperka, line 45: We probably also want to handle the ajax `error` state here, and show some similar error message like "could not complete your request, try again".

cooperka (commented):
Looks great other than minor refactoring in JS, and those feature specs! Thanks for the demo.

app/assets/javascripts/views/export_csv_view.js
jpdimond: thanks for this Kevin! super helpful to learn.

app/assets/javascripts/views/export_csv_view.js
jpdimond: done

app/assets/stylesheets/local/responses/_export_csv_modal.scss
```diff
@@ -0,0 +1,10 @@
+#export-csv-modal {
+  .error_box,
```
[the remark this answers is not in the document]
jpdimond: error_box already exists in the code, so i don't want to change...

app/assets/javascripts/views/export_csv_view.js
```diff
@@ -0,0 +1,53 @@
...
+      $("input[type=submit]").removeAttr("disabled");
+      $(".media-info").hide();
+      $(".error-info").hide();
+    }
+  }
+
+  spaceLeft() {
+    $(".calculating-info").show();
+
+    $.ajax({
```
cooperka: In JS, requests like this happen asynchronously by default, so you should use `async/await` if you want to make sure the call finishes before continuing code execution.

I'm not sure if you've used this much in the past, so with a deadline to deploy this week I went ahead and added it -- you can see in https://github.com/thecartercenter/nemo/pull/850/commits/9fb1f0fa89c7f11ac1037418ba2118ad93d035d0, lemme know if you wanna talk more!

cooperka (approved):
It works great 👏 I made small tweaks as mentioned in comments.

Waiting for specs to pass, then I will merge and deploy it this week!

app/models/utils/bulk_media_packager.rb
```diff
@@ -25,15 +25,21 @@ def media_objects_scope
...
+    def media_meta
+      {space_on_disk: space_on_disk?, media_size: bytes_to_mb(media_size)}
+    end
+
+    def media_size
+      @media_size ||= media_objects_scope.sum("active_storage_blobs.byte_size")
     end
 
     def space_on_disk?
+      return false if Rails.env.test? && ENV.key?("STUB_DISK_FULL")
```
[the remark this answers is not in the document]
jpdimond: I paired with Tom on this one morning, and the ENV way was the way he recommended. Even tho i tried your way first lol (it wasn't working)

PR #851, opened 2021-07-06
11849: Paginate operations so it's not so slow

cooperka:
I copied the same way we paginate everything else; nothing special going on here.

Demo:

operations


jpdimond:
great idea, LGTM!

PR #852, opened 2021-07-12
Production setup instruction updates for config migration

smoyth:


cooperka (approved):
Cheers, glad you figured this one out.

PR #855, opened 2021-08-04
11895 missing form hash

smoyth:
Changes how form ODK XML is generated to use a background job instead of rendering on the fly. This way we can easily get the md5 hash which is now required.


app/models/form.rb
```diff
@@ -307,6 +314,11 @@ def increment_version
     update_columns(downloads: 0, published_changed_at: Time.current)
   end
 
+  def odk_xml_md5
+    return nil unless odk_xml.attached?
+    odk_xml.checksum.unpack1("m0").unpack1("H*") # Convert base64 to hex
```
cooperka, line 319: Thanks for all the contextual comments. This is a lot.

cooperka (commented):
Retroactively LGTM, thanks again!

PR #854, opened 2021-08-03
11878 allow date question searchability

cooperka:
(See new docs in code changes for details.)

This was tricky and interesting! I did not know what sql "tsvector" was until now.

It's managed by [hair_trigger](https://github.com/jenseng/hair_trigger).

PR #857, opened 2021-08-13
11878 Sentry for React

cooperka:
Record errors instead of ignoring them, since we have Sentry installed for Rails already.

This was prompted by Dottie's earlier suggestion that the date filter was disappearing (addressed in #854), which may happen if there's an error in React; until now, we wouldn't have any way to diagnose how frequently this occurs.

### DIFF
--- a/app/javascript/components/ErrorBoundary/__snapshots__/component.test.js.snap
+++ b/app/javascript/components/ErrorBoundary/__snapshots__/component.test.js.snap
@@ -1,5 +1,22 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`renders (default) 1`] = `"children"`;
+exports[`renders (default) 1`] = `
+<ErrorBoundary
+  message="message"
+>
+  children
+</ErrorBoundary>
+`;
 
-exports[`renders (error) 1`] = `<DangerousComponent />`;
+exports[`renders (error) 1`] = `
+<ErrorBoundary
+  message="message"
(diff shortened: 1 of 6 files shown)


app/javascript/components/ErrorBoundary/__snapshots__/component.test.js.snap
```diff
@@ -1,5 +1,42 @@
...
+      />
+    }
+  >
+    <ErrorFallback
+      message="message"
+    >
+      <div
+        className="alert alert-danger"
+        role="alert"
+      >
```
cooperka, line 36: Previously, this test was stubbed too heavily, so it didn't actually trigger the error; the component has always worked, but now it properly shows `className="alert alert-danger"` in the snapshot when there's an error.

app/javascript/components/ErrorBoundary/component.js
```diff
@@ -11,29 +18,13 @@ class ErrorBoundary extends Component {
...
-  };
-
-  componentDidCatch(error, info) {
-    if (process.env.NODE_ENV !== 'test') {
-      console.error('[Boundary error]', error);
-      console.error('[Boundary info]', info);
-    }
-
-    this.setState({ hasError: true });
-  }
```
cooperka, line 25: This method is no longer explicitly needed because it's provided by Sentry.ErrorBoundary.

jpdimond (approved):

PR #859, opened 2021-08-17
Add admin link to quick-view form responses odata

cooperka:
I do this constantly, in development and production, so this is a nice convenient link for admins:

Screen Shot 2021-08-16 at 19 04 53


jpdimond (approved):
Looks good to me!

PR #862, opened 2021-09-07
11921: first pass of import option set translations

jpdimond:


config/locales/en/main.yml
```diff
@@ -423,6 +423,7 @@ en:
         base: "" # this is a sort of hack to get error msg to display properly
         coordinates: "Coordinates"
         shortcode: "Shortcode"
+        value: "value"
```
cooperka, line 426: For consistency with the other headers, I'll capitalize this

app/models/option_sets/import.rb
```diff
@@ -29,24 +35,73 @@ def cleaner
...
     end
 
-    def process_row(row, row_idx)
+    def process_row(row, row_idx, languages)
       # Metadata about the option is included at the end of the row array.
       metadata = row.extract_options!
+      if languages.any?
+        translation_row(row, row_idx, metadata, languages)
+      else
+        multilevel_row(row, row_idx, metadata)
```
cooperka, line 58: This is not necessarily multilevel, just untranslated, so I will rename this for clarity.

app/models/option_sets/import_data_cleaner.rb
```diff
@@ -95,5 +98,14 @@ def extract_row_metadata(row, row_num, meta_headers)
       end
       metadata
     end
+
+    def detect_translations(headers)
+      languages = []
+      headers.each do |h|
+        language = h.match(/\[(..)\]/)
```
cooperka, line 105: I think this may match anywhere within the string, I will make sure it only matches the end of the string. I'll also modify to make it acceptable to use more than 2 characters such as `ar_EG`.

cooperka (commented):
Thanks Jill! All specs are flappers, I'll just tweak a few things as mentioned and include it in the predeploy process today 👍

app/models/option_sets/import.rb
```diff
@@ -29,24 +35,73 @@ def cleaner
...
+      end
+    end
+
+    def create_translated_option(row, metadata, languages)
+      option = {
+        mission: mission,
+        canonical_name: row[0],
+        name_translations: build_translation_json(row, languages)
+      }
+      option[:value] = metadata[:value] if metadata[:value].present?
```
cooperka, line 79: Why create this new special header? It seems fine to do, just wondering why it was needed.

spec/models/option_set_import_spec.rb
```diff
@@ -29,6 +28,42 @@
...
+      expect(option_set.options[2].value).to eq(2)
+      expect(option_set.options[2].name_translations["en"]).to eq("teacher")
+      expect(option_set.options[2].name_translations["fr"]).to eq("professeure")
+      expect(option_set.options[2].name_translations["ar_EG"]).to eq("pwofesè")
+      expect(option_set.options[2].name_translations["foo"]).to eq(nil)
+      expect(option_set.options).to have(3).items
+    end
+  end
+
+  context "with mistaken translations" do
```
jpdimond, line 49: Thanks for adding this. I actually had a failure one too, but it seemed to just gracefully handle things to my surprise so I removed it. Shouldn't have done that, so thanks for adding one in!

app/models/option_sets/import.rb
jpdimond: I think....because the way the data cleaner was working, it wasn't included in the row data. I could be wrong, do you want me to look into it next week?

app/models/option_sets/import_data_cleaner.rb
jpdimond: thank you Kevin!

app/models/option_sets/import.rb
cooperka: No need to spend more time on it, as long as there's no underlying reason like "client asked for this specifically" then I'm happy with the paper trail we've got here. If this somehow becomes an issue later, we can update it as needed.

PR #863, opened 2021-09-18
11881: Respect checkboxes when exporting CSV

cooperka:
Well, the 8-year-old javascript helpers are dusty but it still works! Lemme know if you want to pair and I can explain chronologically how the data is flowing.

Please do test this manually if you have time, the DOM stuff is tricky!

### DIFF
--- a/app/assets/javascripts/views/response_list_view.js
+++ b/app/assets/javascripts/views/response_list_view.js
@@ -13,6 +13,8 @@ ELMO.Views.ResponseListView = class ResponseListView extends ELMO.Views.Applicat
   initialize(options) {
     this.$('#export-dropdown').dropdown();
     new Clipboard('#copy-btn-api_url');
+    this.exportWarningThreshold = options.exportWarningThreshold;
+    this.exportErrorThreshold = options.exportErrorThreshold;
     this.reloadCount = 0;
     if (options.refreshInterval > 0) {
       setInterval(this.fetch.bind(this), options.refreshInterval);
@@ -25,13 +27,26 @@ ELMO.Views.ResponseListView = class ResponseListView extends ELMO.Views.Applicat
   showExportCsvModal(event) {
     event.preventDefault();
 
+    // Reset defaults in case the modal is shown several times.
+    $('#export-count-warning').hide();
+    $('#export-count-error').hide();
+    $("input[type=submit]").prop("disabled", false);
+
     // Calculate how many responses will actually be exported.
     let responsesForm = $('.index-table-wrapper form');
     let checked = responsesForm.find('input.batch_op:checked');
     let selectAll = responsesForm.find('input[name=select_all_pages]');
     let shouldExportAll = checked.length === 0 || selectAll.val();
-    let count = shouldExportAll ? $('.index-table-wrapper').data('entries') : checked.length;
-    $('#export-options-summary').text(I18n.t('response.export_options.summary', { count }));
+    let exportCount = shouldExportAll ? $('.index-table-wrapper').data('entries') : checked.length;
+    $('#export-options-summary').text(I18n.t('response.export_options.summary', { count: exportCount }));
+
+    // Toggle elements as needed.
+    if (exportCount >= this.exportErrorThreshold) {
+      $('#export-count-error').show();
+      $("input[type=submit]").prop("disabled", true);
+    } else if (exportCount >= this.exportWarningThreshold) {
+      $('#export-count-warning').show();
+    }
 
     // Save the user's selection to the export options,
     // ensuring we wipe the slate clean first in case they open/close the modal multiple times in a row.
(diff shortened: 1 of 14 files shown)


cooperka:
Specs are flappers, ready for review @jpdimond !

app/assets/javascripts/views/response_list_view.js
```diff
@@ -22,7 +26,42 @@ ELMO.Views.ResponseListView = class ResponseListView extends ELMO.Views.Applicat
...
+    $("input[type=submit]").prop("disabled", false);
+
+    // Calculate how many responses will actually be exported.
+    let responsesForm = $('.index-table-wrapper form');
+    let checked = responsesForm.find('input.batch_op:checked');
+    let selectAll = responsesForm.find('input[name=select_all_pages]');
+    let shouldExportAll = checked.length === 0 || selectAll.val();
+    let exportCount = shouldExportAll ? $('.index-table-wrapper').data('entries') : checked.length;
+    $('#export-options-summary').text(I18n.t('response.export_options.summary', { count: exportCount }));
+
```
jpdimond, line 42: For my learnings... why is all of this here instead of the initialize function in the modal?

app/assets/javascripts/views/response_list_view.js
```diff
@@ -1,3 +1,5 @@
+var CLONE_MARKER = "CLONE_MARKER";
```
jpdimond, line 1: what is a clone marker?

app/controllers/responses_controller.rb
```diff
@@ -129,7 +129,10 @@ def possible_reviewers
 
   def media_size
     ability = Ability.new(user: current_user, mission: current_mission)
-    packager = Utils::BulkMediaPackager.new(ability: ability, search: params[:search], operation: nil)
+    selected = params[:selectAll].present? ? [] : params[:selected] # Empty selection is equivalent to "all".
+    packager = Utils::BulkMediaPackager.new(
+      ability: ability, search: params[:search], selected: selected, operation: nil
```
jpdimond, line 134: thanks for adding this!

app/views/responses/_export_csv_modal.html.erb
```diff
@@ -8,24 +8,23 @@
...
-            <% elsif responses.total_entries >= ResponsesController::CSV_EXPORT_WARNING %>
-              <div class="alert alert-warning">
-                <%= t("response.export_options.export_warning",
-                  threshold: number_with_delimiter(ResponsesController::CSV_EXPORT_WARNING)) %>
-              </div>
-            <% end %>
+            <!-- Shown via JS. -->
+            <div id="export-count-error" class="alert alert-danger hide">
+              <%= t("response.export_options.export_error",
+                threshold: number_with_delimiter(ResponsesController::CSV_EXPORT_LIMIT)) %>
```
jpdimond, line 14: thanks for cleaning this up!

spec/features/responses/csv_export_spec.rb
```diff
@@ -44,18 +45,94 @@
     expect(result[2][9]).to(eq("Plant"))
   end
 
-  scenario "export with threshold warning" do
+  scenario "exporting csv with nothing checked", :js do
```
jpdimond, line 48: wooo beautiful tests!

spec/models/utils/bulk_media_packager_spec.rb
```diff
@@ -26,33 +26,61 @@
...
 
         Zip::File.open(results.to_s) do |zipfile|
+          expect(zipfile.count).to be(2)
           zipfile.each do |file|
             expect(file.name).to match(/nemo-.+-.+.(jpg|png|tiff)/)
           end
         end
       end
+
+      it "should only zip the selection" do
```
jpdimond, line 73: beautiful.

jpdimond (commented):
Thank you so much for building on my work Kevin! I super appreciate all of this!

app/assets/javascripts/views/response_list_view.js
cooperka: Partly for messy reasons (stuff was there originally and I added to it), partly for convenience (I need access to the form OUTSIDE of the modal, so by default I think `$` is scoped to the modal itself and I'd need to do something more complex to search for the outer form.

I realized at the end of this work that it could be moved but the DOM is so fragile I didn't want to test everything over again 🤷

app/assets/javascripts/views/response_list_view.js
cooperka: I'll add a code comment -- it marks elements as having been added automatically (cloned into the form), so that we can destroy them automatically later without fear.

It's for the case when you open the modal, close it, then open it again -- we clean up first by destroying all the old clones.

app/views/responses/_export_csv_modal.html.erb
cooperka: Your version worked great btw, this only needed to change because now `responses.total_entries` can be dynamically overridden if the user checks the boxes.

app/assets/javascripts/views/response_list_view.js
jpdimond: Well I def support that!

PR #866, opened 2021-09-23
Add db:seed to setup instructions

smoyth:


cooperka (approved):
Thanks! Looks like this is already in the dev setup guide too.

PR #868, opened 2021-10-05
11992 os unique name

jpdimond:
Here is the PR for migrating option set names and preventing duplicate names in the future. 

A few notes on this one:

- I did not put a database level constraint on uniqueness on the name column because optionsets can have the same name across multiple missions. Rather, I put a pre-validation check to add a timestamp if it is a duplicate.
- I needed to test this so wrote some tests and put a static method in the class. One of the tests is commented out because it is testing the condition where there are existing option sets with the same name. It requires commenting out the before validation method on line 41 to work. I'm happy to delete those tests and move the method directly in the migration. I'd love to learn how you test migrations if there is a better way!
- the hound errors are for lines that are too long in the tests, but we might delete them, so I can wait if we want to refactor those.


cooperka (commented):
You have a good core here -- I'm gonna suggest a few changes to restructure how things are enforced:
- Let's add a unique constraint in the db: have a look at a similar migration like `DeleteDuplicateChoiceRowsAndAddUniqueIndex` for an example of this. If it's not enforced in the db, it's physically possible that things will slip thru the cracks and those bugs are HARD to diagnose.
- It seems a bit sneaky to rename an option set after a user clicks submit, instead can we show them an error and let them rename it? Searching the code for examples it looks like something similar to this should do it: `validates :name, uniqueness: {scope: :mission_id}`
- After these two changes, the only place you should be adding a date to the name of an OptionSet should be in the migration file itself. This may also make testing easier.
- Please pet the doge 🐶

jpdimond:
Thanks for the review Kevin! Some follow up inline below.


> You have a good core here -- I'm gonna suggest a few changes to restructure how things are enforced:
> 
> * Let's add a unique constraint in the db: have a look at a similar migration like `DeleteDuplicateChoiceRowsAndAddUniqueIndex` for an example of this. If it's not enforced in the db, it's physically possible that things will slip thru the cracks and those bugs are HARD to diagnose.

Did you see my note about a unique constraint in the DB and missions? I think we want to allow the same option set name across different missions? But maybe I'm missing something here.

> * It seems a bit sneaky to rename an option set after a user clicks submit, instead can we show them an error and let them rename it? Searching the code for examples it looks like something similar to this should do it: `validates :name, uniqueness: {scope: :mission_id}`

Sure! 
> * After these two changes, the only place you should be adding a date to the name of an OptionSet should be in the migration file itself. This may also make testing easier.
cool.

> * Please pet the doge 🐶
*pets doggy*

PR #869, opened 2021-10-05
Add clone declaration for SkipRule conditions association

smoyth:
This was an unfortunate ommission. This code doesn't currently have test coverage because it was always kind of a skunkworks thing. Would be good to add some at some point. It would be interesting since you have to have to separate databases. Or perhaps you could just do the export, blow away the test database mid-test, then import and check. But in any case doing that is beyond the scope of this quick fix.


cooperka (approved):
Simple fix, thanks!

PR #874, opened 2021-10-27
11967 doc link

jpdimond:


cooperka (approved):
Looks great, thanks for the styling too!

I didn't like the old way we had been formatting hints, so I also added a slight enhancement to default to `t_markdown` instead of `simple_format` for all translated hint strings which is easier to use and looks nicer.

cooperka:
All specs are flappers

PR #876, opened 2021-11-15
11997 original response xml export

jpdimond:


spec/models/utils/xml_packager_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+    let(:xml_values) { %w[rhys dimond alfred honey] }
+
+    before do
+      response.odk_xml.attach(io: StringIO.new(xml), filename: "odk-response.xml", content_type: "xml")
+      response.save
+    end
+
+    it "should calculate the size of the files" do
+      expect(packager.xml_size).to eq(674)
```
cooperka, line 38: Spec failure: `undefined method 'xml_size'`

spec/models/utils/xml_packager_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+    end
+
+    it "should zip one xml response" do
+      results = packager.download_and_zip_xml
+      expect(results.basename.to_s).to match(/#{operation.mission.compact_name}-xml-responses.+.zip/)
+      expect(File.exist?(results.to_s)).to be(true)
+    end
+  end
+
+  context "response with repeat groups" do
```
cooperka, line 54: Nice, these specs look good

app/models/utils/xml_packager.rb
```diff
@@ -0,0 +1,70 @@
...
+      responses_ids = download_scope.pluck("id")
+      filename = "#{@operation.mission.compact_name}-xml-responses-"\
+        "#{Time.current.to_s(:filename_datetime)}.zip"
+      zipfile_name = Rails.root.join(TMP_DIR, filename)
+      zip(zipfile_name, responses_ids)
+    end
+
+    def human_readable_xml(response)
+      xml_string = response.odk_xml.download
+      questions = response.form.questionings
```
cooperka, line 33: Nit: A question is different from a questioning, let's call this variable `qings` to prevent confusion

app/models/utils/xml_packager.rb
```diff
@@ -0,0 +1,70 @@
...
+      questions.each { |q| xml_string.gsub!("qing#{q.id}", q.name) }
+      xml_string
+    end
+
+    private
+
+    def zip(zipfile_name, responses_ids)
+      Zip::File.open(zipfile_name, Zip::File::CREATE) do |zipfile|
+        responses_ids.each do |rid|
+          response = Response.find(rid)
```
cooperka, line 44: To simplify a bit, instead of iterating over each ID and finding the response again, I think above you can just pass in `download_scope` instead of `responses_ids = download_scope.pluck("id")`

app/models/utils/xml_packager.rb
```diff
@@ -0,0 +1,70 @@
...
+        rescue Zip::EntryExistsError => e
+          report_error(e)
+          next
+        end
+      end
+      zipfile_name
+    end
+
+    def zip_xml(zipfile, response, rid)
+      xml_filename = "nemo-response-#{rid}.xml"
```
cooperka, line 56: Since users most often deal with shortcodes, I say we use shortcode here instead of ID

app/assets/javascripts/views/export_csv_view.js
```diff
@@ -2,30 +2,34 @@
...
       this.enableSubmitButton();
       $(".media-info").hide();
       $(".error-info").hide();
     }
   }
 
-  async spaceLeft() {
-    $(".calculating-info").show();
+  async spaceLeft(downloadType) {
+    downloadType == "media" ? $(".calculating-info").show() : "";
```
cooperka, line 32: Interesting bug: if you check media FIRST then also check and/or uncheck xml, then the "total size" text gets modified/hidden. I think this JS file may need a bit more refactoring to play nicely together.

Suggestion: instead of making media a special case, you could add an additional "total size" row for xml also; the two different rows can be shown/hidden separately based on the checkboxes. Let me know if this is clear or I can explain verbally 😅

app/models/utils/bulk_media_packager.rb
```diff
@@ -27,25 +22,13 @@ def media_objects_scope
...
-        bytes_to_mib(media_size * 2)
-      space_left >= DISK_ALLOWANCE
+      {space_on_disk: space_on_disk?, media_size: bytes_to_mb(download_size)}
     end
 
     def download_and_zip_images
       FileUtils.mkdir_p(Rails.root.join(TMP_DIR))
 
-      media_ids = media_objects_scope.pluck("media_objects.id")
+      media_ids = attachment_scope.pluck("media_objects.id")
```
cooperka, line 31: Spec failure: I think this is supposed to be `download_scope` now.

cooperka (commented):
Awesome refactoring! This is nearly done.

app/models/utils/bulk_media_packager.rb
jpdimond: fixed bulk media failing specs

app/models/utils/xml_packager.rb
jpdimond: done!

app/models/utils/xml_packager.rb
jpdimond: fixed

app/models/utils/xml_packager.rb
jpdimond: done

spec/models/utils/xml_packager_spec.rb
jpdimond: fixed!

app/assets/javascripts/views/export_csv_view.js
jpdimond: fixed! makes sense!

app/helpers/elmo_form_builder.rb
```diff
@@ -191,7 +191,9 @@ def elmo_field(field_name, options)
         @template.content_tag(:div, human_val, class: "ro-val", 'data-val': val != human_val ? val : nil)
 
       else
-        tag_options = options.slice(:id, :data) # Include these attribs with all tags, if given.
+        # TODO: Ideally, any additional options would pass through to all items
+        #   automatically without need to be allowed here explicitly.
+        tag_options = options.slice(:id, :data, :checked) # Include these attribs with all tags, if given.
```
cooperka, line 196: I allowed `:checked` to pass thru to the checkbox so now it works as you intended in ERB, and we don't need to check this box with JS (which was breaking one of the specs you had removed). This was a head-banger!

cooperka (approved):
I fixed one last issue with JS show/hide logic and it seems to all be working! Your updates look great.

elmo_form_builder is in need of some refactoring love (see separate comment related to specs) but that will take MANY hours so I am not going to attempt anything more than that small patch right now.

cooperka:
Specs are flappers

PR #878, opened 2021-11-30
12053 node v16

cooperka:
❗ Abridged instructions for local developers: follow the same guide as documented in docs/production-setup.md here, but don't bother with delayed job (last 2 steps).

### Changes

This upgrades us from node v10 (deprecated) to v16 (supported until 2024). Hopefully it won't break the servers, we'll have to test it on staging as well and I have a plan for that.

This was required in order to support Enketo, and is also a best practice for security and sustainability.


cooperka:
Merging early so GitHub can give us accurate vulnerability reports with these dependency changes, I'll request an explicit review after the holidays from you @jpdimond !

jpdimond:
This looks good to me @cooperka , thank you!

PR #880, opened 2021-12-06
12056: Don't translate and/or constraints

cooperka:
Very simple fix for a very tricky bug (these strings should never actually have been translated using i18n, since they're meant for computers to read).

I have already manually fixed the forms in question by running ODK::FormRenderJob in the server console.

I'm going to hotfix to health right now; let's chat at standup tomorrow about whether this is worth hotfixing to the rest, since it only applies to new standard form copies on missions that have non-english default languages.

### DIFF
--- a/app/decorators/odk/condition_group_decorator.rb
+++ b/app/decorators/odk/condition_group_decorator.rb
@@ -7,7 +7,7 @@ module ODK
       if members.empty?
         negate? ? "false()" : nil
       else
-        conjunction = true_if == "all_met" ? I18n.t("common.and") : I18n.t("common.or")
+        conjunction = true_if == "all_met" ? I18n.t("common.AND") : I18n.t("common.OR")
         result = decorated_members.map { |m| "(#{m.to_odk})" }.join(" #{conjunction} ")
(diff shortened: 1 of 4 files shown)


app/decorators/odk/condition_group_decorator.rb
```diff
@@ -7,7 +7,7 @@ def to_odk
       if members.empty?
         negate? ? "false()" : nil
       else
-        conjunction = true_if == "all_met" ? I18n.t("common.and") : I18n.t("common.or")
+        conjunction = true_if == "all_met" ? "and" : "or"
```
jpdimond, line 10: should this also be case insensitive?

jpdimond: oh wait this is not xpath, so it doesn't matter?

cooperka: do you mean the `all_met` part? right that's coming hardcoded from nemo not xpath. but I'll check in on slack before merging in case I'm misunderstanding

jpdimond: nm, looks good!

PR #881, opened 2021-12-08
12035: scoped refed option sets to mission

jpdimond:


cooperka (approved):
Quick work, simple fix, great spec!

app/decorators/odk/form_decorator.rb
```diff
@@ -63,8 +63,9 @@ def needs_last_saved_instance?
     # returns array of option sets that are referenced by the default dynamic calculation
     def referenced_option_sets(qing)
       codes = qing.default.scan(ODK::DynamicPatternParser::CODE_ONLY_REGEX)
-      questions = codes.flatten.map { |c| Question.with_code(c).first }
-      questions.any? ? questions.map { |q| q.option_set.presence } : []
+      mission = qing.mission
+      questions = codes.flatten.map { |c| Question.find_by(mission_id: mission.id, code: c) }
```
cooperka, line 67: Oh actually `Question.find_by` can return nil, which is causing spec `context "option sets for dynamic default answers"` to fail

cooperka (changes requested):
Noted spec failure, hopefully a simple tweak

app/decorators/odk/form_decorator.rb
```diff
@@ -63,8 +63,9 @@ def needs_last_saved_instance?
...
+      questions.map { |q| q&.option_set.presence }
```
cooperka, line 68: This fixes the spec but I worry it leaves the nil values in the array that's being returned -- I don't know for a fact that this will break things, but in order to make things easier I'm gonna filter out nil values before returning. Will make that tweak and then merge. Thanks again!

cooperka:
Specs are flappers

cooperka (approved):

app/decorators/odk/form_decorator.rb
jpdimond: ahh thanks for fixing Kevin, I was probably moving too fast and misread.

PR #885, opened 2022-01-14
Update email/password validation to appease brakeman

### LINKED ISSUE #884: (question) Should two validations within the user model include /A and /z?
First off, super 'thank you' to y'all for maintaining and developing this! This is more so of a question/concern to address a common question from some of our technical end users than an actual issue with NEMO.

*****

A common, high confidence warning Brakeman says is to use `/A` and `/z` at the beginning and end of a string validation.

```
Confidence: High
Category: Format Validation
Check: ValidationRegex
Message: Insufficient validation for `email` using `/@/`. Use `\A` and `\z` as anchors
File: app/models/user.rb
Line: 112

Confidence: High
Category: Format Validation
Check: ValidationRegex
Message: Insufficient validation for `password` using `/(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])/`. Use `\A` and `\z` as anchors
File: app/models/user.rb
Line: 122
```

```ruby
# app/models/user.rb:112
  validates :email, format: {with: /@/},
                    length: {maximum: 100},
                    allow_blank: true

# app/models/user.rb:122
validates :password,
    confirmation: {if: :require_password?},
    format: {with: PASSWORD_FORMAT, if: :require_password?, message: :invalid_password},
    length: {minimum: 8, if: :require_password?}

# app/models/user.rb:58
PASSWORD_FORMAT = /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])/.freeze
```

In some other parts of the app, `/A` and `/z` *are* used in validation, like  `:name`  in app/models/mission.rb:52. I'm also not knowledgeable with the security implications of that `.freeze` method.

Should the two validations referenced in user.rb also use it, or is this a false positive from Brakeman?

cooperka:
This is indeed best practice, though there are no current security issues with the old way.

Resolves #884

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -55,7 +55,7 @@ class User < ApplicationRecord
   ROLES = %w[enumerator reviewer staffer coordinator].freeze
   SESSION_TIMEOUT = (Rails.env.development? ? 2.weeks : 60.minutes)
   GENDER_OPTIONS = %w[man woman no_answer specify].freeze
-  PASSWORD_FORMAT = /(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])/.freeze
+  PASSWORD_FORMAT = /\A.*(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9]).*\z/.freeze
   DEFAULT_RECENT_LOGIN_MAX_AGE = 60.minutes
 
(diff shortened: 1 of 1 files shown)


jpdimond:
Looks good to me!

PR #889, opened 2022-02-08
12108: removed read from enumerators

jpdimond:
simple fix :-)


cooperka (approved):
Faaaabulous!

And wow look how few flappers today: only 2! I did some upgrades as part of the node 16 stuff and I think those upgrades resolved about 4 flappers.

PR #895, opened 2022-03-02
12129: prevent underscores in form titles

jpdimond:
follow-up to #849 (see also notes in https://redmine.sassafras.coop/issues/10826)

### DIFF
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -77,10 +77,9 @@ class Form < ApplicationRecord
 
   after_save :update_minimum
 
-  # rubocop:disable Style/RegexpLiteral
-  # Do not allow any slashes, parenthesis, underscores, or pipes as this breaks the API
-  validates :name, presence: true, length: {maximum: 32}, format: {without: /[\/()\\|_]/}
-  # rubocop:enable Style/RegexpLiteral
+  # Disallow specific characters or symbol-only names such as `***` which all break Power BI
+  # (documented at https://github.com/thecartercenter/nemo/pull/895).
+  validates :name, presence: true, length: {maximum: 32}, format: {without: %r{([/%_]|\A\W+\z)}}
 
   validate :name_unique_per_mission
   validates_with Forms::DynamicPatternValidator, field_name: :default_response_name
--- a/app/models/form.rb
+++ b/app/models/form.rb
@@ -78,8 +78,8 @@ class Form < ApplicationRecord
   after_save :update_minimum
 
   # rubocop:disable Style/RegexpLiteral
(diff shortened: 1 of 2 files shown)


cooperka:
test results on Power BI Desktop Version: 2.102.845.0 64-bit (February 2022) for Windows 10 21H2:

**Symbol**|**Error?**|**Notes**
:-----:|:-----:|:-----:
letters and numbers|| 
only numbers|| 
accented letters||act like letters: allowable even if it's exclusively accented
starting symbol (then letters)|| 
ending symbol (after letters)|| 
start AND end symbol (letters in between)|| 
ONLY symbols|Power BI error|removed/replaced somehow (*1)
emoji||act like symbols: allowable as long as there are also letters
!|| 
"|| 
#|| 
$|| 
%|Power BI error|url %-encoding issue (*4)
&|| 
'|| 
(||formerly disallowed
)||formerly disallowed
*|| 
+|| 
,|| 
-|| 
.|| 
/|Power BI error|removed/replaced somehow (*2)
:|| 
;|| 
<|| 
=|| 
\>|| 
?|| 
@|| 
[|| 
\\ ||formerly disallowed
]|| 
^|| 
\_|Power BI error|removed/replaced somehow (*3)
`|| 
{|| 
\| ||formerly disallowed
}|| 
~|| 

explicit messages:

* *1: DataSource.Error: OData: The feed's metadata document appears to be invalid. Error: The metadata document could not be read from the message content. InvalidTypeName : The type name 'Collection(NEMO.Group..G1)' is invalid. The type name must be that of a primitive type, a fully qualified name or an inline 'Collection' or 'Ref' type. : (51, 10)
  * form name was `*`
* *2: Cannot get feed for 'Responses: 2021 :: repeats', as we could not resolve its entity type from the metadata.
  * form name was `2021 / repeats`
* *3: Cannot get feed for 'Responses: 2021  repeats', as we could not resolve its entity type from the metadata.
  * form name was `2021 _ repeats`
* *4: DataSource.Error: OData: The context URI 'odata/v1/$metadata#Responses:%202021%20%25%20repeats' references the entity set or type 'Responses: 2021 %25 repeats'. However, no entity set or type with name 'Responses: 2021 %25 repeats' is declared in the metadata.
  * form name was `2021 % repeats`

cooperka (approved):
published results here and updated regex accordingly -- feels good to know it's a very small subset of symbols and almost everything works without error.

if it were much more than this, I'd suggest we look into more advanced workarounds to get these characters to work but it doesn't seem worthwhile for just this.

thanks for the PR and new specs jill! will merge and deploy tomorrow after specs finish running.

PR #896, opened 2022-03-18
12143 response serialization improvements

cooperka:
Several changes:
* Return `:service_unavailable` instead of `:created` when there's a serialization issue, to allow the user to try again momentarily
  * Note: we're NOT yet destroying the original response because I couldn't get the specs to reproduce that behavior and I don't want to ship untestable code; we should see if these changes alone resolve the issue
* Increase MAX_TRIES by 20% in case it helps
* Add a random delay in between tries so that different requests can complete in different orders in case this can reduce the frequency of serialization conflicts
* Improve logging
* Add more specs

### DIFF
--- a/app/models/odk/response_saver.rb
+++ b/app/models/odk/response_saver.rb
@@ -3,7 +3,8 @@
 module ODK
   # Saves responses and is responsible for handling serialization and duplicate errors
   class ResponseSaver
-    MAX_TRIES = 10
+    # The more tries, the more likely it is to eventually succeed
+    MAX_TRIES = 12
 
     def self.save_with_retries!(params)
       tries = 0
@@ -18,7 +19,14 @@ module ODK
         break
       rescue ActiveRecord::SerializationFailure => e
         tries += 1
+
+        # Note for future: Do we still need to destroy the blank response in this case?
+        # See notes related to https://redmine.sassafras.coop/issues/12142.
         raise e if tries >= MAX_TRIES
+
+        # Delay a short, random number of milliseconds
+        # to allow different requests to try in various orders.
+        sleep(rand(1..100) / 1000.0)
       end
     end
   end
--- a/app/models/odk/response_saver.rb
+++ b/app/models/odk/response_saver.rb
@@ -4,8 +4,6 @@ module ODK
   # Saves responses and is responsible for handling serialization and duplicate errors
   class ResponseSaver
     MAX_TRIES = 10
-    # For testing race conditions, stubbed in tests
-    TEST_SLEEP_TIMER = 0
 
     def self.save_with_retries!(params)
       tries = 0
@@ -15,7 +13,6 @@ module ODK
             Sentry.capture_message("Ignored parallel duplicate")
             return
           end
-          sleep(TEST_SLEEP_TIMER)
           params[:response].save!(validate: false)
         end
         break
--- a/app/models/odk/response_saver.rb
+++ b/app/models/odk/response_saver.rb
@@ -5,14 +5,17 @@ module ODK
   class ResponseSaver
     MAX_TRIES = 10
     # For testing race conditions, stubbed in tests
-    SLEEP_TIMER = 0
+    TEST_SLEEP_TIMER = 0
 
     def self.save_with_retries!(params)
       tries = 0
       loop do
         ActiveRecord::Base.transaction(isolation: :serializable) do
-          return if ODK::ResponseParser.duplicate?(params[:submission_file], params[:user_id])
-          sleep(SLEEP_TIMER)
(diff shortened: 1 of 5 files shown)


spec/models/odk/response_saver_spec.rb
```diff
@@ -55,28 +50,20 @@
...
-  def insert_response_via_second_db_connection(checksum)
-    db = ActiveRecord::Base.connection_pool.checkout
-    db.execute("start transaction isolation level serializable;")
-    db.execute("select * from active_storage_blobs WHERE checksum='#{checksum}';")
-    db.execute("INSERT INTO active_storage_blobs
-      (checksum, byte_size, content_type, filename, key, service_name, created_at)
-      VALUES ('#{checksum}', 10000, 'application/xml', 'simple_response.xml', 'abc123', 'test', NOW());")
-    db.execute("COMMIT;")
-    ActiveRecord::Base.connection_pool.checkin(db)
-  end
```
cooperka, line 81: this is amazing and I want to save it for future usage, it will live on in the commit history -- the spec seems to pass as you wrote it even without this trick, but please advise if I'm missing a key ingredient.

spec/models/odk/response_saver_spec.rb
```diff
@@ -55,28 +50,20 @@
         end
 
         thread2 = Thread.new do
-          # wait until the first thread is sleepin
-          sleep(3)
```
cooperka, line 59: I removed the sleep here because when I was writing the new specs in the other file, it turned out not to be necessary -- both threads running simultaneously will try to write to the database at the same time, throwing the error we want.

jpdimond: ah ok, good!

spec/models/odk/response_saver_spec.rb
jpdimond: it passes for me too lol, i am NOT sure. well glad it is simplified then! I think in getting here, I wanted to ensure that i was creating the conditions for this type of collision. but it is just fine to know it works without and that we are getting the error we want!

app/models/odk/response_saver.rb
```diff
@@ -3,22 +3,30 @@
 module ODK
   # Saves responses and is responsible for handling serialization and duplicate errors
   class ResponseSaver
-    MAX_TRIES = 10
-    # For testing race conditions, stubbed in tests
-    SLEEP_TIMER = 0
+    # The more tries, the more likely it is to eventually succeed
+    MAX_TRIES = 12
```
jpdimond, line 7: i trust your assessment here, but wondering how you arrived at 12?

spec/fixtures/odk/responses/nested_group_form_response_duplicate.xml
```diff
@@ -0,0 +1,27 @@
+<?xml version='1.0' ?>
```
jpdimond, line 1: ah good idea to create a lengthy one instead of using sleep to make it take longer, better to match real conditions!

spec/requests/media/odk_response_spec.rb
```diff
@@ -35,14 +47,113 @@
...
+    end
+
+    it "should safely ignore parallel duplicates", database_cleaner: :truncate do
+      original = nil
+      duplicate = nil
+
+      original_file = prepare_and_upload_submission_file("nested_group_form_response.xml")
+      duplicate_file = prepare_and_upload_submission_file("nested_group_form_response_duplicate.xml")
+
+      thread1 = Thread.new do
```
jpdimond, line 79: ah i'm glad with the more complex version of the form you were able to get the test working this way!

spec/requests/media/odk_response_spec.rb
```diff
@@ -35,14 +47,113 @@
       expect(FileUtils.rm(tmp_files)).to be_empty
     end
 
+    it "should safely ignore simple duplicates", database_cleaner: :truncate do
```
jpdimond, line 50: did creating the request specs help reveal anythinig?

cooperka: it revealed that everything works as I would have expected, no weird obvious bugs, and no blank responses being left behind (which is great, and also prolongs the mystery). enough has changed with the new random delay that I don't want to keep throwing more bologna at the wall until we see if this one stuck.

app/models/odk/response_saver.rb
cooperka: I just increased it by 20% up from 10; didn't want to add too many in case it causes servers to process requests more slowly now with the new delay (but at most 12 tries with up to 100 ms in between adds about 1s to request times in the worst case, so there's room to increase this if it helps significantly).

jpdimond: gotcha!

PR #903, opened 2022-04-15
[wip] 12170 dedupe attachments

jpdimond:
This is a start of the happy flow. A couple questions:

- Should I refactor the response parser into a separate class? e.g. duplicate checker or something? it is feeling a little long
- Some tests i'm considering (also in our doc), would love feedback before going down this route.

The following checks same user for the submission

- [x] Part 1: XML + image1 + incomplete, Part 1: XML + image1 + incomplete (duplicate), part 2: XML + image2 (should accept)
- [x] Part 1: XML + image1 + incomplete, Part 2: XML + image2 (should save), Part 2: XML + image2 (should ignore duplicate)
- [x] Part 1: XML + image1 + incomplete, Part 2: XML + image 2, Part1: XML + image1 + incomplete (should ignore)
- [x] Part 1: XML + image1 + incomplete (image1 exists already as part of a diff response)

Different users

- [ ] (user1) Part 1: XML + image1 + incomplete, (user2) Part1: XML + image1 + incomplete
--> currently we have a unique constraint on the database for the checksum of the odk_xml and the formid, so we are currently not allowing diff users submit the same thing.


cooperka (commented):
agreed that refactoring into some sort of helper/utility class seems wise; any time a static method is calling static helpers it starts looking messy. I don't know what the ruby best practice would be here, so let me know if you'd like advice and we can pair.

your test suggestions look great! 💯

cooperka:
ah, another note from support work I did last week: when we delete responses, we do NOT delete the corresponding attachments. this is common practice, and generally not a problem, but may complicate things when verifying checksums. **the existence of an attachment doesn't imply the existence of a real response pointing to that attachment**.

it would be lots more work to cause attachments to be destroyed on response destroy (including cloud attachments that require web requests and can fail) + clean up legacy data across servers, so instead, please make sure your code verifies existence of records pointed to by blobs instead of implicitly assuming they exist.

here's an example diff that was required to run a migration locally on the old code:

```diff
 blobs.each do |blob|
   attachment = ActiveStorage::Attachment.find_by(blob_id: blob.id)
   next if attachment.nil?
-   response = Response.find(attachment.record_id)
-   return true if response.present? && response.user.id == user_id
+   responses = Response.where(id: attachment.record_id)
+   return true if responses.present? && responses.first.user.id == user_id
 end
```

app/models/odk/response_parser.rb
```diff
@@ -217,7 +191,7 @@ def lookup_and_check_form(id, version)
     end
 
     def existing_response
-      existing_response = Response.find_by(odk_hash: calculate_odk_hash, form_id: response.form_id)
+      existing_response = Response.find_by(odk_hash: calculate_odk_hash, form_id: response.form_id, user_id: response.user_id)
```
jpdimond, line 194: This is what would be needed to support diff users with exact same response, but we have  database constraints that prevent doing this.

spec/fixtures/odk/responses/multiple_part_media.xml
```diff
@@ -2,5 +2,5 @@
 <data id="*form1*" version="*formver1*">
   <*itemcode1*>Some words</*itemcode1*>
   <*itemcode2*>the_swing.jpg</*itemcode2*>
-  <*itemcode3*>another_swing.jpg</*itemcode3*>
+  <*itemcode3*>sassafras.jpg</*itemcode3*>
```
jpdimond, line 5: added a diff image so we werent' getting the same checksum

app/models/odk/duplicate_checker.rb
```diff
@@ -0,0 +1,59 @@
...
+    def duplicate?
+      existing_xml_blobs = find_blobs(@files[:xml_submission_file])
+      return false unless dupe_response_and_user?(existing_xml_blobs)
+
+      # If multi part submission, check other attachments
+      other_files = @files.except(:xml_submission_file)
+      return true if other_files.blank?
+
+      other_files.each do |f|
+        return true if dupe_response_and_user?(find_blobs(f[1]))
```
cooperka, line 23: i think this will return a false positive if there are 2 images and either of them (but not both) is a duplicate -- you could use `other_files.all? { |f| dupe?(f) }` to check every one instead of any one.

app/models/odk/response_parser.rb
```diff
@@ -8,36 +8,10 @@ module ODK
...
-      end.base64digest
-    end
-    # rubocop:enable Naming/MethodParameterName, Lint/AssignmentInCondition
-
     def initialize(response: nil, files: nil, awaiting_media: false)
       raise "Submissions must have a mission" if response.mission.nil?
       @response = response
-      @raw_odk_xml = files.delete(:xml_submission_file).read
+      @raw_odk_xml = files[:xml_submission_file].read
       @files = files
```
cooperka, line 15: @files is passed in to answer_parser, so by no longer deleting xml_submission_file above, answer_parser will try to attach the xml to the response as if it were a media attachment.

i can't tell if @files actually needs to include xml_submission_file for anything, maybe we can just revert the line above without issue?

app/models/odk/response_parser.rb
cooperka: gotcha -- yes if we've got database constraints for this, let's simplify! no need to check user id 🎉

spec/requests/media/odk_response_spec.rb
```diff
@@ -207,4 +255,25 @@ def prepare_odk_media_upload_fixture(filename, form)
...
+  def post_submission(submission_file, image_name, image, incomplete)
+    submission_params = {
+      xml_submission_file: submission_file,
+      image_name => image
+    }
+    submission_params["*isIncomplete*"] = "yes" if incomplete
+    post(submission_path, params: submission_params, headers: auth_header)
+  end
+
+  def expect_submission(status, num_response, num_attachments)
```
cooperka, line 268: for clarity/readability of the spec code above, can you please make these method params named? like `expect_submission(status: :created, num_response: 1, num_attachments: 2)`

spec/requests/media/odk_response_spec.rb
```diff
@@ -207,4 +255,25 @@ def prepare_odk_media_upload_fixture(filename, form)
       formver: [form.number],
       itemcode: ODK::DecoratorFactory.decorate_collection(form.preordered_items).map(&:odk_code))
   end
+
+  def post_submission(submission_file, image_name, image, incomplete)
```
cooperka, line 259: same suggestion here: make `incomplete` a named method param (the rest are fine since it's quite obvious what's being passed in)

spec/requests/media/odk_response_spec.rb
```diff
@@ -69,128 +69,176 @@
...
 
-      expect(form_response.form).to eq(form)
-      expect(form_response.answers.count).to eq(3)
+        expect_answers
+      end
+    end
+
+    # Currently database unique constraint does not allow two responses with the same form_id and odk_xml checksum
+    # We should consider if we want to remove this constraint?
+    # What is the liklihood that two users will submit the exact same response and formid?
```
cooperka, line 190: if nobody's complained so far, it seems exceedingly unlikely and not worth spending more effort supporting it. feel free to get rid of these commented specs.

app/models/odk/duplicate_checker.rb
cooperka: in which case, we should have a spec for multiple media attachments :D 

(unless that's difficult for some reason and you can determine that Collect only ever attaches 1 media per submission, which seems plausible but I don't want to rely on it unless we have to)

cooperka (commented):
just a few little things left yay!

app/models/odk/response_parser.rb
jpdimond: removed!

spec/requests/media/odk_response_spec.rb
jpdimond: done

spec/requests/media/odk_response_spec.rb
jpdimond: done

spec/requests/media/odk_response_spec.rb
jpdimond: done

app/models/odk/duplicate_checker.rb
jpdimond: thanks for thinking of this edge case! Given this description from here: https://docs.getodk.org/openrosa-form-submission/#form-submission

It seems that it could be multiple parts.... but not sure if that is how they actually implemented it?

jpdimond: i'm gonna do a multiple media attachment spec...i don't trusts it

cooperka: I think this edge case is still present (the loop ends early with "yes it's duplicate" as soon as it finds a file that's duplicated, even if there may be a file later on that's not yet saved) -- can be resolved by switching to `all?` instead of `each`

app/models/odk/duplicate_checker.rb
```diff
@@ -0,0 +1,59 @@
...
+    def find_blobs(file)
+      checksum = compute_checksum_in_chunks(File.new(file))
+      ActiveStorage::Blob.where(checksum: checksum)
+    end
+
+    def dupe_response_and_user?(blobs)
+      blobs.each do |blob|
+        attachment = ActiveStorage::Attachment.find_by(blob_id: blob.id)
+        next if attachment.nil?
+        response = Response.find(attachment.record_id)
```
cooperka, line 39: `Response.find` will fail if an orphaned attachment no longer has an associated response, let's use safer `find_by(id:)`

cooperka (approved):

PR #911, opened 2022-06-23
12221 fix odata group i18n

cooperka:
Fixes the issue where metadata displays the english group name instead of whatever locale was specified as the mission default (but the response data uses the mission default, so Power BI can't handle the difference).

### DIFF
--- a/app/overrides/o_data_controller_override.rb
+++ b/app/overrides/o_data_controller_override.rb
@@ -34,7 +34,9 @@ ODataController.class_eval do # rubocop:disable Metrics/BlockLength
   end
 
   def transform_schema_for_metadata(_schema)
-    OData::SimpleSchema.new(distinct_forms)
+    in_locale(current_mission.default_locale) do
+      OData::SimpleSchema.new(distinct_forms)
+    end
   end
 
   def transform_json_for_collection(json)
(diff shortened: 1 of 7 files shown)


app/overrides/o_data_controller_override.rb
```diff
@@ -100,4 +102,13 @@ def entity.plural_name
       name
     end
   end
+
+  # Performs an action in the given locale, resetting to the original locale after finishing.
+  def in_locale(locale)
```
cooperka, line 107: Note: duplicated method because this code is run inside of the odata engine, not nemo.

spec/fixtures/odata/multilingual_metadata.xml
```diff
@@ -0,0 +1,44 @@
...
+        </Key>
+        <Property Name="ResponseSubmitDate" Type="Edm.DateTimeOffset" Nullable="true"/>
+        <Property Name="ResponseSubmitterName" Type="Edm.String" Nullable="true"/>
+        <Property Name="ResponseID" Type="Edm.Guid" Nullable="true"/>
+        <Property Name="ResponseShortcode" Type="Edm.String" Nullable="true"/>
+        <Property Name="ResponseReviewed" Type="Edm.Boolean" Nullable="true"/>
+        <Property Name="FormName" Type="Edm.String" Nullable="true"/>
+      </EntityType>
+      <EntityType Name="Responses: *form1*" OpenType="true" BaseType="NEMO.Response">
+        <Property Name="Groupe Un" Type="NEMO.Group.Sample Form 1.G1" Nullable="true"/>
```
cooperka, line 28: this property's `Name` is the only difference from the normal version

jpdimond: tricky! so will this only work for spanish?

jpdimond: rather..french? oh i see nm, this is the fixture.

cooperka:
approved on slack, thanks jill!

PR #915, opened 2022-07-14
12238: Fix crash when trying to create a new form on DF servers

cooperka:
Quick win for the servers that have the `NEMO_USE_DATA_FACTORY` env flag set. This helper is called on `forms/new` causing the crash, but only needed on `forms/show`.


cooperka:
Specs are flappers.

PR #917, opened 2022-07-21
12241 fix multilingual multilevel csv export in non-default language

cooperka:
This specific bug was only happening when the user's locale did not match the mission's default locale. Added a couple new specs for similar cases along with the fix.

Will include in this week's release (rather that hotfixing) after specs pass.


cooperka:
merging for release

PR #919, opened 2022-09-27
12272 enketo upgrade & bug fixes

cooperka:
- Fix Enketo form parsing for extremely large responses (by writing to disk instead of passing as CLI arguments)
- Upgrade enketo & enketo-transformer to latest versions
- Upgrade webpacker and associated libraries (e.g. babel) to latest supported versions (we can't go to v6 yet or remove webpacker entirely because that's a huge refactor but this is at least a big step in the right direction)

Benchmark results since so much changed:
- Our application.js is even smaller than before, woo! (1.42 MB)


app/controllers/responses_controller.rb
```diff
@@ -379,13 +380,18 @@ def enketo
...
-    # as UTF-8 in order for the rest of the page to work for some kinds of forms.
-    command = Terrapin::CommandLine.new("node", ":transformer :xml")
-    command.run(
-      transformer: Rails.root.join("lib/enketo-transformer-service/index.js"),
-      xml: @response.form.odk_xml.download
-    ).force_encoding("utf-8").chomp.html_safe # rubocop:disable Rails/OutputSafety
+    # Save XML to disk instead of trying to pass a massive string
+    # via command line (can fail with exceptionally large forms).
+    FileUtils.mkdir_p(TMP_DOWNLOADS_PATH)
+    @response.form.odk_xml.open(tmpdir: TMP_DOWNLOADS_PATH) do |file|
```
cooperka, line 386: This is what writes to disk

babel.config.js
```diff
@@ -1,58 +1,48 @@
...
-  const currentEnv = api.env();
-  const isDevelopmentEnv = api.env('development');
-  const isProductionEnv = api.env('production');
-  const isTestEnv = api.env('test');
+// See also Ruby config/webpack/environment.js
+//
+// See docs at https://babeljs.io/docs/en/options
+//
+// Initialized based on default config from webpacker at
+// https://github.com/rails/webpacker/blob/master/package/babel/preset.js
```
cooperka, line 6: Most of these changes were just pulled automatically from the preset documented here, aside from our app-specific plugins

config/webpack/environment.js
```diff
@@ -1,11 +1,27 @@
...
+//
+// See docs at https://webpack.js.org/configuration/
+
 // Ignore JS test files.
 environment.loaders.append('ignore', {
   test: /\.test\.js$/,
   loader: 'ignore-loader',
 });
 
+// Transpile Enketo files.
```
cooperka, line 16: This supports a big breaking change in enketo so now we can compile their modern syntax. Previously, they shipped code that was fully transpiled already and we didn't need to take this step.

lib/enketo-transformer-service/index.js
```diff
@@ -1,9 +1,12 @@
+const fs = require('fs');
+
 const transformer = require('enketo-transformer');
 
 async function transform() {
   // Parse command-line arguments.
   const args = process.argv.slice(2);
-  const xform = args[0];
+  const xformPath = args[0];
+  const xform = fs.readFileSync(xformPath);
```
cooperka, line 9: Now that we write to disk, this reads from disk.

cooperka (commented):
Some clarifying comments:

cooperka:
specs are flappers

app/controllers/responses_controller.rb
```diff
@@ -379,13 +380,18 @@ def enketo
...
   # Returns a string that's safe to print in a JS script.
   def enketo_form_obj
-    # Terrapin seems to return an ASCII-encoded string, so we must interpret it
-    # as UTF-8 in order for the rest of the page to work for some kinds of forms.
-    command = Terrapin::CommandLine.new("node", ":transformer :xml")
-    command.run(
-      transformer: Rails.root.join("lib/enketo-transformer-service/index.js"),
-      xml: @response.form.odk_xml.download
-    ).force_encoding("utf-8").chomp.html_safe # rubocop:disable Rails/OutputSafety
+    # Save XML to disk instead of trying to pass a massive string
```
jpdimond, line 383: ah neat way of handling this! is there ever any reason to not write to disk? e.g. if it is a very small string, is it better to just keep it in memory? wondering if there is a threshold, or if it matters at all?

config/webpack/environment.js
jpdimond: hmm wonder why they did that?

PR #920, opened 2022-09-27
12278: Force UTF-8 for converting enketo response to JSON

cooperka:
previously this was a hard error, now it works as expected


spec/features/responses/enketo_spec.rb
```diff
@@ -212,7 +212,7 @@
   # Make it seem like the response was submitted by ODK in the first place.
   def mock_submission(response)
     submission_file = prepare_odk_response_fixture("single_question", form, return_file: true, values: [
-      "enketo answer"
+      "enketo ånswer" # Unicode character å for testing.
```
jpdimond, line 215: ah nice to have a test for this!

PR #925, opened 2022-11-30
12332: enqueue xml job when increment version or update form when live

jpdimond:


app/controllers/forms_controller.rb
```diff
@@ -131,6 +131,8 @@ def update
       @form.assign_attributes(form_params)
       authorize!(:rename, @form) if @form.name_changed?
       @form.save!
+      ODK::FormRenderJob.perform_later(@form) if @form.live?
+
       if params[:save_and_go_live].present?
         @form.update_status(:live)
```
cooperka, line 137: looks like it can also be marked live after the point where we're checking if it's live, i'll make a tweak to address this

spec/features/forms/form/form_status_and_version_spec.rb
```diff
@@ -113,4 +113,41 @@
     click_link("Yourform")
     expect(page).to have_css("div#status", text: /Status\n Live/)
   end
+
+  scenario "changing version when live should enqueue the render form xml job for odk" do
+    visit(forms_path(mode: "m", mission_name: get_mission.compact_name, locale: "en"))
```
cooperka, line 118: weirdly this spec is failing with "mission not found", i'll look into it before merge

cooperka (commented):
simple, great! i'll address a couple small things before merging & running predeploy process today.

app/controllers/forms_controller.rb
cooperka: ah, but setting live causes it to render. so i think this is fine, just confusing to me, i added a comment to clarify.

cooperka (approved):
nice sleuthing! fix makes sense to me (and if it still somehow flaps we'll always catch it later)

somehow this branch got duplicate commits for everything after merging main, so i did a rebase and manually deleted the duplicates. here's the before/after for posterity (git diff is identical regardless of which dupes were chosen, so seems like we're all good):

before, with dupes:

Screen Shot 2022-12-14 at 20 40 46

after deleting the older ones and keeping the newer ones:

Screen Shot 2022-12-14 at 20 45 05

cooperka:
reminder for posterity that this feature was shipped earlier in https://github.com/thecartercenter/nemo/commits/release-13.5, and the merged code is now going out officially in 13.6

PR #934, opened 2023-02-28
12389 export form ODK XML for easier Collect import

cooperka:
Adds 2 new links:
- export ODK XML for a particular form
- export ODK XML for all published forms (zip file)

Docs: https://getnemo.readthedocs.io/en/latest/content/export/xml.html


cooperka:
specs are flappers, ready for review at your leisure!

jpdimond:
This is awesome Kevin!!

I think I was expecting to be able to export the XML on the show/edit form page. Is there a reason you bundled them all up?

cooperka:
it's for both pages: you can export any individual form, or you can export in bulk.

currently, forms aren't rendered into XML until they're published, so that's why it's only for live/paused forms. it would require several additional things in order to render into XML sooner because it depends on some things such as having a version specified.

PR #937, opened 2023-03-21
12404 limit repeats (and ruby 2.7.2)

jpdimond:
adds a "repeat count" field to clamp repeat instances to a specific number

Screen Shot 2023-04-05 at 17 02 28


jpdimond:
@cooperka I'm not sure why the test db is failing, have you seen this before?

app/views/qing_groups/_modal_body.html.erb
```diff
@@ -23,5 +23,9 @@
     <%= f.field(:display_logic,
           partial: "questionings/display_logic",
           locals: {form_item: @qing_group}) %>
+    
+    <%= f.field(:repeat_count,
+      label: t("qing_group.repeat_count"),
+      hint: t("activerecord.hints.qing_group.repeat_count")) if @qing_group.repeatable? %>
```
cooperka, line 29: this field should probably be hidden if you uncheck "Repeatable?" just like "Repeat Item Name" gets hidden. idk where that logic is exactly, likely javascript?

cooperka: oooh could we also make this a dropdown, filtered by only **integer or counter** questions in the form? we're gonna have no end of support requests if people try to enter invalid XML here

cooperka (commented):
hopefully fixed the specs by committing the changes to schema file after running the migration locally 🤞

config/locales/en/main.yml
```diff
@@ -605,6 +605,7 @@ en:
           Treat as text: Accept options for this option set as if they were text entries"
       qing_group:
         group_item_name: "You can enter $QuestionCode to include the value of a question (for ODK Collect only).\n\nFor example, suppose you are making a repeat group where the code for the first question is 'HouseholdID' and you enter\n\n&nbsp;&nbsp;&nbsp;Household: $HouseholdID \n\n You create one group and answer '123' for 'HouseholdID.' Then you create a second group and answer '789' for 'HouseholdID.' The first group item name will be 'Household: 123.' The second group item name will be 'Household: 789.'"
+        repeat_count: "You can enter $QuestionCode to a previous question to set a fixed amount of times a repeat group is repeated. Integers and calculations are not supported at this time."
```
cooperka, line 608: tested with enketo and it's unsupported, let's add this: "ODK Collect only; integers and calculations are not supported at this time."

cooperka: (whole string will also need to be updated if we make it a dropdown)

jpdimond:
> hopefully fixed the specs by committing the changes to schema file after running the migration locally 🤞

ahhhhh that might be part of it. looks like it is still failing with ruby 2.7.1 now, i'll try pushing up with 2.7.2

app/views/qing_groups/_modal_body.html.erb
jpdimond: agreed re: hidden. 

yep, that would be ideal re: having a drop down with filtered questions. I won't have time to do this before I leave, so feel free to take it forward!

cooperka:
specs ran, and are flappers! still working on tweaks before merge.

cooperka:
converted textbox to dropdown to select integer/counter question on the form; tested and verified on Collect!

... this was extremely complex, kudos for implementing this in the first place Jill 🏆🏆🏆

cooperka:
waiting for specs to pass before merge & deploy tomorrow

PR #942, opened 2023-04-04
8634: Update hint text

ozzmotic:
Added hint text to specify only letters, numbers, and spaces. these appear to be the only validations

image

in `app/models/mission.rb`:
```
  validates :name, presence: true
  validates :name, format: {with: /\A[a-z][a-z0-9 ]*\z/i, message: :let_num_spc_only},
                   length: {minimum: 3, maximum: 32}, if: proc { |m| m.name.present? }
  validate :compact_name_unique
```


ozzmotic:
other thought is that the hint text is fairly long and has some useful info. might be better for it to be outside of the hint button in case users don't think to click on it?

also, you can't tab to the hint buttons -> something for later?

ozzmotic:
Re: making the hint buttons tabbable: 

I can do this, but there are a couple issues stopping me right now so I think it would be best for a future PR. For one, the hint element comes before both the label and form field. In `layouts/_elmo_form_field.html.erb` it's clear there is a specific reason for this and I don't want to mess this up:

```
<%= content_tag(:div, class: wrapper_classes, id: field_name, data: {field_name: field_name}) do %>
  <%# This appears first here but is usually moved to the end using flex box %>
  <%# We do it this way so we can move the control to its own line in certain cases %>
  <%= hint_popover(hint_html) %>
  ...
```

Second, the help text only appears when clicking the button -- focusing the element with tab or hitting space doesn't trigger it. I would need to do a little more digging to understand how the hints work.

cooperka (approved):
this PR is great, and thanks for the info/research for additional fixes. that does seem trickier than it initially appeared.

PR #945, opened 2023-04-25
12428 repeat limit bug

jpdimond:


cooperka (approved):
code LGTM -- and please do continue manually testing this one until satisfied, it's a complex sort of change. all specs are passing woo 🎉

jpdimond:
> 

I manually tested this and it works! Given the two test cases, and that the other specs are passing, Im feeling pretty confident. But we can talk more if more testing is needed.

PR #946, opened 2023-05-04
9548: Upgrade date time picker

ozzmotic:
The original issue this PR fixes is that previously, in form submissions, dates and times had to be manually selected from a dropdown and the starting year was in the 1800s. Rather than just changing the start and end dates for the dropdown, this PR adds a new gem, flatpickr. Now users can choose dates and times from an interactive element that appears upon clicking the field.

Note: because flatpickr css has to be integrated into the theme files, you have to run `bundle exec rake theme:preprocess` otherwise it won't work!

Other associated changes:
- In the process of attempting to unit test this, I realized that all unit tests were failing with a google API error. This appeared to be an Apple M1 issue and this was confirmed by @jpdimond. In order to solve this I had to update the gems `selenium-webdriver` and `webdrivers`. 
- Unfortunately, current testing logic doesn't easily enable testing for flatpickr. It was determined that having automated test coverage for it isn't essential, so I have commented out the tests that attempt to fill in dates and times and expect those values in the form spec file.


spec/rails_helper.rb
```diff
@@ -108,17 +107,6 @@
     create(:setting, mission: nil)
   end
 
-  # Print browser logs to console if they are non-empty.
```
ozzmotic, line 111: same as above, this isn't supported in the new version of the webdrivers.

is this required or useful in other contexts or is it okay to delete?

spec/support/contexts/response_tree.rb
```diff
@@ -190,12 +169,4 @@ def expect_read_only_value(path, value)
     el = page.find("[data-path='#{path.join('-')}']")
     expect(el).to have_content(value)
   end
-
-  def temporal_mapping
-    {year: "1i", month: "2i", day: "3i", hour: "4i", minute: "5i", second: "6i"}
-  end
-
-  def control_for_temporal(path, type, subfield)
-    find("##{path_selector(path, "#{type}_value_#{temporal_mapping[subfield]}")}")
```
ozzmotic, line 199: we don't need any of this anymore because we no longer use subfields for times and dates

spec/features/responses/form_rendering_and_submission_spec.rb
```diff
@@ -150,9 +150,11 @@
...
       fill_in_question([7], with: "Dog")
       fill_in_question([8], with: %w[Dog Cat])
-      fill_in_question([9], with: "Mar 12 #{Time.current.year} 18:32:44")
-      fill_in_question([10], with: "Apr 4 #{Time.current.year}")
-      fill_in_question([11], with: "03:08:23")
+      # Note: Dates and times are now selected using a javascript dialog box,
+      # so the current "fill_in_question" method does not work.
+      # fill_in_question([9], with: "#{Time.current.year}-03-12 18:32")
+      # fill_in_question([10], with: "#{Time.current.year}-04-04")
+      # fill_in_question([11], with: "03:08")
```
cooperka, line 157: general note about these commented test lines:
- like jill said, yes let's remove the comments related to time ENTRY (e.g. `fill_in_question`) since we don't need the tests for entry anymore
- but let's make sure there's at least 1 test related to time RENDERING (e.g. `expect_value`) when the time has already been saved in the database, since that's unrelated to the new picker.

app/views/responses/answers/_single_editable.html.erb
```diff
@@ -22,20 +22,23 @@
...
+          dateFormat: "H:i:S",
           noCalendar: true,
-          dateFormat: "H:i",
           time_24hr: true
         })
       })
     <% end %>
+  
+    <% time_val = answer.send("#{type}_value") %>
+    <% if time_val.nil? then time_val = answer.time_value else time_val = l(time_val, format: :time_only) end %>
```
ozzmotic, line 39: This works, but is there a better way to do this? Essentially the problem was that for this time field, it was trying to slot in an entire date object, which in Rails, if there is only a time and no date, always has a default date of 2000-01-01, so the time field would always show 20:00:01. 

To resolve this, I do this logic here to convert the time value to time only.

spec/rails_helper.rb
cooperka: this is fairly important for debugging sometimes -- it's the only place where JS logs (like console.log) get printed when running specs. do you know what specifically doesn't work in the newer version?

spec/support/contexts/response_tree.rb
```diff
@@ -90,18 +90,8 @@ def fill_in_question(path, opts)
...
-        control_for_temporal(path, qtype_name, :year).select(t.strftime("%Y"))
-        control_for_temporal(path, qtype_name, :month).select(t.strftime("%b"))
-        control_for_temporal(path, qtype_name, :day).select(t.day.to_s)
-      end
-      unless qtype_name == "date"
-        control_for_temporal(path, qtype_name, :hour).select(t.strftime("%H"))
-        control_for_temporal(path, qtype_name, :minute).select(t.strftime("%M"))
-        control_for_temporal(path, qtype_name, :second).select(t.strftime("%S"))
-      end
+      fill_in(path_selector(path, "#{qtype_name}_value"), with: value)
```
cooperka, line 94: 🙌

cooperka (commented):
tested in browser, works great!!

i made a couple of small tweaks (see separate comments), and i have a few suggestions:
- all 3 fields currently look "grayed out" when editing, is there an easy way to make the background white so they look more clickable?
- there seems to be a couple of new spec failures, seems like one related to time and one related to webdrivers:
  * `rspec ./spec/features/responses/display_logic_spec.rb:172 # response form display logic different question types various conditions on questionings should work`
  * `rspec ./spec/features/users/user_import_spec.rb:15 # user import happy path`

what do you think about merging this without the webdrivers change for now, and making that change separately?

we could open a new redmine issue for later and prioritize it in the next meeting. it sounds like the specs aren't running on your machine without the webdriver stuff, which is important, but i don't want that change to delay merging this feature since the spec that's breaking may be tricky to debug and getting the JS logs back may require more config work.

spec/rails_helper.rb
ozzmotic: hmm, this was quite a while ago so I'm not exactly sure what was going wrong or the text of the error messages I was getting.

ozzmotic:
> what do you think about merging this without the webdrivers change for now, and making that change separately?

This makes sense!

spec/rails_helper.rb
ozzmotic: Created new issue in redmine and reverted commit!

PR #948, opened 2023-06-01
11090: Add error message for 500 internal server error

ozzmotic:
Issue: Trying to delete an option node that's in use shows a generic "Server Error" page with no details. 

I added in a check for error code 500, however, I'm not able to figure out how to filter specifically for `ActiveRecord::DeleteRestrictionError`. I think it is because this error comes from Rails but this code is catching errors from ajax, so the only info I can access is from that. Maybe there is a way to get more specific with it?

image

### DIFF
--- a/app/controllers/option_sets_controller.rb
+++ b/app/controllers/option_sets_controller.rb
@@ -136,9 +136,9 @@ class OptionSetsController < ApplicationController
     @option_set.save!
     raise StandardError if @option_set.errors?.attribute_names[0] == :name
   rescue ActiveRecord::DeleteRestrictionError
-    render(json: "error", status: :conflict)
+    create_or_update_error(:conflict)
   rescue StandardError
-    render(json: "error", status: :bad_request)
+    create_or_update_error(:bad_request)
   else
     create_or_update_success
   end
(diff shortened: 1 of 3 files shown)


jpdimond:
awesome work! Does the server return anything else other than status: 500 in the response? If not, we can find the place in the code to catch the error and send a more specific error back. Happy to pair on this since it is backend stuff.

app/controllers/option_sets_controller.rb
```diff
@@ -133,16 +133,14 @@ def prep_form_vars
...
-              end
-      render(json: "error", status: error)
-    end
+    @option_set.save!
+    raise StandardError if @option_set.errors?.attribute_names[0] == :name
+  rescue ActiveRecord::DeleteRestrictionError
+    render(json: "error", status: :conflict)
+  rescue StandardError
+    render(json: "error", status: :bad_request)
+  else
```
ozzmotic, line 142: maybe there is a better way to do this that doesn't have multiple render() commands?

jpdimond: ooh this is looking so much better! nice!

you could write a separate render_errors method that takes the status as a param?

ozzmotic: Just attempted this! But I'm noticing many other render() calls in the controller as well...

jpdimond: this is beautiful!!!! 

yeah, that's ok, it's more of a best practice.

PR #950, opened 2023-06-28
Prevent dragging form items in cases related to skip rule targets

ozzmotic:
Within the form editor: 

- If a question to be moved contains a skip rule, it now cannot be moved after its target (i.e, the question that will be skipped to)
- If the item is a skip rule target (i.e., it is a question that would be skipped to if certain conditions were true), it now cannot be moved before the question that refers to it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ NOTES
 frpc.ini
 
 # IDE
+.vscode/*
 .idea/*
 !/.idea/runConfigurations/
 
--- a/app/assets/javascripts/views/form_items_draggable_list_view.js
+++ b/app/assets/javascripts/views/form_items_draggable_list_view.js
@@ -78,50 +78,48 @@ ELMO.Views.FormItemsDraggableListView = class FormItemsDraggableListView extends
     });
   }
 
-  // Checks if the given position (indicated by placeholder) for the given item, or any of its children,
-  // would invalidate any conditions.
+  // Checks if the given position (indicated by placeholder) for the given item, or any of its children, would invalidate any conditions.
   // Returns false if invalid.
   check_condition_order(placeholder, item) {
-    //console.log("check_condition_order called", placeholder, item)
-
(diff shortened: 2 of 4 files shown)


ozzmotic:
Moved to draft, fixing failing specs and errors related to if the rule is to skip to :end

ozzmotic:
I am still getting one failing test in `./spec/features/responses/display_logic_spec.rb`, although I didn't change anything in the display logic. I'm not sure the cause.

ozzmotic:
Removed review request because I'm going to attempt to fix the failing test!

app/assets/javascripts/views/response_condition_checker.js
```diff
@@ -88,7 +88,7 @@ ELMO.Views.ResponseConditionChecker = class ResponseConditionChecker extends ELM
 
   // Uses a special array comparison method if appropriate.
   testEquality(a, b) {
-    if ($.isArray(a) && $.isArray(b)) {
```
ozzmotic, line 91: this syntax is deprecated

app/assets/javascripts/views/response_condition_checker.js
```diff
@@ -137,24 +137,6 @@ ELMO.Views.ResponseConditionChecker = class ResponseConditionChecker extends ELM
             .val();
         }).get();
 
-      case 'datetime': case 'date': case 'time':
```
ozzmotic, line 140: we don't need ANY of this anymore :D

config/locales/en/main.yml
```diff
@@ -908,6 +908,7 @@ en:
     pause: "Pause"
     print: "Print"
     sms_guide: "SMS Guide"
+    destroy: "Delete"
```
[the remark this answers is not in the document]
cooperka: yes perfect. english is edited directly in this file, and other languages are edited on the transifex website (usually by paid translation teams, not us).

cooperka (approved):
:shipit: 💯

PR #954, opened 2023-07-14
7739: Always ignore self-reference

cooperka:
Follow-up to #950 -- add `isSelf` check to each condition. Otherwise, a question that skips depending on its own answer can never be moved.


ozzmotic:
- Confirmed that when an item refers to itself, it can be moved
- Confirmed that when an item refers to itself, it still cannot be moved after its skip target

Looks good to me!

PR #955, opened 2023-07-19
6366, 11820: Fix dragging up out of groups

ozzmotic:
https://github.com/thecartercenter/nemo/assets/4603869/755b7e72-6ea9-40ec-9e79-5a69e7971273

Tested with skip rules


cooperka:
fixing years of strife with only +6 characters 😎 nice work getting to the bottom of this one

PR #960, opened 2023-10-11
12492: fix rendering to XML if there's a disabled cascading select questioning

cooperka:
scenario:
- form has a cascading (multilevel) select that's disabled
- DelayedJob tries to render the form to XML automatically

result:
- `undefined method 'body_tags' for nil:NilClass`

fix:
- don't attempt to render non-existent child

note that BOTH the `if` and `else` eventually attempt to use `visible_children[0]`, but they only currently check if there are `children` and not whether the children are visible. this code logic is extremely complex and branching, but the fix here is likely safe.

tested and verified using the prod form linked from the redmine issue (screenshots not uploaded for privacy reasons).

### DIFF
--- a/app/decorators/odk/qing_group_decorator.rb
+++ b/app/decorators/odk/qing_group_decorator.rb
@@ -174,6 +174,8 @@ module ODK
     end
 
     def main_body_tags(xpath)
+      return +"" unless visible_children.present?
+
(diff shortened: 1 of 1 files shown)


jpdimond:
Looks good to me kevin, thank you!

PR #961, opened 2023-10-12
Add form rank helpers scripts

cooperka:
to audit recent rank mismatch issue; requires more investigation.


lib/tasks/form_rank_helpers.rake
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+namespace :rank do
```
jpdimond, line 3: really nice work kevin!

PR #963, opened 2023-10-18
12494: add new 'meta' section to ODK XML

cooperka:
to fix ODK Central importing:

Screen Shot 2023-10-18 at 16 40 12

tested and verified working after this change.

per the [ODK XForms docs](https://getodk.github.io/xforms-spec/#metadata), this section was missing:

 ```diff
<model>
   <instance>
       <data id="asdf" orx:version="123">
           <question1/>
           <question2/>
+          <orx:meta>
+            <orx:instanceID/>
+          </orx:meta>
       </data>
   </instance>
 </model>
```

note: other meta options include `timeStart`/`timeEnd` (automatic timestamps when the form entry was started/ended), `deviceID`, etc. if we wish to use them in the future.

### DIFF
--- a/app/decorators/odk/form_decorator.rb
+++ b/app/decorators/odk/form_decorator.rb
@@ -9,11 +9,10 @@ module ODK
     IR_QUESTION = "ir01"
     IR_CODE = "ir02"
 
-    def default_response_name_instance_tag
-      if default_response_name.present?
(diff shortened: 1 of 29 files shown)


jpdimond:
Looks great to me, thanks Kevin!

PR #968, opened 2023-11-29
11403 remove old api

jpdimond:
Removes all API and associated tests.


cooperka:
fyi: force-rebased on main after recent PR to see if that helps hound. waiting for specs

PR #972, opened 2023-12-14
12506: discard jobs on ActiveJob::DeserializationError

cooperka:
if, for example, a form (or other object) is created and then deleted before its associated background job can run, the background job will loudly fail with a server alert and possibly cause the DJ queue to backup until it's manually deleted.

instead, we should skip the job if its target no longer exists. tested and verified locally with a series of different fail cases.

the event is also logged to Sentry for analytics.


cooperka:
specs are flappers

PR #990, opened 2024-04-02
Fix relevant column not correctly displaying option set values

plastichotsprings:
Fixes an edge case where conditions would not output correctly when option set node values are what is being conditioned upon


cooperka (approved):
🚀

PR #1031, opened 2025-01-10
12522: Add constraint rejection messages to XLSForm export

plastichotsprings:
This PR adds support to the XLSForm export feature for constraint messages. 

If there are multiple constraint rules for a particular question, the method combines all provided messages into one string (per locale/translation), separated by a semicolon, in order to match the XLSForm standard format.

This PR also slightly rearranges the column orders to put the constraint messages column beside the constraint column for better readability.

image


app/models/forms/export.rb
```diff
@@ -149,8 +154,20 @@ def to_xls
...
+            # NEMO allows multiple constraint messages for each rule, whereas XLSForm only supports one message per row.
+            # Thus, if there are multiple constraints or rules for this question, combine all provided messages into one string (per locale), separated by a semicolon
+            locales.each_with_index do |locale, locale_index|
+              # Attempt to get a message for that constraint for that language (may be nil if a translation is not provided)
+              constraint_message = c.rejection_msg_translations&.dig(locale.to_s)
+
+              if constraint_message.present?
+                constraint_msg_to_push[locale_index] += constraint_message
+                # Add semicolon concatenator, unless at the end
+                constraint_msg_to_push[locale_index] += "; " unless c_index + 1 == q.constraints.length
```
cooperka, line 168: this works perfectly fine, just sharing a helpful alternative: `[1, 2, 3].join("; ")`

app/models/forms/export.rb
```diff
@@ -218,7 +243,11 @@ def to_xls
                 q.question.hint_translations&.dig(locale.to_s))
             end
             questions.row(row_index).push(q.code, q.required.to_s,
-              conditions_to_push, constraints_to_push, choice_filter)
+              conditions_to_push, choice_filter, constraints_to_push)
+
+            constraint_msg_to_push.each do |message|
+              questions.row(row_index).push(message)
+            end
```
cooperka, line 250: rather than having a separate `each` loop here, it may be possible to combine these: 

`[].push(1, 2, *[3, 4])` is equivalent to `[].push(1, 2, 3, 4)`, with the `*` "splat" operator deconstructing the array

cooperka (approved):
tested and verified importing into Central 🙌 nice work

Screenshot

PR #1032, opened 2025-01-17
12624: xlsform export bug fix: duplicate column header error 

plastichotsprings:
if a form references 2 different option sets, and the option sets have a duplicate level name, ODK Central will fail to upload the generated XLSForm with the error: `the XLSForm could not be converted: Duplicate column header`

To fix this, we prepend the option set name (which is unique in NEMO) to the level name whenever it's referenced in the XLSForm.

### DIFF
--- a/app/models/forms/export.rb
+++ b/app/models/forms/export.rb
@@ -154,10 +154,9 @@ module Forms
 
           # if we have an option set, identify and save it so that we can add it to the choices sheet later.
           # then, write the question, splitting it into multiple questions if there are option set levels.
-          os_name = ""
           choice_filter = ""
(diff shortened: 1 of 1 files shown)


cooperka (approved):
nice and straightforward. thanks for mentioning that the os_name is guaranteed unique. 

tested and verified in ODK Central! while testing i did have to add the `allow_choice_duplicates` setting to allow big cascading selects, e.g. where "Main St" might be a street that exists simultaneously in multiple unique cities.

i also created `unique_level_name` so there's only 1 method that generates these level names since it's being done 3 times. wouldn't want to later make a tweak that changes 2/3 and forget the third.

PR #1039, opened 2025-02-14
XLSForm: Fix bug with nested groups not ending correctly

plastichotsprings:
This fixes a bug where nested groups were not ending in the correct order. Previously the code used counters that incremented each time a group was encountered, leaving no way to track which one was most recently begun. Using an array we check the last item to determine which type of group (regular or repeat) should be ending. Then, at the end of the form, we check the array one last time to close groups from last to first, until the array is completely cleared and we are back to base ancestry depth.


db/schema.rb
plastichotsprings, line 1: Uh, I'm not sure how these files changed - I didn't edit them myself?

cooperka: they auto-update with db:migrate, no worries i reverted it

PR #1041, opened 2025-02-26
12614: XLSForm export: Add default value, repeat count, and media prompt columns

plastichotsprings:
Straightforwardly adds additional columns from NEMO question settings. Notes:

- NEMO allows you to "preload last saved answer as a default" - how is this accomplished in XLSForm? (SEE BELOW) ✅ 
- XLSForm allows you to put in a static value as a repeat count, in NEMO you must reference another question's value. ✅ 
- When uploading media_prompt files, NEMO changes the filename to a string of random numbers (e.g., `675f1bab-7de2-4527-b743-fb204c1ebcf5_media_prompt.png`). We may need to warn users because it may not be trivial to actually locate the files they have previously uploaded in order to transfer them over to the new system. Or, is there a way to locate the original filename? ✅ 
- Media_prompt column is tested and working with audio and image uploads, but i didn't have an .mp4 (only allowed video type) of under 50 MB to test with. ✅

### DIFF
--- a/app/models/forms/export.rb
+++ b/app/models/forms/export.rb
@@ -176,6 +176,8 @@ module Forms
           constraint_msg_to_push = Array.new(locales.length, [])
 
           # obtain default response values, or else an empty string
+          # if preload last saved value is checked, indicate this using XLSForm format
+          # https://docs.getodk.org/form-logic/#values-from-the-last-saved-record
           if q.preload_last_saved
             default_to_push = "${last-saved##{q.code}}"
           else
--- a/app/models/forms/export.rb
+++ b/app/models/forms/export.rb
@@ -77,14 +77,8 @@ module Forms
       locales = @form.mission.setting.preferred_locales
 
       # Write sheet headings at row index 0
-      questions.row(0).push("type")
-
-      locales.each do |locale|
-        questions.row(0).push("label::#{language_name(locale)} (#{locale})",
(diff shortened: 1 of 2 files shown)


plastichotsprings:
> ## 🔍 Existing Issues For Review
> 
> Your pull request is modifying functions with the following pre-existing issues:
> 
> 📄 File: **app/models/forms/export.rb**
> Function 	Unhandled Issue
> **`to_xls`** 	[**NoMethodError: undefined method `first' for nil (NoMethodError)**](https://sassafras.sentry.io/issues/6334294216/?referrer=github-open-pr-bot) FormsContro... <br> `Event Count:` **1**
> 
> Did you find this useful? React with a 👍 or 👎

this should be okay since i'm using safe navigation `q.media_prompt.content_type&.split("/")&.first` to get the file type. or is this bad practice..?

plastichotsprings:
"preload last saved answer" would look like this as a default ([docs](https://docs.getodk.org/form-logic/#values-from-the-last-saved-record)): ${last-saved#TextQ1}

cooperka (changes requested):
just needs ::Language settings, but i confirmed it does all work once adding those into the excel file manually

cooperka:
spec is a flapper

cooperka (approved):
tested and verified in central 🏅

PR #1046, opened 2025-03-25
12614: XLSForm export: add support for repeat_item_name

plastichotsprings:
This PR adds XLSForm export support for [naming repeat group instances to help with navigation](https://docs.getodk.org/form-logic/#naming-repeats-to-help-with-navigation). If a NEMO form includes a repeat group with an item name, the export method will now create an inner group whose label is equal to that item name.

### DIFF
--- a/app/models/forms/export.rb
+++ b/app/models/forms/export.rb
@@ -142,7 +142,12 @@ module Forms
 
               # write translated group item names on the inner group row
               locales.each do |locale|
-                questions.row(row_index + 1).push(q.group_item_name_translations&.dig(locale.to_s))
+                # Any instance of "$..." indicates that the user may be referring to another question.
+                # However, XLSForm requires a syntax of "${...}" to refer to questions.
+                # Use regex to make this syntax change.
+                item_name_string = q.group_item_name_translations&.dig(locale.to_s).gsub(/\$(#{Question::CODE_FORMAT})/, "${\\1}")
+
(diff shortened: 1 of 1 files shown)


plastichotsprings:
XLSForm, it is possible to refer to the answers of other questions in the repeat item name using this syntax: `${Question2}`. XLSForm example:
image

In NEMO, this is accomplished with this slightly different syntax (no curly braces): `$Question2`

image

Do we need to use a regex to locate instances of `$` and add curly braces there?

plastichotsprings:
> In XLSForm, it is possible to refer to the answers of other questions in the repeat item name using this syntax: `${Question2}`. XLSForm example: image
> 
> In NEMO, this is accomplished with this slightly different syntax (no curly braces): `$Question2`
> 
> image
> 
> Do we need to use a regex to locate instances of `$` and add curly braces there? But how would we determine where to _end_ the curly brace? Of course, that would also incorrectly modify instances where the dollar sign is used when NOT referring to another question. But maybe that's fine and fairly unlikely?
> 
> We could take the item name string and see if it contains any of the question codes in the form. If it does, find that section of the string and insert a beginning and ending curly brace before and after it ...?

`"asdf $Q1 asdf".gsub(/\$(#{CODE_FORMAT})/, "${\\1}")`

cooperka (approved):
tested and verified in ODK central 🎉

